PR #72, opened 2020-04-10
Titles

CarsonF:
Adds `useTitle` hook and `Title` component to set pieces of the page title in a tree structure

Simple example with routing omitted:
Page title is `Notifications - Profile - CORD Field`
```tsx
const App = () => (
  <TitleProvider title="CORD Field">
    <Profile />
  </TitleProvider>
);

const Profile = () => (
  <div>
    <Title title="Profile" />
    <Notifications />
    ....
  </div>
);

const Notifications = () => {
  useTitle('Notifications');
  return <div>...</div>;
};
```

Also updates automatically with dynamic labels that need to be fetched from API

### DIFF
--- a/src/components/title.tsx
+++ b/src/components/title.tsx
@@ -1,0 +1,69 @@
+import createContainer from 'constate';
+import { unionBy } from 'lodash';
+import { useEffect, useState } from 'react';
+
+export const Title = ({ title }: { title: string }) => {
+  useTitle(title);
+  return null;
+};
+
+export const useTitle = (title?: string) => {
+  const context = useTitleContext();
+  const [id] = useState(uniqid);
+  // Ignore context changes which change when the title changes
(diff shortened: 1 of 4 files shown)


src/components/title.tsx
```diff
@@ -0,0 +1,69 @@
+import createContainer from 'constate';
```
richardgirges, line 1: I understand why the name of the file here is lowercased - because despite this being a `.tsx` file, there is no traditional react component being exported by this file, rather it's just hooks, containers, etc. 

But to the untrained developer eye, this could be seen as precedence to not capitalize component file names. Just something to note as this codebase evolves.

CarsonF: Yeah I wasn't really sure what to call the file since there's 3 things it exports. Good point though.
Maybe we have a separate directory for hooks?

PR #75, opened 2020-04-13
Setup GraphQL Client & Codegen

richardgirges:
## Updates:
* Switched developer server to run on port 3001 instead of port 3000 (since the cord-api-v3 is  3000)
* GraphQL codegen and apollo hooked up, with example component showcasing how to use query and mutation hooks that are auto-generated
* Scenes directory and initial page-routing
* `/devtest` scene for playing around and testing things (will remove this later)

## Problems that need to be addressed
* [x] The git pre-commit hook is throwing an error in regards to the new .eslintignore update:
  * `cord-field/src/generated/graphql.tsx
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override`
* [x] `src/App.tsx` - I could not get TS to stop complaining about `<ApolloProvider />` - kept getting an error regarding `children` being undefined.
* [x] TS relative/absolute auto-import issue


.env
```diff
@@ -0,0 +1 @@
+REACT_APP_SERVER_HOST=http://localhost:3000
```
CarsonF, line 1: What about `REACT_APP_API_BASE_URL`?

codegen.js
```diff
@@ -0,0 +1,38 @@
...
+  path: `.env`,
+});
+
+module.exports = {
+  schema: [
+    {
+      [`${process.env.REACT_APP_SERVER_HOST}/graphql`]: {
+        // TODO: any headers here or is introspection 100% public?
+        // headers: {
+        // },
```
CarsonF, line 11: introspection should be public for us.

package.json
```diff
@@ -2,17 +2,21 @@
   "name": "cord-field",
   "version": "3.0.0-alpha",
   "scripts": {
-    "start": "react-app-rewired start",
+    "start": "PORT=3001 react-app-rewired start",
```
CarsonF, line 5: This can go in your `.env.local` file

package.json
```diff
@@ -2,17 +2,21 @@
...
     "build": "react-app-rewired build",
     "test": "react-app-rewired test",
     "lint": "eslint --ext .ts,.tsx --fix --max-warnings 0 .",
     "type-check": "tsc -p .",
-    "analyze": "yarn build --analyze"
+    "analyze": "yarn build --analyze",
+    "generate": "graphql-codegen --config codegen.js",
+    "generate:watch": "graphql-codegen --watch --config codegen.js"
```
CarsonF, line 12: Is there not a default config location?

src/App.tsx
```diff
@@ -11,6 +14,9 @@ const theme = createTheme();
  * This prevents git diff churning
  */
 const providers = [
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  <ApolloProvider client={apolloClient} />,
```
CarsonF, line 19: Is it complaining about missing children? If so you can do the same thing `ThemeProvider` is doing right below.

src/Root.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Route, Routes } from 'react-router-dom';
+import DevTest from './scenes/DevTest';
+import Home from './scenes/Home';
+
+export default () => (
+  <Routes>
+    <Route path="/" element={<Home />} />
+    <Route path="/devtest" element={<DevTest />} />
```
CarsonF, line 9: What do you think about using `children` instead of `element`? I think it might look cleaner.

package.json
richardgirges: @CarsonF the default config is `codegen.yaml` - we want to use JS so that we can do things like read environment variables

src/App.tsx
richardgirges: Didn't even catch the `ThemeProvider` line, doh!

src/Root.tsx
richardgirges: Good call!

package.json
CarsonF: Oh the backend is the same. Ok then `.env`

Issue #80, opened 2020-04-13
Create Authentication Scenes


CarsonF:
- [ ] Create scenes for
  - [x] Login
  - ~Forgot Password~ Moved to #92 
  - [x] Reset Password
  - [ ] Sign Up
  - The overall layout should be done once separately and/or abstracted
- ~Separate dark theme~ Moved to #113 
  - This could be used anywhere in the app, but will be specifically enforced for those scenes above
  - It only needs to override what will be different from the main theme
- [x] Customization to main theme to make it align with mockups
- [x] Determine session storage w/ @CarsonF 
- [x] Implement authentication with Apollo client

There's purposely a lot of breadth here to help solidify architecture across the codebase.

PR #88, opened 2020-04-14
More storybook stories; hookup Login scene to mutation

richardgirges:
This is just a small step towards #80 

## In this PR:
* Moved components into designated folders (e.g. `ProgressButton.tsx` is now `ProgressButton/index.tsx`)
* Setup hierarchical navigation convention in Storybook
* Created a couple of stories for `ProgressButton` and `ErrorButton`
* Setup a custom storybook configuration (see added files in `./.storybook`) to accomplish the following:
  * Connect to CRA's `config-overrides.js` logic
  * Enable automatic code snippet generation in each story
* Hooked up Login page to `login` mutation (still incomplete, need to figure out post-auth routing, session persistence, etc)
* Created the notion of "Scene" stories
  * Modular components that belong in scenes can also have stories now. e.g. `LoginForm` now has it's own story, despite being co-located with the `Login` scene. This is represented in the navigation hierarchy in storybook

Storybook previews:

image

image

image


CarsonF:
I want to review more that was just the first thing I saw. Overall looks good. I just have some storybook questions. :+1:

richardgirges:
@CarsonF sorry for the lag, just updated the components to be exported via `index.ts` files as suggested

src/scenes/Login/Login.stories.tsx
```diff
@@ -0,0 +1,12 @@
+import { action } from '@storybook/addon-actions';
+import React from 'react';
+import { MockApp } from '../../App';
+import { LoginForm } from './LoginForm';
+
+export default { title: 'Scenes.Login' };
+
+export const _LoginForm = () => (
```
CarsonF, line 8: What's the leading underscore for?

src/App.tsx
```diff
@@ -27,6 +27,14 @@ export const App = () => (
   </Nest>
 );
 
+// Helper component, for wrapping the app's providers around stories and tests
+export const MockApp: FC<{ children: ReactNode }> = ({ children }) => (
+  <Nest elements={providers}>
+    <CssBaseline />
+    {children}
+  </Nest>
+);
```
CarsonF, line 36: I have a few thoughts I'll put all here:
- We probably need always display the storybook components in context of our theme. Otherwise there's going to be a disconnect.
- It would be great to have that automatically applied everywhere.
- Conversely, I don't think we want the other providers initialized. All of these components for storybook should be context free, right? So they don't need any app-level context providers.

src/scenes/Login/index.tsx
```diff
@@ -1,13 +1,31 @@
...
 export const Login = () => {
+  const [login] = useLoginMutation();
+
   return (
     <LoginForm
       onSubmit={async (data) => {
-        await sleep(1000);
-        alert(JSON.stringify(data, undefined, 2));
+        try {
+          const res: LoginMutationResult['data'] = await login({
```
CarsonF, line 12: you shouldn't need to explicitly define the return type...
```suggestion
          const res = await login({
```

src/scenes/Login/index.tsx
```diff
@@ -1,13 +1,31 @@
...
+          // TODO: post-login authentication (session storage, etc)
+          if (res?.login.success) {
+            alert(`Welcome ${res.login.user?.realFirstName.value}`);
+          } else {
+            alert('Login failed. Please check your email or password.');
+          }
+        } catch (e) {
+          alert('Login failed. Please contact support.');
+          console.log(e);
+        }
```
CarsonF, line 28: I know this is still placeholder. But errors can and should be returned as strings or `{ [fieldName]: string }` objects.

src/scenes/Login/queries.graphql
```diff
@@ -0,0 +1,14 @@
+mutation Login($email: String!, $password: String!) {
+  login(input: { email: $email, password: $password }) {
```
CarsonF, line 2: What do you think about using the input object instead, so fields don't have to be explicitly enumerated?
(That was the point of using input objects)
```suggestion
mutation Login($input: LoginInput!) {
  login(input: $input) {
```

.storybook/webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+        options: {
+          presets: [['react-app', { flow: false, typescript: true }]],
+        },
+      },
+      require.resolve('react-docgen-typescript-loader'),
+    ],
+  });
+  config.resolve.extensions.push('.ts', '.tsx');
+  return config;
+};
```
CarsonF, line 24: What are we doing here that was missing before?

richardgirges: The `react-docgen-typescript-loader` was missing before, and is needed in order to make the `@storybook/addon-info` plugin work.

src/scenes/Login/Login.stories.tsx
richardgirges: We're already importing `LoginForm` so that var name is taken. However, Storybook automatically generates titles for the components based on the story name, thus I had to call it `_LoginForm` so that the story would be titled "Login Form"

src/App.tsx
richardgirges: Your last point about not needing providers is a good catch. Now that I'm playing around with it, I'm realizing I didn't need this wrapper at all. I think this was a classic case of me pre-emptively adding complexity where it wasn't needed. I'm going to eliminate the `MockApp` component for now

src/scenes/Login/index.tsx
richardgirges: Whoops! I was debugging something and this shouldn't have made it in.

src/scenes/Login/Login.stories.tsx
CarsonF: Blehhhh. I see, but that's stlil bleh lol. What about
```tsx
import LoginForm as Form
```

src/scenes/Login/queries.graphql
richardgirges: Good call!

.storybook/webpack.config.js
CarsonF: Ok because the addon-info pulls info from TS source right?

By defining this config is a default not applied anymore? Because there's a lot there besides the one loader addition.

Just trying to understand.

src/scenes/Login/index.tsx
richardgirges: Can you elaborate here? Not sure I understand, sorry!

src/App.tsx
CarsonF: > I think this was a classic case of me pre-emptively adding complexity where it wasn't needed.

I NEVER do that 😝 

Sounds good 👍

src/scenes/Login/index.tsx
CarsonF: I'm not saying you need to change anything right now. Eventually we'll want replace the failures alerts with a return value. Which I was trying to say final-form will interrupt as errors to display.

.storybook/webpack.config.js
richardgirges: This webpack config is merged with the default storybook webpack config. So anything in this file should be seen as an incremental override. Frankly, the rabbit hole of research I did on how to get the react-docgen-typescript-loader to work _alongside_ our existing create-react-app storybook configuration led me to this implementation, but I didn't do enough research to determine what we actually needed. 

Check it out now, I think I just simplified it significantly, it looks like there was a lot of unnecessary stuff in there, just based on trial-and-error deleting and reloading storybook after removing line-by-line.

src/App.tsx
richardgirges: 😆

src/scenes/Login/Login.stories.tsx
richardgirges: Good call!

src/scenes/Login/index.tsx
richardgirges: Ohh got it! Actually, this reminds me of another thing we need to keep in mind: error formatting in apollo

.storybook/webpack.config.js
CarsonF: Much better 🤩 Thanks!

src/scenes/Login/index.tsx
CarsonF: Yeah I hoping that now that we are actually consuming the API we'll better know how to do errors from the server.

PR #103, opened 2020-04-21
session & login

richardgirges:
1. Enable `credentials: include` in apolloClient
2. Created `UserSession` context for looking up existing session and hydrating the user session post-login
3. Implemented app bootstrapping in `Root.tsx` - not sure if this is the best place to put this, need feedback here
4. Hydrate `UserSession` context state after successful login and route to the home scene

### DIFF
--- a/src/scenes/Authentication/Login/Login.tsx
+++ b/src/scenes/Authentication/Login/Login.tsx
@@ -1,28 +1,36 @@
 import { ApolloError } from '@apollo/client';
 import { FORM_ERROR } from 'final-form';
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
 import { Except } from 'type-fest';
 import { useLoginMutation } from '../../../api';
+import { useSession } from '../../../components/Session';
 import { LoginForm, LoginFormProps as Props } from './LoginForm';
 
 export const Login = (props: Except<Props, 'onSubmit'>) => {
   const [login] = useLoginMutation();
+  const navigate = useNavigate();
+  const [session, sessionLoading, setUserSession] = useSession();
+
+  useEffect(() => {
+    if (!sessionLoading && session) {
+      navigate('/');
+    }
+  }, [navigate, session, sessionLoading]);
 
   const submit: Props['onSubmit'] = async (input) => {
     const invalidCreds = {
       [FORM_ERROR]: `Something wasn't right. Try again, or reset password.`,
     };
     try {
-      const res = await login({
+      const { data } = await login({
         variables: { input },
       });
 
-      // TODO: post-login authentication (session storage, etc)
-      if (res?.login.success) {
-        alert(`Welcome ${res.login.user?.realFirstName.value}`);
-      } else {
-        return invalidCreds;
-      }
+      // TODO: Navigate to a returnTo place if it exists.
+      navigate('/');
+
+      setUserSession(data.login.user);
     } catch (e) {
       if (
         e instanceof ApolloError &&
--- a/codegen.yml
+++ b/codegen.yml
@@ -8,8 +8,8 @@ hooks:
 generates:
   ./src/api/graphql.generated.tsx:
     plugins:
-      # Until hooks & common can be merged into one import
-      - { add: /* eslint-disable import/no-duplicates */ }
+      # import/no-duplicates is here until hooks & common can be merged into one import
+      - add: /* eslint-disable import/no-duplicates, @typescript-eslint/no-empty-interface */
       - typescript
       - typescript-operations
       - typescript-react-apollo
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -4,6 +4,6 @@ import { App } from './App';
 
 test('renders HOME', () => {
   const { getByText } = render(<App />);
-  const linkElement = getByText(/HOME/i);
+  const linkElement = getByText(/Welcome/i);
   expect(linkElement).toBeInTheDocument();
 });
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,6 +5,7 @@ import { LocalizationProvider } from '@material-ui/pickers';
 import React, { cloneElement, FC, ReactElement } from 'react';
 import { BrowserRouter } from 'react-router-dom';
 import { apolloClient } from './api';
+import { SessionProvider } from './components/Session';
 import { TitleProvider } from './components/title';
 import { Root } from './scenes/Root';
 import { createTheme } from './theme';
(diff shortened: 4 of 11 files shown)


src/components/UserSession/index.tsx
```diff
@@ -0,0 +1,15 @@
+import constate from 'constate';
+import { useState } from 'react';
+import { LoggedInUserFragment } from '../../api';
+
+function useUserSession() {
+  const [userSession, setUserSession] = useState<LoggedInUserFragment | null>(
+    null
+  );
+
+  return { userSession, setUserSession };
```
CarsonF, line 10: Let's move the effect to load the session here. We'll never not want to load the current session, even if it comes from a cache in the future.

Let's also change the return type
```
[userSession, loading, setter]
```
That makes it easier to use a different var name in components. also we'll need the loading boolean.

src/components/UserSession/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export const [UserSessionProvider, useUserSessionContext] = constate(
```
CarsonF, line 13: How about?
```suggestion
export const [UserSessionProvider, useSession] = constate(
```

src/scenes/Authentication/Login/Login.tsx
```diff
@@ -1,25 +1,30 @@
...
+      const { data } = await login({
         variables: { input },
       });
 
       // TODO: post-login authentication (session storage, etc)
-      if (res?.login.success) {
-        alert(`Welcome ${res.login.user?.realFirstName.value}`);
+      if (data?.login?.success) {
+        setUserSession(data?.login?.user);
+        navigate('/');
```
CarsonF, line 27: TODO: Navigate to a `returnTo` place if it exists.

src/scenes/Root.tsx
```diff
@@ -18,6 +20,33 @@ const useStyles = makeStyles(() => ({
...
+
+  React.useEffect(() => {
+    // If the session query has finished loading
+    if (!loadingSession || error) {
+      // Check if an existing session exists and add it to session context
+      if (data?.session.user) {
+        setUserSession(data.session.user);
+      } else {
+        // Otherwise go to the login scene
+        navigate('/login');
```
CarsonF, line 37: We don't want to do this here. There are anonymous routes, like forgot/reset password.

A `useRequireLoggedIn` hook might not be a bad abstraction though.

src/scenes/Root.tsx
```diff
@@ -18,6 +20,33 @@ const useStyles = makeStyles(() => ({
...
+
+      // We've completed our bootstrap
+      setBootstrapping(false);
+    }
+  }, [loadingSession, data, error, setUserSession, navigate]);
+
+  // TODO: what does initial bootstrapping state look like?
+  if (bootstrapping) {
+    return <div>LOADING...</div>;
+  }
```
CarsonF, line 48: Good question. This is fine for now.

typings/react-router-dom/index.d.ts
```diff
@@ -86,8 +86,8 @@ export interface NavigateOptions<S = unknown> {
 }
 
 export function useNavigate<S = unknown>():
-  | ((delta: number) => void)
-  | ((to: LocationDescriptor, options?: NavigateOptions<S>) => void);
+  | ((delta: any) => void)
+  | ((to: any, options?: NavigateOptions<S>) => void);
```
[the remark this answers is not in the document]
richardgirges: We can't pass in any strings to `navigate` with the original typedefs for some reason. e.g. `navigate('/login')` or `navigate('login')` were both invalid.

src/components/UserSession/index.tsx
richardgirges: Is it necessary to send back `setter` since we'll be moving the session query here?

CarsonF: Yeah unfortunately. On login/logout. Unless that can be automatically connected to via apollo client listener?

richardgirges: Err nevermind, we'll need the setter for post-login

typings/react-router-dom/index.d.ts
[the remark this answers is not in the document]
CarsonF: Doh! 🤦 Ok I'll fix.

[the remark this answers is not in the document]
CarsonF: #105

src/components/UserSession/index.tsx
richardgirges: running into a couple of issues with returning arrays for the return type:

1. because we're going with array return type, there will be times when we needlessly import the `session` and `loading` property even when all we need is the setter 

2. for some reason, constate is returning weird typings for the `setter`:
```typescript
const [_userSession, _loading, setUserSession] = useSession();

// See error output from this line:
setUserSession(data.login.user);

^ const setUserSession: boolean | LoggedInUserFragment | React.Dispatch<React.SetStateAction<LoggedInUserFragment | null>> | null
This expression is not callable.
  Not all constituents of type 'boolean | LoggedInUserFragment | Dispatch<SetStateAction<LoggedInUserFragment | null>>' are callable.
    Type 'false' has no call signatures.ts(2349)
```

src/scenes/Root.tsx
richardgirges: what should we do here in the meantime? just allow any route to load until we build the `useRequireLoggedIn` hook?

src/components/UserSession/index.tsx
richardgirges: I just pushed up a version without array types, but instead an object type as a return type. This seems cleaner IMO - let me know what you think.

src/scenes/Root.tsx
richardgirges: Just added a "TODO" for now, let me know how to proceed!

CarsonF: Yeah we don't have to enforce it yet.

src/components/UserSession/index.tsx
CarsonF: It's actually TS trying to infer the type based on the array items. It thinks you want "an array with items that are of any of these types".

Instead if you declare the array "as a constant", `[] as const`. then it knows you want a tuple instead.

richardgirges: `as const` did it. But our linter isn't too happy about unused values. So at the moment this:
```typescript
const { setUserSession } = useSession();
```

has become this:
```typescript
  const [_userSession, _loading, setUserSession] = useSession();
```

Cool if I disable `@typescript-eslint/no-unused-vars` for this line?

src/scenes/Root.tsx
richardgirges: 👍

CarsonF:
@richardgirges I renamed UserSession to Session to try to be consistent with the API.
I'm also trying to take advantage of the apollo client/cache instead of managing the state ourselves.
Hope you don't mind!

richardgirges:
@CarsonF not at all, looks great man!

PR #108, opened 2020-04-23
Move help icon

blakekrammes:
My only concern is that it might not be immediately apparent that you need to scroll down to see the rest of the languages in the list on Plan. If you do scroll all the way down now though, the help icon won't block the last one.

### DIFF
--- a/src/app/internships/internship-engagements/internship-engagements.component.scss
+++ b/src/app/internships/internship-engagements/internship-engagements.component.scss
@@ -12,6 +12,7 @@
   @include mat-elevation(3);
   z-index: 1;
   background: mat-background($theme, card);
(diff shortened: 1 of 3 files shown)


blakekrammes:
I thought about moving it right, but it blocked other things on different views. I can look at the internship engagements.

blakekrammes:
Do I need to rebase on master?

CarsonF:
I changed the base branch to `develop` this is where beta is deployed from. Can you rebase on develop?

blakekrammes:
Yeah. I can do that. I guess the problem is I created this fix branch from master?

blakekrammes:
All is well?

PR #122, opened 2020-04-27
created org-scenes

aaronkchsu:
## created org-scenes
* Organization name created by a user is unique? We are creating org node even though org name requested is same with last one internally. I think we need validation for this in server.

### DIFF
--- a/src/scenes/Organizations/Create/CreateOrganization.tsx
+++ b/src/scenes/Organizations/Create/CreateOrganization.tsx
@@ -1,6 +1,12 @@
-import React from 'react';
+import { IconButton } from '@material-ui/core';
+import { Close } from '@material-ui/icons';
+import { FORM_ERROR } from 'final-form';
+import { useSnackbar } from 'notistack';
+import React, { useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
 import { Except } from 'type-fest';
-import { useCreateOrganizationMutation } from '../../../api';
+import { handleFormError, useCreateOrganizationMutation } from '../../../api';
+import { useSession } from '../../../components/Session';
 import {
   CreateOrganizationForm,
   CreateOrganizationFormProps as Props,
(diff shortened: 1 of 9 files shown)


src/scenes/Organizations/Create/CreateOrganization.tsx
```diff
@@ -0,0 +1,53 @@
...
+            </IconButton>
+          ),
+        }
+      );
+    } catch (e) {
+      return await handleFormError(e, {
+        Default: {
+          [FORM_ERROR]: `Something wasn't right. Try again.`,
+        },
+      });
```
CarsonF, line 48: Error handlers can just return a string to indicate the same thing.
```diff
- Default: {
-   [FORM_ERROR]: `Something wasn't right. Try again.`,
- },
+ Default: `Something wasn't right. Try again.`,
```
However, I'm also thinking we should probably not copy/paste this generic error string everywhere. Since we aren't handling a specific case here let's just leave it off. Currently that means that the server message will be used. Later if we want to replace that with a generic message we can do that in the `handleFormErrors` function.
```suggestion
      return await handleFormError(e);
```

I know this is copied from the pattern in Authentication. It's a little unfortunate that we started with that, since because of the nature of it, it breaks the rules.

src/scenes/Organizations/Create/CreateOrganization.tsx
```diff
@@ -0,0 +1,53 @@
...
+    if (!sessionLoading && !session) {
+      navigate('/login');
+    }
+  }, [navigate, session, sessionLoading]);
+
+  const submit: Props['onSubmit'] = async (input) => {
+    try {
+      await createOrg({
+        variables: { input },
+      });
```
CarsonF, line 31: We'll also want to navigate to the new org after it's created, but that can wait

src/scenes/Organizations/Create/CreateOrganization.tsx
aaronkchsu: ok, got it.

src/scenes/Organizations/Create/CreateOrganization.tsx
aaronkchsu: ok

Issue #97, opened 2020-04-20
Sensitivity Component


CarsonF:
![Screen Shot 2020-04-20 at 11 06 36 AM](https://user-images.githubusercontent.com/932566/79773427-12132900-82f7-11ea-9cc8-c2c1d48830e0.png)

- Should accept a `value` prop that's a `Sensitivity`.
- Should have a storybook file / example (use a knob for value).
- Colors can be hardcoded in this component since they are specific to this component. 

Details & svg images in [Figma](https://www.figma.com/file/IfEmcJeuNY6ucgjYctlKB7/Cord-Field?node-id=1524%3A14).

PR #125, opened 2020-04-28
Project list item card

jmc637:


jmc637:
ListItemCard

jmc637:
Just using default image in public folder for now

jmc637:
I was looking at Figma and realized I need to make it so color of chip changes with sensitivity type. Also some of the sizing and margins are off. Will fix that today.

CarsonF:
@rayjay00 is actually working on the sensitivity component #97 which will be used here. Should just be able to be dropped in, so I wouldn't work about it.

I'll review more after lunch 😄

jmc637:
Okay nice, won't worry about that then

jmc637:
Margin and sizing are fixed according to figma diagram. Left sensitivity component spacing alone since it's going to be replaced.
ListItemCardV2

src/components/ProjectListItemCard/ProjectListItemCard.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import { number, select, text } from '@storybook/addon-knobs';
+import React from 'react';
+import { ProjectStatus } from '../../api';
+import { ProjectListItemCard } from './ProjectListItemCard';
+
+export default { title: 'Components/ProjectListItemCard' };
+
+const statusOptions: string[] & ProjectStatus[] = [
```
CarsonF, line 8: This should be fine...?
```suggestion
const statusOptions: ProjectStatus[] = [
```

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+  Chip,
+  makeStyles,
+  Typography,
+} from '@material-ui/core';
+import { VerifiedUserOutlined } from '@material-ui/icons';
+import { FC } from 'react';
+import * as React from 'react';
+import { ProjectStatus, Sensitivity } from '../../api';
+
+const useStyles = makeStyles({
```
CarsonF, line 15: These styles look good. For the `padding` & `margin` though let's use the `spacing` function from the theme.
It keeps everything on an 8px grid and is a bit easier for us with those rails.

Check out some examples in [LoginForm](https://github.com/SeedCompany/cord-field/blob/v3/src/scenes/Authentication/Login/LoginForm.tsx#L31-L69)

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+          <div className={leftContent}>
+            <Typography variant="h4">{name}</Typography>
+            <Typography color="primary" variant="body2">
+              <Box
+                component="span"
+                m={1}
+                display="inline"
+                color="text.secondary"
+              >
+                18397
```
CarsonF, line 117: This will be some kind of ID for the project and needs to be passed in

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+            </Typography>
+          </div>
+          <div className={rightContent}>
+            <Typography variant="h1">{numberOfLanguageEngagements}</Typography>
+            <Typography variant="body2" color="primary">
+              Languages
+            </Typography>
+            <Typography variant="body2" color="primary">
+              Engagements
+            </Typography>
```
CarsonF, line 153: This is one label, we should combine these. If the line break is important, use a `<br />`.

src/components/ProjectListItemCard/ProjectListItemCard.tsx
CarsonF: Also I know these won't match figma exactly, so round to the nearest value.

jmc637:
I'm going to close this because I just rebased and now the branches conflict. I'm delete the branch and repush the rebased one.

CarsonF:
There's no to start over. Rebase, fix the conflicts, and force push. GitHub will pick up the change just fine. For next time :)

CarsonF:
Re-upped in #135

PR #137, opened 2020-04-29
Sensitivity component

rayjay00:
Addresses issue #97 

This PR:
 * Adds sensitivity component and related story
 * Adds the component to `ProjectListItemCard`
 * **Added some styles/theme color configs** - This is where I'd like some feedback on this PR - primarily with how to add colors/reference them and any other best practices. See my comments in the code. Also, I'm not super familiar with TypeScript so extra eyes are appreciated. 👍 

**Screens**
image
image
image

### DIFF
--- a/src/components/Sensitivity/Sensitivity.tsx
+++ b/src/components/Sensitivity/Sensitivity.tsx
@@ -12,8 +12,7 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
     marginBottom: spacing(),
   },
   icon: {
-    height: spacing(2),
-    width: spacing(2),
+    fontSize: 16,
     color: palette.text.secondary,
     marginRight: spacing(1),
   },
--- a/src/components/Sensitivity/Sensitivity.tsx
+++ b/src/components/Sensitivity/Sensitivity.tsx
@@ -21,13 +21,13 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
     borderRadius: 4,
     color: 'white',
   },
-  chipLow: {
+  Low: {
     backgroundColor: colors.grey[400],
   },
-  chipMedium: {
+  Medium: {
     backgroundColor: palette.warning.main,
   },
-  chipHigh: {
+  High: {
     backgroundColor: palette.error.main,
   },
 }));
@@ -39,20 +39,17 @@ export interface SensitivityProps {
 export const Sensitivity: FC<SensitivityProps> = ({ value }) => {
   const classes = useStyles();
 
-  const chipClass = clsx({
-    [classes.chip]: true,
-    [classes.chipHigh]: value === 'High',
-    [classes.chipMedium]: value === 'Medium',
-    [classes.chipLow]: value === 'Low',
-  });
-
   return (
     <div>
       <div className={classes.iconWrapper}>
         <VerifiedUserOutlined className={classes.icon} />
         <Typography variant="body2">Sensitivity</Typography>
       </div>
-      <Chip className={chipClass} size="small" label={value} />
+      <Chip
+        className={clsx(classes.chip, classes[value])}
+        size="small"
+        label={value}
+      />
     </div>
   );
 };
--- a/src/components/Sensitivity/Sensitivity.tsx
+++ b/src/components/Sensitivity/Sensitivity.tsx
@@ -1,0 +1,58 @@
+import { Chip, colors, makeStyles, Typography } from '@material-ui/core';
+import { VerifiedUserOutlined } from '@material-ui/icons';
+import clsx from 'clsx';
+import { FC } from 'react';
+import * as React from 'react';
+import { Sensitivity as SensitivityType } from '../../api';
+
+const useStyles = makeStyles(({ palette, spacing }) => ({
+  iconWrapper: {
+    display: 'flex',
+    alignItems: 'center',
+    marginBottom: spacing(),
+  },
+  icon: {
+    height: spacing(2),
+    width: spacing(2),
+    color: palette.text.secondary,
+    marginRight: spacing(1),
+  },
+  chip: {
+    borderRadius: 4,
+    color: 'white',
+  },
+  chipLow: {
+    backgroundColor: colors.grey[400],
+  },
+  chipMedium: {
+    backgroundColor: palette.warning.main,
+  },
+  chipHigh: {
+    backgroundColor: palette.error.main,
+  },
+}));
+
+export interface SensitivityProps {
+  value: SensitivityType;
+}
+
+export const Sensitivity: FC<SensitivityProps> = ({ value }) => {
+  const classes = useStyles();
+
+  const chipClass = clsx({
+    [classes.chip]: true,
+    [classes.chipHigh]: value === 'High',
+    [classes.chipMedium]: value === 'Medium',
+    [classes.chipLow]: value === 'Low',
+  });
+
+  return (
+    <div>
+      <div className={classes.iconWrapper}>
+        <VerifiedUserOutlined className={classes.icon} />
+        <Typography variant="body2">Sensitivity</Typography>
+      </div>
+      <Chip className={chipClass} size="small" label={value} />
+    </div>
+  );
+};
--- a/src/components/Sensitivity/Sensitivity.tsx
+++ b/src/components/Sensitivity/Sensitivity.tsx
@@ -14,7 +14,7 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
   icon: {
     fontSize: 16,
     color: palette.text.secondary,
-    marginRight: spacing(1),
+    marginRight: 2,
   },
   chip: {
     borderRadius: 4,
--- a/src/components/ProjectListItemCard/ProjectListItemCard.tsx
+++ b/src/components/ProjectListItemCard/ProjectListItemCard.tsx
@@ -3,14 +3,13 @@ import {
   Card,
   CardContent,
   CardMedia,
-  Chip,
   makeStyles,
   Typography,
 } from '@material-ui/core';
-import { VerifiedUserOutlined } from '@material-ui/icons';
 import { FC } from 'react';
 import * as React from 'react';
-import { ProjectStatus, Sensitivity } from '../../api';
+import { ProjectStatus, Sensitivity as SensitivityType } from '../../api';
+import { Sensitivity } from '../Sensitivity';
 
 const useStyles = makeStyles(({ spacing }) => ({
   card: {
@@ -46,12 +45,6 @@ const useStyles = makeStyles(({ spacing }) => ({
     alignItems: 'flex-end',
     justifyContent: 'flex-end',
   },
-  sensitivityIcon: {
-    margin: spacing(0, 1),
-  },
-  sensitivityChip: {
-    width: '75px',
-  },
   statusLabel: {
     marginTop: 'auto',
   },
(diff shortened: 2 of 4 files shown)


src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,55 @@
...
+      backgroundColor: ({ value }) => {
+        if (value === 'High') {
+          return secondary.main;
+        }
+
+        if (value === 'Medium') {
+          return warning.main;
+        }
+
+        return '#8f928b';
```
rayjay00, line 31: This felt a little weird to add this color here. I need this light grey color for when the button says `Low` - is there a best practice to pull this in from the theme? It's used in a different spot, too, so I made a variable but couldn't quite pin down how to reference it. Any guidance is appreciated here!

src/theme/index.ts
```diff
@@ -23,14 +25,18 @@ const createPalette = (options: PaletteOptions) => {
     },
     secondary: {
       main: '#ff5a5f',
+      dark: midGrey,
+    },
+    warning: {
+      main: '#F2994A',
```
rayjay00, line 31: This is the orange color for when the label is `Medium`.

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const Sensitivity: FC<SensitivityProps> = (props) => {
+  const { icon, iconWrapper, chip } = useStyles(props);
+
+  return (
+    <Box>
+      <Box className={iconWrapper}>
+        <VerifiedUserOutlined className={icon} color="disabled" />
+        <Typography variant="body2" color="textSecondary">
```
rayjay00, line 48: The color of this should be `#3C444E`. What's the best way to add that in to the theme config? `textSecondary` is a bit too light 🤷

rayjay00:
Seeing this in the CI - will resolve: 
image

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,55 @@
...
+    },
+    iconWrapper: {
+      display: 'flex',
+      alignItems: 'center',
+      marginBottom: spacing(),
+    },
+    chip: {
+      borderRadius: '4px',
+      color: 'white',
+      backgroundColor: ({ value }: { value: string }) => {
```
CarsonF, line 22: Ok so using property functions like this actually creates a new css class for each component that's mounted. Because of this I've come to the opinion that it's just easier to use inline `styles` instead of this. It's also easier on TypeScript, as you've seen a little bit of 😄

src/components/Sensitivity/Sensitivity.tsx
CarsonF: This is use case is a little tricky.
An argument could be made that these colors should just be defined here and not pulled from the theme. For example, we don't want to couple _High_ to our `secondary` color palette. That color could change and that doesn't mean we'd want this usage to change.
However, we also have `error` and `warning` definitions as well, which probably do mean the same thing as what we are trying to communicate here.

So it's more about mapping the intention of the theme option to the usage, rather than the color. If we can find a gray that makes sense in the theme and can be changed to the gray in designs and it still work across the app then let's do that. Otherwise we can just hardcode it here.

I'll dig more.

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,55 @@
+import { Box, Chip, makeStyles, Typography } from '@material-ui/core';
+import { VerifiedUserOutlined } from '@material-ui/icons';
+import { FC } from 'react';
+import * as React from 'react';
+import { Sensitivity as SensitivityType } from '../../api';
```
CarsonF, line 5: Hmm this is probably bad DX since we would be constantly fighting which import - the component or the type.
Since the type name is auto-generated, it would be better if we could come up with a different name for the component. Any ideas?

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,55 @@
...
+  value: SensitivityType;
+}
+
+export const Sensitivity: FC<SensitivityProps> = (props) => {
+  const { icon, iconWrapper, chip } = useStyles(props);
+
+  return (
+    <Box>
+      <Box className={iconWrapper}>
+        <VerifiedUserOutlined className={icon} color="disabled" />
```
CarsonF, line 47: Bringing up intent again, we aren't trying to communicate that this icon is disabled. Rather it's just a gray color.
It's possible that it matches our `textSecondary` color which would be a better fit.

src/components/Sensitivity/Sensitivity.tsx
CarsonF: Based on the designs it looks like it's the primary text color.
That's defined here https://github.com/SeedCompany/cord-field/blob/d6e89115d4f8cf37cb75a69e47d273a9a1cbd0e8/src/theme/index.ts#L38
The color probably just changed in designs since I last looked.
I believe that's the default color for `Typography`, but don't take my word for it.

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,55 @@
...
+export interface SensitivityProps {
+  value: SensitivityType;
+}
+
+export const Sensitivity: FC<SensitivityProps> = (props) => {
+  const { icon, iconWrapper, chip } = useStyles(props);
+
+  return (
+    <Box>
+      <Box className={iconWrapper}>
```
CarsonF, line 46: The `Box` component also defines a new css class for each component mounted, because of that I try to avoid them. Esp in cases where there could be lots of instance of the component on screen at once, like this.

So just replace with `<div>`.

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,55 @@
...
+      width: spacing(2),
+      marginRight: spacing(0.5),
+    },
+    iconWrapper: {
+      display: 'flex',
+      alignItems: 'center',
+      marginBottom: spacing(),
+    },
+    chip: {
+      borderRadius: '4px',
```
CarsonF, line 20: Let's keep an eye out. If other "chip" designs also have 4px radius then we should just move this to the theme.

src/theme/index.ts
```diff
@@ -23,14 +25,18 @@ const createPalette = (options: PaletteOptions) => {
     },
     secondary: {
       main: '#ff5a5f',
+      dark: midGrey,
```
CarsonF, line 28: This doesn't make sense. light & dark should be different _shades_ of the same _hue_.

src/theme/index.ts
CarsonF: I'm cool with this. But I want to double check with Patrick that he envisions this color for all warning like usages.

src/components/Sensitivity/Sensitivity.tsx
rayjay00: Agreed - thinking on it!

src/components/Sensitivity/Sensitivity.tsx
rayjay00: That is no bueno (but good to know)! I'm going inline.

src/components/Sensitivity/Sensitivity.tsx
rayjay00: 10-4

src/theme/index.ts
rayjay00: yeah this is an oxymoron! Thanks - will update to what it was before.

src/theme/index.ts
rayjay00: Ok sure! Happy to check in with Patrick on this.

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,55 @@
...
+    },
+  })
+);
+
+export interface SensitivityProps {
+  value: SensitivityType;
+}
+
+export const Sensitivity: FC<SensitivityProps> = (props) => {
+  const { icon, iconWrapper, chip } = useStyles(props);
```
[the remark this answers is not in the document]
rayjay00: I'll make this consistent with your way of doing it 👍

src/components/Sensitivity/Sensitivity.tsx
rayjay00: `SensitivityStatus`? 🤷

rayjay00:
Ready for re-review. Just pushed a bunch of commits addressing all comments except those which will need Patrick's input and renaming the component (`SensitivityStatus` 🤷). Rather than styling inline, I went for adding a class conditionally since the colors were already in scope of the css. Let me know your thoughts. 👍 @CarsonF

src/components/Sensitivity/Sensitivity.tsx
CarsonF: Let's punt lol

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const chipClass =
+    value === 'High'
+      ? classes.chipHigh
+      : value === 'Medium'
+      ? classes.chipMedium
+      : classes.chipLow;
+
+  return (
+    <div>
+      <Box className={classes.iconWrapper}>
```
CarsonF, line 55:
```suggestion
      <div className={classes.iconWrapper}>
```

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const Sensitivity: FC<SensitivityProps> = ({ value }) => {
+  const classes = useStyles();
+
+  const chipClass =
+    value === 'High'
+      ? classes.chipHigh
+      : value === 'Medium'
+      ? classes.chipMedium
+      : classes.chipLow;
```
CarsonF, line 51: There's a handy library that makes this kinda thing easier. 
```suggestion
  const chipClass = clsx({
    [classes.chip]: true,
    [classes.chipHigh]: value === 'High',
    [classes.chipMedium]: value === 'Medium',
    [classes.chipLow]: value === 'Low',
  });
```
Hmm same amount of lines. But another point I wanted to make is that the common chip class can be still be separated and multiple classes used.

rayjay00: Cool - I’ve this with the classNames package before - seems similar. Will add!

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -43,19 +40,19 @@ export interface SensitivityProps {
...
-      ? classes.chipHigh
-      : value === 'Medium'
-      ? classes.chipMedium
-      : classes.chipLow;
+  const chipClass = clsx({
+    [classes.chip]: true,
+    [classes.chipHigh]: value === 'High',
+    [classes.chipMedium]: value === 'Medium',
+    [classes.chipLow]: value === 'Low',
+  });
```
CarsonF, line 48: I thought of a better way to do this, if you want to tweak it again...
```suggestion
  const chipClass = clsx(classes.chip, classes[`chip${value}`]);
```

src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <div>
+      <div className={classes.iconWrapper}>
+        <VerifiedUserOutlined className={classes.icon} color="disabled" />
```
CarsonF, line 53: Change this to the secondary text color via css

PR #138, opened 2020-04-29
Form field stories

aaronkchsu:


src/components/form/CheckboxField.stories.tsx
```diff
@@ -0,0 +1,33 @@
+import { action } from '@storybook/addon-actions';
+import { boolean, select, text } from '@storybook/addon-knobs';
+import React from 'react';
+import { Form } from 'react-final-form';
+import { CheckboxField as CB } from './CheckboxField';
+
+export default { title: 'Components/CheckboxField' };
+
+export const CheckboxField = () => (
```
CarsonF, line 9: Could you scope all of these under `Forms`? Also if there's only one story I leave out the name in the title to prevent unnecessary nesting.

Actually maybe it would make sense to move Field there as well.

```suggestion
export default { title: 'Components/Forms/Fields' };

export const Checkbox = () => (
```

That way the UI shows
```
Components/Forms/Fields/Checkbox
                        Date
                        Email
                        ...
```

`FieldGroup` and `SubmitButton` shouldn't be under `Fields` though, just `Forms`

src/components/form/DateField.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+            name="DateField"
+            disabled={boolean('Disabled', false)}
+            fullWidth={boolean('Full Width', false)}
+            variant={select(
+              'Variant',
+              ['standard', 'outlined', 'filled'],
+              'standard'
+            )}
+            helperText={text('HelperText', 'DatePicker')}
+            initialValue={new Date().toISOString().slice(0, 10)}
```
CarsonF, line 26:
```suggestion
            initialValue={CalendarDate.local()}
```

src/components/form/DateField.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <form onSubmit={handleSubmit}>
+          <DF
+            name="DateField"
+            disabled={boolean('Disabled', false)}
+            fullWidth={boolean('Full Width', false)}
+            variant={select(
+              'Variant',
+              ['standard', 'outlined', 'filled'],
+              'standard'
+            )}
```
CarsonF, line 24: Let's remove `variant`s in all the stories. We default to `filled` variant in our theme and probably don't want that to be changed on a per use basis.

src/components/form/FieldGroup.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Form } from 'react-final-form';
+import { FieldGroup as FG } from './FieldGroup';
+
+export default { title: 'Components/FieldGroup' };
+
+export const FieldGroup = () => (
+  <Form onSubmit={action('onSubmit')}>
+    {({ handleSubmit }) => (
+      <form onSubmit={handleSubmit}>
+        <FG prefix="Prefix" children={text('Children', 'FieldGroup')} />
```
CarsonF, line 13: Something like this would be a better example:
```suggestion
        <FG prefix="user">
          <TextField name="firstName" ... />
          <TextField name="lastName" ... />
        </FG>
```

src/components/form/SubmitButton.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+            'Variant',
+            ['contained', 'text', 'outlined'],
+            'contained'
+          )}
+          fullWidth={boolean('Full Width', false)}
+          color={select(
+            'Color',
+            ['inherit', 'primary', 'secondary', 'default', 'error'],
+            'primary'
+          )}
```
CarsonF, line 26: Let's try to keep the knob defaults to the same defaults as the component.
https://github.com/SeedCompany/cord-field/blob/5fb7981fde2869c50c8c551b4f953da7bcaacae0/src/components/form/SubmitButton.tsx#L50-L53

src/components/form/DateField.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export default { title: 'Components/DateField' };
+
+export const DateField = () => (
+  <LocalizationProvider dateAdapter={LuxonUtils}>
+    <Form onSubmit={action('onSubmit')}>
+      {({ handleSubmit }) => (
+        <form onSubmit={handleSubmit}>
+          <DF
+            name="DateField"
```
CarsonF, line 17: `name` is the field name so it should always be _camelCase_. Ah appropriate name would be nice too, like `startDate` here.

PR #139, opened 2020-04-29
Register Scene

### LINKED ISSUE #80: Create Authentication Scenes

richardgirges:
Closes #80 

This PR contains:
* Register scene
* Register form
* Register story

### DIFF
--- a/src/api/graphql.generated.tsx
+++ b/src/api/graphql.generated.tsx
@@ -62,6 +62,11 @@ export interface AttachUserToSecurityGroupInput {
   request: AttachUserToSecurityGroup;
 }
 
+export interface BaseNodeConsistencyInput {
+  /** The BaseNode type */
+  baseNode: Scalars['String'];
+}
+
 export type BibleBook =
   | 'Genesis'
   | 'Exodus'
@@ -2016,7 +2021,7 @@ export interface Query {
   user: User;
   /** Look up users */
   users: UserListOutput;
-  /** Check out whether a provided email exists or not in User Table */
+  /** Checks whether a provided email already exists */
   checkEmail: Scalars['Boolean'];
   /** Check Consistency across User Nodes */
   checkUserConsistency: Scalars['Boolean'];
@@ -2079,6 +2084,8 @@ export interface Query {
   directory: Directory;
   file: File;
   fileNode: FileOrDirectory;
+  /** Check Consistency in File Nodes */
+  checkFileConsistency: Scalars['Boolean'];
   /** Look up all states on workflow */
   states: StateListOutput;
   /** Look up all next possible states on workflow */
@@ -2229,6 +2236,10 @@ export interface QueryFileArgs {
 
 export interface QueryFileNodeArgs {
   id: Scalars['ID'];
+}
+
+export interface QueryCheckFileConsistencyArgs {
+  input: BaseNodeConsistencyInput;
 }
 
 export interface QueryStatesArgs {
@@ -3318,10 +3329,6 @@ export interface UserEducationArgs {
   input?: Maybe<EducationListInput>;
 }
 
-export interface UserEmailInput {
-  email: Scalars['String'];
-}
-
 export interface UserFilters {
   /** Only users matching this first name */
   displayFirstName?: Maybe<Scalars['String']>;
@@ -3428,6 +3435,16 @@ export type LogoutMutation = { __typename?: 'Mutation' } & Pick<
   Mutation,
   'logout'
 >;
+
+export interface RegisterMutationVariables {
+  input: CreateUserInput;
+}
+
+export type RegisterMutation = { __typename?: 'Mutation' } & {
+  createUser: { __typename?: 'CreateUserOutput' } & {
+    user: { __typename?: 'User' } & LoggedInUserFragment;
+  };
+};
 
 export interface ResetPasswordMutationVariables {
   input: ResetPasswordInput;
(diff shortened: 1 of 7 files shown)


src/scenes/Authentication/Register/Register.tsx
```diff
@@ -0,0 +1,53 @@
...
+      setSuccess(true);
+      setUserSession(data.createUser.user);
+      const returnTo = decodeURIComponent(query.get('returnTo') ?? '/');
+      navigate(returnTo, { replace: true });
+    } catch (e) {
+      return await handleFormError(e, {
+        Default: {
+          [FORM_ERROR]: `Something wasn't right. Try again, or reset password.`,
+        },
+      });
```
CarsonF, line 48: Error handlers can just return a string to indicate the same thing.
```diff
- Default: {
-   [FORM_ERROR]: `Something wasn't right. Try again.`,
- },
+ Default: `Something wasn't right. Try again.`,
```
However, I'm also thinking we should probably not copy/paste this generic error string everywhere. Since we aren't handling a specific case here let's just leave it off. Currently that means that the server message will be used. Later if we want to replace that with a generic message we can do that in the `handleFormErrors` function.
```suggestion
      return await handleFormError(e);
```

src/scenes/Authentication/Register/RegisterForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+  state.formState = {
+    ...state.formState,
+    submitError: undefined,
+    submitErrors: undefined,
+  };
+};
+
+// decorators get re-created if array identity changes
+// so make constant outside of render function
+const decorators = [blurOnSubmit, focusLastActiveFieldOnSubmitError];
```
CarsonF, line 128: I think focusFirstFieldWithSubmitError would be better here.
```suggestion
const decorators = [blurOnSubmit, focusFirstFieldWithSubmitError];
```

src/scenes/Authentication/Register/RegisterForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+  );
+};
+
+const clearSubmitErrors: Mutator<RegisterInput> = (args, state) => {
+  state.formState = {
+    ...state.formState,
+    submitError: undefined,
+    submitErrors: undefined,
+  };
+};
```
CarsonF, line 124: This mutator isn't used/needed

src/scenes/Authentication/Register/RegisterForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+                autoFocus
+              />
+              <TextField
+                name="lastName"
+                label="Last Name"
+                placeholder="Enter Last Name"
+                required
+              />
+              <EmailField />
+              <PasswordField />
```
CarsonF, line 105: I think we probably need a second confirm password field. I've done the same thing in 7748eb91dc62aa40e25512e60bc3bf97052eac81. It's a lot of code to duplicate, but it's only once so I guess it's fine.

src/scenes/Authentication/Register/RegisterForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+    '@media (max-height: 700px)': {
+      display: 'none',
+    },
+  },
+  verse: {
+    fontStyle: 'italic',
+  },
+  reference: {
+    marginTop: spacing(1),
+  },
```
CarsonF, line 54:
```suggestion
```

src/scenes/Authentication/Register/queries.graphql
```diff
@@ -0,0 +1,7 @@
+mutation Register($input: CreateUserInput!) {
```
CarsonF, line 1: I'm thinking it would probably be better to name these files based on what's in them. i.e. `register.graphql`
It might make searching a bit easier...

src/scenes/Authentication/Register/Register.tsx
richardgirges: Perfect 👍

src/scenes/Authentication/Register/queries.graphql
richardgirges: This will make for a way better developer experience. Renaming to `register.graphql`

src/scenes/Authentication/Register/Register.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const query = useSearchParams();
+  const [session, sessionLoading, setUserSession] = useSession();
+  const [success, setSuccess] = useState(false);
+
+  // Redirect to homepage if already logged in (and not from successful login)
+  useEffect(() => {
+    if (!sessionLoading && session && !success) {
+      navigate('/', { replace: true });
+    }
+  }, [navigate, session, sessionLoading, success]);
```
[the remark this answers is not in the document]
richardgirges: You mean like a hard-load with `location.href` ?

src/scenes/Authentication/Register/RegisterForm.tsx
richardgirges: 10-4

src/scenes/Authentication/Register/Register.tsx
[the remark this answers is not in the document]
CarsonF: No, I mean we have this logic in login and now in register. I'm wondering if we just pull it out to Authentication and apply it to forgot/reset password too.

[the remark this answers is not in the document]
richardgirges: gotcha - how should we detect successful "login", "signup", and "reset password" actions in the `Authentication` component?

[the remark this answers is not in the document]
CarsonF: Idk lol. We can come back to it.

src/api/graphql.generated.tsx
```diff
@@ -1121,8 +1121,9 @@ export interface LoginInput {
```
CarsonF, line 1121: I think your schema file is outdated. This looks like it's reverting recent changes

richardgirges: Hmm, I just did a git pull on the master branch in cord-api-v3 but don't see any changes in the schema file o_O

CarsonF: DId you start the server?

richardgirges: yeah I restarted the server after git pull

CarsonF: Well I know `LoginOutput` doesn't have `success` anymore and you added it here...

richardgirges: Okay I deleted the `schema.graphql` file, restarted api server, and that seems to have updated the generated file. But now I'm getting this error:
```
TypeScript error in /Users/richard/repo/open/cord-field/src/scenes/Authentication/ForgotPassword/ForgotPasswordForm.tsx(4,10):
Module '"../../../../../../../../../Users/richard/repo/open/cord-field/src/api"' has no exported member 'UserEmailInput'.  TS2305

    2 | import React from 'react';
    3 | import { Form, FormProps } from 'react-final-form';
  > 4 | import { UserEmailInput } from '../../../api';
```

richardgirges: Can't seem to find any record of `UserEmailInput` in the server

CarsonF: Ah I just changed that....

You could just use `{ email: string}` instead.

richardgirges: 👍

PR #143, opened 2020-05-01
Sidebar component with svgs for sidebar logo and intership logo

jmc637:


jmc637:
Hey, I'm putting in this pr for an initial review. I had to import a svg for this but ended up making a svg component in react. I didn't want to use the picture component for this because it would get rendered as an img and I wouldn't be able to change the color by styling. Should there be a separate folder for react svg's or is the way I'm doing it okay?

CarsonF:
Yeah we don't need to use `Picture` for vector images.
Take a look at [`components/Icons/CordIcon`](https://github.com/SeedCompany/cord-field/blob/v3/src/components/Icons/CordIcon.tsx).

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,135 @@
+import {
```
CarsonF, line 1: I'd move this to `scenes/Root/Sidebar` - It's not reusable.

src/components/Sidebar/InternshipsSvg.tsx
```diff
@@ -0,0 +1,51 @@
+import { SvgIcon } from '@material-ui/core';
```
CarsonF, line 1: I'd put this in `components/Icons` as it is re-usable.

public/images/sidebar-icon.svg
```diff
@@ -0,0 +1,27 @@
+<svg width="248" height="136" viewBox="0 0 248 136" fill="none" xmlns="http://www.w3.org/2000/svg">
```
CarsonF, line 1: I'd move this right next to the sidebar component. The complication figures out static assets
The path can be imported like [this](https://github.com/SeedCompany/cord-field/blob/0d1b88608ba583fad4e9d7dcccec545c1664ebf1/src/scenes/Authentication/Authentication.tsx#L7)

----

I'm also seeing this rendered now too. It has the CORD Icon and some text as well as the background. I'm wondering if it would just be better to layer the cord icon and text on top to reduce this svg size and keep the text styling consistent with the theme (font, etc.).

src/components/Sidebar/Sidebar.tsx
jmc637: This is moved

src/components/Sidebar/InternshipsSvg.tsx
jmc637: This is moved to appropriate folder

public/images/sidebar-icon.svg
jmc637: I split this up so I can reuse the cord icon. Text and svg is layered on top of that background svg

jmc637:
Hey, I made the changes you suggested. Can you take a look when you get a chance

src/components/Icons/InternshipsIcon.tsx
```diff
@@ -0,0 +1,26 @@
...
+// Can't change svg styling without having it be an svg. Maybe should have a separate folder
+// in source for svg's. Can't import svg's outside of src due to create react app settings
+export const InternshipsIcon: FC<SvgIconProps> = (props) => (
+  <SvgIcon
+    viewBox="0 0 16 17"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+    {...props}
+  >
+    <g clipPath="url(#clip0)" strokeWidth=".5">
```
CarsonF, line 14: These two aren't doing anything.
```suggestion
    <g>
```

src/components/Icons/InternshipsIcon.tsx
```diff
@@ -0,0 +1,26 @@
...
+      <path d="M8.304 13.177a.284.284 0 01-.28-.262c-.213-3.5-3.671-5.787-3.705-5.81a.279.279 0 01-.082-.385.282.282 0 01.39-.08 10.33 10.33 0 011.89 1.678 7.64 7.64 0 012.07 4.563.277.277 0 01-.07.202.284.284 0 01-.195.093l-.018.001zM10.773 4.72a.283.283 0 01-.199-.082.277.277 0 01-.082-.197V2.347c0-.074.03-.144.082-.197a.284.284 0 01.482.197v2.094a.275.275 0 01-.174.258.284.284 0 01-.109.021zM12.891 4.023H11.48a.284.284 0 01-.2-.082.277.277 0 010-.395.284.284 0 01.2-.081h1.411a.277.277 0 110 .558z" />
+      <path d="M14.03 15.81H2.183c-.375 0-.733-.148-.998-.41a1.387 1.387 0 010-1.972c.265-.261.623-.408.998-.408H14.03a1.425 1.425 0 011.065.482.276.276 0 01-.028.393.285.285 0 01-.398-.028.848.848 0 00-.64-.29H2.184a.853.853 0 00-.557.268.834.834 0 000 1.139c.145.155.344.25.557.267H14.03c.074 0 .146.03.199.082a.278.278 0 01-.2.476z" />
+    </g>
+    <defs>
+      <clipPath id="clip0">
+        <path fill="#fff" d="M.134 0h16.065v16H.134z" />
+      </clipPath>
+    </defs>
```
CarsonF, line 24: This isn't needed either
```suggestion
```

src/scenes/Root/Sidebar/Sidebar.stories.tsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import { Sidebar } from './Sidebar';
+
+export default { title: 'Components/Sidebar' };
+
+export const SidebarStory = () => <Sidebar />;
```
CarsonF, line 6: This shows up as `Components/Sidebar/Sidebar Story`. The extra nesting is redundant since there's only one and we don't want _Story_ in the name. Also it's under Scenes now.

```suggestion
export default { title: 'Scenes/Root' };

export const Sidebar = () => <Sidebar />;
```

src/scenes/Root/Sidebar/SwooshIcon.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { FC } from 'react';
+
+const useStyles = makeStyles(() => ({
+  root: {
+    width: '248px',
+    height: '136px',
+  },
+}));
+
+export const SwooshIcon: FC<SvgIconProps> = ({
```
CarsonF, line 12: I know it's using SvgIcon, but it's not an Icon. `Swoosh` or `SwooshBackground` or something like that.

src/scenes/Root/Sidebar/SwooshIcon.tsx
```diff
@@ -0,0 +1,61 @@
+import { makeStyles, SvgIcon, SvgIconProps } from '@material-ui/core';
+import * as React from 'react';
+import { FC } from 'react';
+
+const useStyles = makeStyles(() => ({
+  root: {
+    width: '248px',
+    height: '136px',
```
CarsonF, line 8:
```suggestion
    width: '100%',
    height: 'initial',
```

src/scenes/Root/Sidebar/SwooshIcon.tsx
```diff
@@ -0,0 +1,61 @@
...
+}));
+
+export const SwooshIcon: FC<SvgIconProps> = ({
+  classes: classesProp,
+  ...rest
+}) => {
+  const classes = useStyles();
+  return (
+    <SvgIcon
+      classes={{ root: classes.root, ...classesProp }}
```
CarsonF, line 19: If `root` gets passed in it replaces our def which we don't want.
```suggestion
      classes={{ root: clsx(classes.root, classesProp.root), ...classesProp }}
```

src/scenes/Root/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,158 @@
...
+            className={createButton}
+            aria-controls="create-menu"
+            aria-haspopup="true"
+            variant="contained"
+            color="secondary"
+            startIcon={<Add />}
+            onClick={handleClick}
+          >
+            Create New Item
+          </Button>
```
CarsonF, line 108: Use `<ErrorButton>` instead. I know it's weird but let's go with it for now. Also you should use `fullWidth` prop instead of defining a pixel width.

src/scenes/Root/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const [anchorEl, setAnchorEl] = useState(null);
+
+  const handleClick = (e: any) => {
+    setAnchorEl(e.currentTarget);
+  };
+
+  const handleClose = () => setAnchorEl(null);
+
+  return (
+    <List className={list} component="nav" aria-label="sidebar">
```
CarsonF, line 88: I don't think it makes sense for the entire sidebar to be a list. Just the actual list section.

src/scenes/Root/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,158 @@
...
+        <ListItem className={menuHeaderListItem}>
+          <ListItemText
+            primaryTypographyProps={{
+              variant: 'body2',
+              className: menuHeaderText,
+            }}
+          >
+            MENU
+          </ListItemText>
+        </ListItem>
```
CarsonF, line 132: Have you considered using a `<ListSubheader>` for this? It might still need some tweaking but maybe less?

Example here https://material-ui.com/components/lists/#nested-list

src/scenes/Root/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+        <SidebarListLink
+          to="/Projects"
```
CarsonF, line 135: urls should always be lowercase (_kebab-case_)

src/scenes/Root/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,158 @@
...
+    padding: spacing(0, 2),
+  },
+  floating: {
+    position: 'absolute',
+    top: spacing(3),
+    left: spacing(4),
+  },
+  cordIcon: {
+    height: '45px',
+    width: '38px',
```
CarsonF, line 57: Use `fontSize` instead.

jmc637:
Hi I tested with changing to dark theme. The colors are close but not exactly what is in figma. Here's what it looks like

Screen Shot 2020-05-04 at 3 58 58 PM

Is it okay if the colors are slightly different?

jmc637:
PR changes are committed. Just commented colors out for the moment due to above question

src/components/Icons/InternshipsIcon.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { FC } from 'react';
+
+// Can't change svg styling without having it be an svg. Maybe should have a separate folder
+// in source for svg's. Can't import svg's outside of src due to create react app settings
+export const InternshipsIcon: FC<SvgIconProps> = (props) => (
+  <SvgIcon
+    viewBox="0 0 16 17"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+    {...props}
```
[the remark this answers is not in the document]
jmc637: Unnecessary attributes removed

src/components/Icons/InternshipsIcon.tsx
```diff
@@ -0,0 +1,26 @@
+import { SvgIcon, SvgIconProps } from '@material-ui/core';
+import * as React from 'react';
+import { FC } from 'react';
+
+// Can't change svg styling without having it be an svg. Maybe should have a separate folder
+// in source for svg's. Can't import svg's outside of src due to create react app settings
+export const InternshipsIcon: FC<SvgIconProps> = (props) => (
```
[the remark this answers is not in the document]
jmc637: I changed it to be PlantIcon in case someone else it has to use it for a different case

src/components/Icons/InternshipsIcon.tsx
jmc637: Removed

src/scenes/Root/Sidebar/Sidebar.stories.tsx
jmc637: Moved to correct location

CarsonF:
Cool. Let me take a look.

src/scenes/Root/Sidebar/SwooshIcon.tsx
jmc637: Changed to SwooshBackground

src/scenes/Root/Sidebar/SwooshIcon.tsx
jmc637: Changed

CarsonF:
Hey @jmc637. I was messing around with the colors. I ended up doing #150 which allows us more flexibility here with sidebar theme. However, I don't expect to need to create multiple (child) themes too often - only a few times maybe.

I ended up leaving the list item colors defaulted. There's lots of subtleties in the normal, hover, selected, & ripple colors that make it trickier and haven't been thought through in designs. We can come back later and tweak colors if needed.

I also split off the header section into a separate file, made some more styling changes, and connected the sidebar to `Root`.

Let me know what you think and we can merge in the AM.

src/scenes/Root/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const classes = useStyles();
+  const [anchorEl, setAnchorEl] = useState(null);
+
+  const handleClick = (e: any) => {
+    setAnchorEl(e.currentTarget);
+  };
+
+  const handleClose = () => setAnchorEl(null);
+
+  const createMenu = (
```
jmc637, line 45: Do you think we should we split these off into separate files?

src/scenes/Root/Sidebar/sidebar.theme.ts
```diff
@@ -0,0 +1,27 @@
+import { createTheme } from '../../../theme';
+
```
jmc637, line 2: When do we create a separate theme and override styles vs just overriding with classes like before? I'm not really partial to either way but wanted to know if there was some cutoff for when we should do one vs the other.

src/scenes/Root/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,112 @@
...
+      onClose={handleClose}
+    >
+      <MenuItem>
+        {/* Waiting on information for what goes in here */}
+        Project
+      </MenuItem>
+    </Menu>
+  );
+
+  const navList = (
```
jmc637, line 60: Same as above

src/scenes/Root/Sidebar/Sidebar.tsx
CarsonF: Yeah there's definitely a good chance we'll want to refactor this into a separate component. Let's wait and see an actual usage first.

src/scenes/Root/Sidebar/Sidebar.tsx
CarsonF: Same answer as above 😉

src/scenes/Root/Sidebar/sidebar.theme.ts
CarsonF: We talked on Zoom, but to summarize here:
- Basically never create a separate theme lol
- classes vs changing app theme is a judgement call. It's probably better to start with classes and then if we see a trend we can move it to theme.
  Layout should probably mostly just be in classes.
  Colors, fonts, etc should mostly come from theme.

PR #144, opened 2020-05-01
Typography story

aaronkchsu:


src/scenes/Authentication/Logout/Logout.tsx
```diff
@@ -11,10 +11,14 @@ export const Logout = () => {
...
-      navigate('/login', { replace: true });
-    });
+    logout()
+      .then(() => {
+        setCurrentUser(null);
+        navigate('/login', { replace: true });
+      })
+      .catch((e) => {
+        console.log(e);
+      });
```
CarsonF, line 21: What was going wrong with logout? Were you not getting a snackbar shown with the error?

src/components/MUITypography/Typography.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+  typography: {
+    fontFamily: 'Arial',
+    button: {
+      fontSize: 10,
+    },
+  },
+});
+
+export const Story = () => (
+  <ThemeProvider theme={theme}>
```
CarsonF, line 18: Storybook is already using our theme. No need to redefine it here, or change it.

src/components/MUITypography/Typography.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <Typography
+      variant={select(
+        'Variant',
+        ['body1', 'caption', 'h1', 'h5', 'subtitle1'],
+        'body1'
+      )}
```
CarsonF, line 24: It might be more useful to see all the typography variants next to each other so we can visually compare easily.
We should at least include h1-h4, body1/2, caption.

src/components/MUITypography/Typography.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+      color={select(
+        'Color',
+        ['primary', 'secondary', 'error', 'textPrimary'],
+        'secondary'
```
CarsonF, line 28:
```suggestion
        ['initial', 'inherit', 'primary', 'secondary', 'error', 'textPrimary', 'textSecondary'],
        'initial'
```

src/components/MUITypography/Typography.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+        'Align',
+        ['left', 'right', 'center', 'justify', 'inherit'],
+        'left'
+      )}
+    >
+      {text('Children', 'Children')}
+    </Typography>
+    <ProgressButton color="secondary" variant="contained">
+      ProgressButton
+    </ProgressButton>
```
CarsonF, line 40: `ProgressButton` already has a story
```suggestion
```

src/scenes/Authentication/Logout/Logout.tsx
aaronkchsu: The reason why I wrote this is because there is lint issue with this. would you recommend to add lint config for this?

CarsonF: What's the lint problem?

aaronkchsu: `14:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises  `

CarsonF: Ah. Run `yarn` and the updated config will be applied.

aaronkchsu: Ah, I didn't know config is already updated.

src/components/MUITypography/Typography.stories.tsx
aaronkchsu: Actually, I was gonna add special options. but i will keep to add all options for future stories. thanks

CarsonF:
I think there's still some confusion. I'm looking for something like this: https://material-ui.com/components/typography/#component

aaronkchsu:
> I think there's still some confusion. I'm looking for something like this: https://material-ui.com/components/typography/#component

got it, I will update with more use-cases

src/components/MUITypography/Typography.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+  'button',
+  'overline',
+];
+
+const VariantValues: string[] = [
+  'Heading.',
+  'Lorem ipsum dolor sit amet, consectetur adipisicing elit. Quos blanditiis tenetur.',
+];
+
+export const Story = () => (
```
CarsonF, line 28: - We don't want _Story_ in the name of the story - it's redundant. Let's call it `Typography` instead. We'll need to rename the actual component so that will work.
- We also don't want to needlessly nest, since there's only one story here we don't need `/Typography` in the title.

PR #148, opened 2020-05-04
Virtualized Lists

### LINKED ISSUE #95: Infinite Scroll & Virtualized Lists
# Needs
* windowing lol
* lazy loading / pagination / infinite scroll
* List needs auto size to fill parent
* Window scrolling - maybe? Since this is an app we could just scroll with in a section
* No need for grids

# Options
1. react-virtualized
1. react-window + helper libraries
1. react-window with hook for scrolling (prebuilt or custom)
  * [react-infinite-scroll-hook](https://github.com/onderonur/react-infinite-scroll-hook) uses intervals to check for more should be loaded :-1: 
  * I'd rather use a hook based on `IntersectionObserver` to be more performant. We already have `react-intersection-observer` imported, but maybe this doesn't fit the need.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466802153) by [Unito](https://www.unito.io)

richardgirges:
Closes #95 

Still in draft mode, need feedback


richardgirges:
Primary issues I'm facing in this PR right now: 

- scrolling is buggy
- List doesn't expand to the full height of the screen

CarsonF:
> * List doesn't expand to the full height of the screen

Need to wrap the whole thing and flex it. Like [here](https://github.com/SeedCompany/website-v3/blob/e43c14597459e80da7be77636e969247e59b82a8/src/components/Shared/Common/CardGridList.tsx#L19-L30).

> * scrolling is buggy

Scrolling is always a bit like that with windowing. Probably just need to tweak the overscan options.

richardgirges:
Awesome so the flex wrapping worked for problem #1. 

Scrolling is still a little iffy though. I upped the overscan rows to 100 before I was able to make it feel seamless. But 100 rows is way too much and on top of that, scrolling backward yields jagged/weird results. I can't help but think there's something else missing here.

CarsonF:
I'm going to wait for #153 to be merged in and then pick this up.

PR #149, opened 2020-05-04
Person card added

jmc637:


src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+  },
+  cardActions: {
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'center',
+    padding: spacing(2),
+    border: '0.5px solid #e0e0e0',
+  },
+  activeColor: {
+    backgroundColor: palette.primary.main,
```
CarsonF, line 54: Let's hardcode `#619A55` here. We always think of "active" as green, regardless of our theme colors.

src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+  },
+  notActiveColor: {
+    // TODO not sure exactly what color this should be
+    backgroundColor: palette.error.main,
```
CarsonF, line 58: A "disabled" color from our theme would probably be better.

src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+    personImage,
+    roleName,
+    organizationName,
+    cardActions,
+    activeColor,
+    notActiveColor,
+  } = useStyles();
+
+  console.log('ACTIVE: ', active, activeColor);
+
```
CarsonF, line 93:
```suggestion

```

src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+  return (
+    <Card className={personCard}>
+      <CardContent className={cardContent}>
+        <Picture className={personImage} source={imageSource} />
+        <Typography
+          color="textPrimary"
+          variant="subtitle2"
```
CarsonF, line 100:
```suggestion
          variant="body2"
```

src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+              [activeColor]: active,
+              [notActiveColor]: !active,
+            })}
+          />
+          {/* TODO does this text chage when inactive */}
+          Active
+        </Typography>
+      </CardContent>
+      <CardActions className={cardActions}>
+        <Link external={false} to={personRoute}>
```
CarsonF, line 132: `external` defaults to false, no need to include it. Also let's use a button here instead.
```suggestion
        <ButtonLink to={personRoute}>
```

src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <Typography color="textSecondary" variant="caption">
+          <Box
+            component="span"
+            className={clsx(statusCircle, {
+              [activeColor]: active,
+              [notActiveColor]: !active,
+            })}
+          />
+          {/* TODO does this text chage when inactive */}
+          Active
```
CarsonF, line 128: Yeah I'm going to guess "Inactive"

src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <Typography
+          color="textSecondary"
+          variant="caption"
+          className={roleName}
+        >
+          {role}
+        </Typography>
+        <Typography color="textSecondary" variant="caption">
+          <Box
+            component="span"
```
CarsonF, line 121: The `Box` component actually creates a new css class for every element that's mounted. I try to avoid it for that reason. Especially for list items which will have lots on page at once.
Here we aren't even using the benefits of the component since we haven't used any of the "styling props"
(i.e. `mr={1}`)
```suggestion
          <span
```

src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+}));
+
+export interface PersonCardProps {
+  name: string;
+  organization: string;
+  role: string;
+  active: boolean;
+  personRoute: string;
+  imageSource: string;
+}
```
CarsonF, line 69: This will probably end up getting a single person object prop instead of all of these individually. I'm fine with this for now though.

src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+    personName,
+    statusCircle,
+    cardContent,
+    personImage,
+    roleName,
+    organizationName,
+    cardActions,
+    activeColor,
+    notActiveColor,
+  } = useStyles();
```
CarsonF, line 90: I've held off on commenting on this until now, because it feels petty. But typically we don't de-structure the css classes. It helps to not populate the variable namespace and make the quick distinction between them and props.
It does feel arbitrary though since we do destructure props. The only thing I can say is it would probably help to be consistent across the codebase.

src/components/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+    marginRight: spacing(1),
+    borderRadius: '50%',
+    display: 'inline-block',
+  },
+  cardActions: {
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'center',
+    padding: spacing(2),
+    border: '0.5px solid #e0e0e0',
```
CarsonF, line 51:
```suggestion
    border: `0.5px solid ${colors.grey[300]}`,
```
Does a half pixel even work? I think browsers just round up, right?

jmc637:
PR Changes are in for thiis

PR #164, opened 2020-05-07
Simple project list

aaronkchsu:


src/scenes/Projects/queries.graphql
```diff
@@ -0,0 +1,68 @@
+query projects($input: ProjectListInput!) {
```
CarsonF, line 1:
```suggestion
query Projects($input: ProjectListInput!) {
```

src/scenes/Projects/queries.graphql
```diff
@@ -0,0 +1,68 @@
...
+  projects(input: $input) {
+    ...projects
+  }
+}
+
+fragment projects on ProjectListOutput {
+  items {
```
CarsonF, line 8: What's really needed here is a fragment for the item in the list. We will only fetch a limited number of properties for each project when listing to keep the data down. 

```graphql
fragment ProjectListItem on Project {
  ...
}
```
The interface generated from this, `ProjectListItemFragment` can be used on the `ProjectListItemCard` instead of doing a `Partial<Project>`. The former accurately reflects the properties available, where the latter is a total guess.


With that fragment in place the query is only a couple lines making a fragment not really worth it.
```graphql
projects(input: $input) {
  items {
    ...ProjectListItem
  }
  total
  hasMore
}
```

src/scenes/Projects/queries.graphql
CarsonF, line 1: Also I want to stop using "queries" for the filename. Let's make these files match what they are querying. `projects.graphql` is fine. That lets us go-to/open the file easier with our editors.

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -53,28 +53,16 @@ const useStyles = makeStyles(({ spacing }) => ({
...
-  name: string;
-  countryName: string;
-  region: string;
-  sensitivity: SensitivityType;
-  status: ProjectStatus;
-  numberOfLanguageEngagements: number;
-  esadDate: string;
-}
+export type ProjectListItemCardProps = Partial<Project> & {
+  projectImagePath?: string; // where is this comming?
```
CarsonF, line 57: Projects don't currently have images, and probably won't for the initial release. We should just hard code a placeholder for now.

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -88,60 +76,72 @@ export const ProjectListItemCard: FC<ProjectListItemCardProps> = ({
...
-        <CardContent className={cardContent}>
-          <div className={leftContent}>
-            <Typography variant="h4">{name}</Typography>
-            <Typography color="primary" variant="body2">
-              <Box
-                component="span"
-                m={1}
-                display="inline"
-                color="text.secondary"
+    <div className={className}>
```
CarsonF, line 79: I don't think an extra div should be used here. Either apply the className to Card, or wrap this component with the div where it is used. Separation of concerns.

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -53,28 +53,16 @@ const useStyles = makeStyles(({ spacing }) => ({
...
-  countryName: string;
-  region: string;
-  sensitivity: SensitivityType;
-  status: ProjectStatus;
-  numberOfLanguageEngagements: number;
-  esadDate: string;
-}
+export type ProjectListItemCardProps = Partial<Project> & {
+  projectImagePath?: string; // where is this comming?
+  numberOfLanguageEngagements?: number; // where is this comming?
```
CarsonF, line 58: Ok so `projects()` returns a list of `Project`s, which is an interface. There are two concrete types implementing this interface, `InternshipProject` and `TranslationProject`. Both of these concrete types have an `engagements` property, which is a list that has a total count on it.  We can use this syntax below to access their properties.

```graphql
items {
  ... on InternshipProject {
    engagements {
      total
    }
  }
  ... on TranslationProject {
    engagements {
      total
    }
  }
}
```
That ^ can be put on the Project fragment mentioned in my other comment along with the other properties that are shared on the Project interface.

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -137,7 +124,7 @@ export const ProjectListItemCard: FC<ProjectListItemCardProps> = ({
               <Box component="span" m={1} display="inline" color="text.primary">
                 ESAD:
               </Box>
-              {esadDate}
+              {props.estimatedSubmission?.value}
```
CarsonF, line 122: We should hide the label if there is no value. Also let's not use `Box` here for performance reasons.

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -92,24 +75,25 @@ export const ProjectListItemCard: FC<ProjectListItemCardProps> = ({
...
                 m={1}
                 display="inline"
                 color="text.secondary"
               >
-                {id}
+                {props.id}
               </Box>
-              {countryName}, {region}
+              {props.location?.value?.name.value},{' '}
+              {props.location?.value?.region.value?.name.value}
```
CarsonF, line 94: This needs to be a bit smarter. Right now it's just showing a comma when there is no location. A helper function could be useful for this as well since we'll probably need to display locations in multiple areas

```tsx
const displayLocation = (location: Location /* From API, not global */) => '';
```

src/scenes/Projects/projects.graphql
```diff
@@ -0,0 +1,71 @@
...
+    total
+  }
+}
+
+fragment ProjectListItem on Project {
+  id
+  createdAt
+  name {
+    value
+  }
```
CarsonF, line 16:
```suggestion
```
name is duplicated

src/scenes/Projects/projects.graphql
```diff
@@ -0,0 +1,71 @@
...
+        }
+      }
+    }
+  }
+  mouStart {
+    value
+  }
+  mouEnd {
+    value
+  }
```
CarsonF, line 50:
```suggestion
```
Since we aren't showing these on the ListItemCard we should leave it out

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -107,13 +107,8 @@ export const ProjectListItemCard: FC<ProjectListItemCardProps> = ({
...
           <div className={rightContent}>
-            <Typography variant="h1">
-              {0}{' '}
-              {/* Should be updated with implementation of engagements for project. Graphql query is also pending for this. */}
-            </Typography>
+            <Typography variant="h1">{props.engagements.total}</Typography>
             <Typography variant="body2" color="primary" align="right">
-              Languages
-              <br />
               Engagements
```
CarsonF, line 112: We actually need to display "Language Engagements" or "Internship Engagements" based on the `type`.

src/scenes/Projects/ProjectList.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <div className={classes.option}>
+        <Card className={classes.optionCard} variant="outlined">
+          Sort Options
+        </Card>
+        <Card className={classes.optionCard} variant="outlined">
+          Filter Options
+        </Card>
+      </div>
+      <div className={classes.content}>
+        <Typography variant="h3">{data?.projects.total} Projects</Typography>
```
CarsonF, line 56: Could we get some spacing after this? Maybe using `gutterBottom` or `paragraph`?

src/scenes/Projects/ProjectList.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <div className={classes.root}>
+      <Typography variant="h2">My Projects</Typography>
+      <div className={classes.option}>
+        <Card className={classes.optionCard} variant="outlined">
+          Sort Options
+        </Card>
+        <Card className={classes.optionCard} variant="outlined">
+          Filter Options
+        </Card>
```
CarsonF, line 53: Could we get 1 spacing margin between these options?

Issue #141, opened 2020-04-30
Top Page Header


CarsonF:
![Screen Shot 2020-04-30 at 6 37 51 PM](https://user-images.githubusercontent.com/932566/80769134-de63aa80-8b11-11ea-8c71-74429161d29c.png)

All functionality can be stubbed other than:
- Displaying actual name via `useSession`
- Avatar opens menu with a logout link

PR #166, opened 2020-05-08
Search and Profile Toolbar

rayjay00:
This PR resolves issue #141 and adds the Header component, which is comprised of two new components - `HeaderSearch` and `ProfileToolbar`. 

Clicking the avatar icon toggles the menu for logout. 

image

### DIFF
--- a/src/components/Picture/Picture.tsx
+++ b/src/components/Picture/Picture.tsx
@@ -308,14 +308,15 @@ const PictureImpl = ({
       }}
       onError={(e) => {
         const url = e.currentTarget.srcset;
-        // eslint-disable-next-line no-console
-        console.error('Image failed to load\n', url);
         if (!isMounted()) {
           return;
         }
         setErrored(url);
         if (rest.onError) {
           rest.onError(e);
+        } else {
+          // eslint-disable-next-line no-console
+          console.error('Image failed to load\n', url);
         }
       }}
     />
--- a/src/scenes/Projects/ProjectList.tsx
+++ b/src/scenes/Projects/ProjectList.tsx
@@ -7,7 +7,7 @@ const useStyles = makeStyles(({ spacing }) => ({
   root: {
     flex: 1,
     overflowY: 'scroll',
-    padding: spacing(5),
+    padding: spacing(4),
   },
   options: {
     margin: spacing(3, 0),
--- a/src/scenes/Root/Header/Header.tsx
+++ b/src/scenes/Root/Header/Header.tsx
@@ -1,0 +1,26 @@
+import { makeStyles } from '@material-ui/core';
+import { FC } from 'react';
+import * as React from 'react';
+import { HeaderSearch } from './HeaderSearch';
+import { ProfileToolbar } from './ProfileToolbar';
+
+const useStyles = makeStyles(({ spacing }) => ({
+  header: {
+    padding: spacing(4),
+    width: '100%',
+    display: 'flex',
+    alignItems: 'baseline',
+    justifyContent: 'space-between',
+  },
+}));
+
+export const Header: FC = () => {
+  const classes = useStyles();
+
+  return (
+    <header className={classes.header}>
+      <HeaderSearch />
+      <ProfileToolbar />
+    </header>
+  );
+};
--- a/src/scenes/Root/Header/Header.tsx
+++ b/src/scenes/Root/Header/Header.tsx
@@ -5,9 +5,8 @@ import { HeaderSearch } from './HeaderSearch';
 import { ProfileToolbar } from './ProfileToolbar';
 
 const useStyles = makeStyles(({ spacing }) => ({
-  header: {
-    padding: spacing(4),
-    width: '100%',
+  root: {
+    padding: spacing(4, 4, 0, 4),
     display: 'flex',
     alignItems: 'baseline',
     justifyContent: 'space-between',
@@ -18,7 +17,7 @@ export const Header: FC = () => {
   const classes = useStyles();
 
   return (
-    <header className={classes.header}>
+    <header className={classes.root}>
       <HeaderSearch />
       <ProfileToolbar />
     </header>
--- a/src/scenes/Root/Header/HeaderSearch/HeaderSearch.stories.tsx
+++ b/src/scenes/Root/Header/HeaderSearch/HeaderSearch.stories.tsx
@@ -1,0 +1,6 @@
+import React from 'react';
+import { HeaderSearch as HS } from './HeaderSearch';
+
+export default { title: 'Scenes/Root/Header' };
+
+export const HeaderSearch = () => <HS />;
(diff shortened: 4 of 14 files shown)


src/scenes/Header/HeaderSearch/HeaderSearch.tsx
```diff
@@ -0,0 +1,38 @@
...
+      borderBottomColor: 'primary',
+    },
+  },
+}));
+
+export const HeaderSearch: FC = () => {
+  const classes = useStyles();
+
+  return (
+    <Input
```
CarsonF, line 27:
```suggestion
    <TextField variant="outlined"
```
or at least
```suggestion
    <OutlinedInput
```
if we aren't going to use the extra bits from TextField, like `label` & `helperText`.

src/scenes/Header/HeaderSearch/HeaderSearch.tsx
```diff
@@ -0,0 +1,38 @@
...
+    color: 'textSecondary',
+    padding: '4px',
+    borderRadius: '6px',
+    '&:before': {
+      display: 'none',
+    },
+    '&:after': {
+      borderBottomColor: 'primary',
+    },
+  },
```
CarsonF, line 20: You shouldn't need any of these other than `width` & `maxWidth` after change below.
Also `width: 100%` can be applied with `fullWidth` prop on component.
Also `maxWidth: 545` - you declare it as a number and leave off the `px`.

Also by convention the class used on the "root" element in the component is called `root`.
The classes are already scoped to the file so this would be `HeaderSearch-root-###`

src/scenes/Header/ProfileToolbar/ProfileToolbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+  },
+}));
+
+export const ProfileToolbar: FC = () => {
+  const classes = useStyles();
+  const [session] = useSession();
+  const [isProfileMenuOpen, setProfileMenuOpen] = useState(false);
+
+  return (
+    <div className={classes.menuWrap}>
```
CarsonF, line 44: I would just use a fragment here instead.
```suggestion
    <>
```

src/scenes/Header/ProfileToolbar/ProfileToolbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <div className={classes.infoWrap}>
```
CarsonF, line 45: Try using a `Card` here or if that's too much, then fallback to `Paper` (Card composes Paper).

src/scenes/Header/ProfileToolbar/ProfileToolbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <Typography className={classes.name} color="primary">
+          Hi, {session?.displayFirstName.value ?? 'Friend'}
```
CarsonF, line 47: Since this is the logged in user there's no reason not to use their real name. `session` will always be given by the time we get here so there's no reason for the fallback "friend". 
```suggestion
          Hi, {session?.realFirstName.value}
```

src/scenes/Header/ProfileToolbar/ProfileToolbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+        </Typography>
+        <AccountCircle
+          onClick={() => setProfileMenuOpen(!isProfileMenuOpen)}
+          className={clsx(classes.iconProfile, classes.icon)}
+        />
+        <NotificationsNone className={classes.icon} />
+        <MoreVert className={classes.icon} />
```
CarsonF, line 54: These should be wrapped in [IconButton](https://material-ui.com/components/buttons/#icon-buttons)s. You may not need the css customization either.

src/scenes/Home/index.tsx
```diff
@@ -1,14 +1,9 @@
...
 export const Home = () => {
   useTitle('Home');
-  const [session] = useSession();
 
-  return (
-    <div>
-      <div>Welcome, {session?.displayFirstName.value ?? 'Friend'}</div>
-    </div>
-  );
+  return <Header />;
```
CarsonF, line 8: Instead of making changes here, let's apply the `Header` in `Root`.

src/scenes/Header/ProfileToolbar/ProfileToolbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+  },
+  menuWrap: {
+    position: 'relative',
+  },
+  infoWrap: {
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'flex-end',
+    backgroundColor: 'white',
+    boxShadow: '0px 9px 12px rgba(209, 218, 223, 0.43)',
```
CarsonF, line 21: Let's use `Card`/`Paper`'s `elevation` property instead of this. It isn't going to match directly to the designs, but it will be close. Later we can change this in the theme if we want.

I think we've already customized `Card` so it defaults to the closest one.
https://github.com/SeedCompany/cord-field/blob/439d795856c0528c94fb3eed4c829f1f02856344/src/theme/overrides.ts#L5

src/scenes/Header/ProfileToolbar/ProfileToolbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+    width: '250px',
```
CarsonF, line 22: Hardcoding `width` isn't a good idea here, since name is a dynamic length. This is an example of where we, as devs, have read between the lines of the mock ups.
We could:
- Use `minWidth` instead so if the name is longer than expecting we can stretch to it.
- Leave off the extra spacing
- Add a left margin/padding to create that illusion of extra space we aren't filling up.

src/scenes/Header/ProfileMenu/ProfileMenu.tsx
```diff
@@ -0,0 +1,46 @@
...
+    fontSize: 16,
+  },
+  signOut: {
+    padding: '14px 0 6px 0',
+    borderTop: '1px solid #E5E5E5',
+    display: 'block',
+  },
+}));
+
+export const ProfileMenu: FC = () => {
```
CarsonF, line 29: Check out the `Menu` in `Sidebar`. Let's try to adapt that here. I'm also fine with this being more "basic" than the designs for now.

src/scenes/Header/ProfileToolbar/ProfileToolbar.tsx
rayjay00: I vote leave off the extra spacing 🤷

src/scenes/Header/HeaderSearch/HeaderSearch.tsx
rayjay00: Decided to go with `OutlinedInput` since I can use the input adornment with it for the icon - 
image

src/scenes/Home/index.tsx
rayjay00: Ok cool - what will go in here then?

PR #167, opened 2020-05-08
Language Engagement Card

aaronkchsu:
LanaugeEngagementCardProps would be updated with the generated Fragment interface connected to API.
Also, the rendered element should be updated with this props.


src/components/LanguageEngagementCard/LanguageEngagementCard.stories.tsx
```diff
@@ -0,0 +1,7 @@
+import { text } from '@storybook/addon-knobs';
+import React from 'react';
+import { LanguageEngagementCard as LEC } from './LanguageEngagementCard';
+
+export default { title: 'Components/LanguageEngagementCard' };
```
CarsonF, line 5:
```suggestion
export default { title: 'Components' };
```

src/components/LanguageEngagementCard/LanguageEngagementCard.stories.tsx
```diff
@@ -0,0 +1,7 @@
...
+
+export const ProjectListItemCard = () => <LEC id={text('id', '18937')} />;
```
CarsonF, line 7:
```suggestion
export const LanguageEngagementCard = () => <LEC id={text('id', '18937')} />;
```

src/components/LanguageEngagementCard/LanguageEngagementCard.tsx
```diff
@@ -0,0 +1,156 @@
...
+  editInfo: {
+    display: 'flex',
+    alignItems: 'center',
+    marginTop: 'auto',
+  },
+}));
+
+export interface LanguageEngagementCardProps {
+  id: string;
+} // This should be updated with the generated Fragment interface connected to API.
```
CarsonF, line 63: It might not be a bad idea just to define the fragment here next to this file. GraphQL is meant to be queried right next to its usage so the result is tailored to exactly what is used. Since it's a fragment it's also only defining the shape and not where the data is coming from which works great.

This would leave less to do later as more TODOs could be finished now.

src/components/Icons/ProductIcon.tsx
```diff
@@ -0,0 +1,26 @@
+import { SvgIcon, SvgIconProps } from '@material-ui/core';
+import * as React from 'react';
+import { FC } from 'react';
+
+export const ProductIcon: FC<SvgIconProps> = (props) => (
```
CarsonF, line 5: Let's call it a `ScriptIcon` instead or a name that matches exactly what the icon is. 

This is also important here because we'll have multiple types of products, and each type will have their own icon I'm guessing.

src/api/graphql.generated.tsx
```diff
@@ -44,6 +44,15 @@ export interface AddStateOutput {
   state: State;
 }
 
+export interface AdminInputDto {
```
CarsonF, line 47: Outdated schema

aaronkchsu: Sure , I will. thanks

PR #168, opened 2020-05-08
Field Overview Card

### LINKED ISSUE #159: Field Overview Card
Screen Shot 2020-05-05 at 11 40 49 AM

```tsx
interface FieldOverviewCardProps {
  title: string;
  value: string; // preformatted
  icon: React.ComponentType<SvgIconProps>;
  updatedAt: DateTime;
  to: LocationDescriptor;
  viewLabel: string;
}
```

## Requirements
- [ ] Entire top section is a `CardActionAreaLink` going to `to` prop
- [ ] Bottom left view button is a text button link also going to `to` prop
- [ ] Updated is formatted with Luxon formatting
- [ ] Icon should be rendered in a gray circle with the blue color
- [ ] Storybook story displaying card

richardgirges:
Closes #159 

Field Overview Card component and stories

### DIFF
--- a/src/components/FieldOverviewCard/FieldOverviewCard.stories.tsx
+++ b/src/components/FieldOverviewCard/FieldOverviewCard.stories.tsx
@@ -1,0 +1,30 @@
+import { makeStyles } from '@material-ui/core';
+import { text } from '@storybook/addon-knobs';
+import { DateTime } from 'luxon';
+import React from 'react';
+import { FieldOverviewCard as FieldOverviewCardComponent } from '.';
+import { CordIcon, HugeIcon } from '../Icons';
+
+export default { title: 'Components' };
+
+const useStyles = makeStyles(() => ({
+  card: {
+    width: '327px',
+  },
+}));
+
+export const FieldOverviewCard = () => {
+  const classes = useStyles();
+
+  return (
+    <FieldOverviewCardComponent
+      className={classes.card}
+      title={text('title', 'Project Budget')}
+      to="/foo"
+      value="$45,978"
+      icon={<HugeIcon icon={CordIcon} />}
+      updatedAt={DateTime.local()}
+      viewLabel={text('viewLabel', 'Budget History')}
+    />
+  );
+};
--- a/src/components/FieldOverviewCard/FieldOverviewCard.stories.tsx
+++ b/src/components/FieldOverviewCard/FieldOverviewCard.stories.tsx
@@ -1,30 +1,26 @@
-import { makeStyles } from '@material-ui/core';
-import { text } from '@storybook/addon-knobs';
+import { Box } from '@material-ui/core';
+import { AccountBalance } from '@material-ui/icons';
+import { date, text } from '@storybook/addon-knobs';
 import { DateTime } from 'luxon';
 import React from 'react';
 import { FieldOverviewCard as FieldOverviewCardComponent } from '.';
-import { CordIcon, HugeIcon } from '../Icons';
+import { HugeIcon } from '../Icons';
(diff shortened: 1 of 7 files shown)


src/components/FieldOverviewCard/FieldOverviewCard.stories.tsx
```diff
@@ -0,0 +1,18 @@
+import { text } from '@storybook/addon-knobs';
+import { DateTime } from 'luxon';
+import React from 'react';
+import { FieldOverviewCard as FieldOverviewCardComponent } from '.';
+import { CordIcon } from '../Icons';
+
+export default { title: 'Components/FieldOverviewCard' };
```
CarsonF, line 7:
```suggestion
export default { title: 'Components' };
```

Only a single story, nesting isn't necessary.

src/components/FieldOverviewCard/FieldOverviewCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+            {value}
+          </Typography>
+        </div>
+      </CardActionAreaLink>
+      <div className={classes.bottomArea}>
+        <ButtonLink color="primary" to={to}>
+          {viewLabel}
+        </ButtonLink>
+        <Typography color="textSecondary" variant="body2">
+          {updatedAt.toFormat('MM/dd/yyyy')}
```
CarsonF, line 75: We want to try to avoid `toFormat` and use `toLocaleString` instead. This uses the `Intl` code to handle differences in...internalization lol (like flipping months and days over the pond).
More info: https://moment.github.io/luxon/docs/manual/formatting.html#tolocalestring--strings-for-humans-

Also, prefix with "Updated ".


```suggestion
          Updated {updatedAt.toLocaleString(DateTime.DATE_SHORT)}
```

src/components/FieldOverviewCard/FieldOverviewCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  to,
+  viewLabel,
+}) => {
+  const classes = useStyles();
+
+  return (
+    <Card className={classes.card}>
+      <CardActionAreaLink to={to} className={classes.topArea}>
+        <div className={classes.icon}>
+          <Icon />
```
CarsonF, line 59: We should try to get the blue here. Either css or maybe `htmlColor` prop.

Also, `fontSize="large"`.

Also, maybe we should move this to separate comp. We will already need to use it in another spot. `HugeIcon` maybe? lol

I'm also wondering if we should to to change it to a ReactNode and let them pass in the created element. We would probably need to clone it, but that would allow tweaking.
For example, the CORD icon looks better offset 4px to the left because human eyes 🤪

src/components/FieldOverviewCard/FieldOverviewCard.tsx
```diff
@@ -0,0 +1,80 @@
+import { Card, makeStyles, SvgIconProps, Typography } from '@material-ui/core';
+import { DateTime } from 'luxon';
+import { FC } from 'react';
+import * as React from 'react';
+import { ButtonLink, CardActionAreaLink } from '../Routing';
+
+const useStyles = makeStyles(({ spacing, palette }) => ({
+  card: {
+    width: '321px',
+  },
```
CarsonF, line 10: Let's let that be passed in with a `className` prop.

src/components/FieldOverviewCard/FieldOverviewCard.tsx
richardgirges: Got it!

PR #170, opened 2020-05-10
Breadcrumb

jmc637:


jmc637:
Initial pr for breadcrumb. I didn't want to clone the last element so I could change the props to make it a typography element so I just overrode the styling on it. I couldn't think of an easier way to do this without cloning the element or allowing developers to pass in a Typography element as the last child of <Breadcrumbs>. Let me know if you think this fix is too hacky

src/components/Breadcrumb/Breadcrumbs.tsx
```diff
@@ -0,0 +1,23 @@
...
+    '& .MuiBreadcrumbs-li:last-child a': {
+      pointerEvents: 'none',
+      cursor: 'default',
+      textDecoration: 'none',
+      color: palette.text.primary,
+    },
+  },
+}));
+
+export const Breadcrumbs: FC = ({ children }) => {
```
richardgirges, line 17: We should add a Props interface here, and it should extend the material-ui `BreadcrumbProps` interface. e.g.:

Import the props interface:
```typescript
import { BreadcrumbsProps as MUIBreadcrumbsProps } from '@material-ui/core';
```

Define props for this component:
```typescript
export interface BreadcrumbsProps extends MUIBreadcrumbsProps {
  children: ReactNode;
}
```

Defining the component:
```typescript
export const Breadcrumbs: FC<BreadcrumbsProps> = ({ children, ...restProps }) => {
```

jmc637: I changed it to take in rest props and also have the interface

CarsonF:
> I didn't want to clone the last element and change the props to make it a typography element so I just overrode the styling on it. I couldn't think of an easier way to do this without cloning the element or allowing developers to pass in a Typography element as the last child of the Breadcrumbs component. Let me know if you think this fix is too hacky

Well we don't want to let `Typography` be passed in. We want to keep that an implementation detail here. The Usage section in the issue doesn't reference `Typography`, only `Breadcrumbs` & `Breadcrumb`.

There's a couple things we could do here.

We could make the `to` prop optional, and if it's not given then fallback to "selected" style (Typography instead of Link). Then in use the last item could just leave off the `to` prop.

The other thing is matching the `to` to the current path and if the path is the current one we style it as selected (Typography). A similar example of that logic:
https://github.com/SeedCompany/cord-field/blob/3ccc2bb02c5a10ffdf4e0d6d004fc9552e4147d7/src/components/Routing/ListItemLink.tsx#L32-L33

jmc637:
Hey I changed the Breadcrumb to only be a Typography element if the route matches. Also the Breadcrumbs component is removed because all it did was style the last Breadcrumb.

PR #181, opened 2020-05-14
Language List Item Card

jmc637:


jmc637:
Hi, this is an initial pr for language list item card. I had some questions about the data that's coming in for this card. 

- Is the name property on Language the correct name to use in this or am I supposed to use project name? 
- I saw a discussion on Slack about location on Language but I wasn't quite sure what I should do for the card. The card has location in the diagram but should that be removed?

jmc637:
I missed adding in loading state, I'll add that in today

src/components/KeyValProp/KeyValProp.tsx
```diff
@@ -0,0 +1,29 @@
...
+import * as React from 'react';
+
+export type KeyValPropProps = TypographyProps & {
+  label?: string;
+  LabelProps?: TypographyProps;
+  value?: string | number | null;
+  ValueProps?: TypographyProps;
+};
+
+export const KeyValProp: FC<KeyValPropProps> = ({
```
CarsonF, line 12: I left it in the one file un-exported so I didn't really care about the name. But then it got copied to another file, so it's good that you've extracted it out. The name is leaves something to be desired though...it's ambiguous.

I'm thinking about a more descriptive name...any thoughts?

src/components/KeyValProp/KeyValProp.tsx
```diff
@@ -0,0 +1,29 @@
+import { Typography, TypographyProps } from '@material-ui/core';
+import { FC } from 'react';
+import * as React from 'react';
+
+export type KeyValPropProps = TypographyProps & {
+  label?: string;
+  LabelProps?: TypographyProps;
+  value?: string | number | null;
```
CarsonF, line 8: We could probably just make this a `ReactNode`.

src/components/LanguageListItemCard/LanguageListItemCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+  name,
+  ethnologueName,
+  rodNumber,
+  ethnologuePopulation,
+}) => {
+  const classes = useStyles();
+
+  return (
+    <Card className={clsx(classes.root, className)}>
+      <CardActionAreaLink to="/some-path">
```
CarsonF, line 45:
```suggestion
      <CardActionAreaLink to={`/languages/${id}`}>
```

src/components/LanguageListItemCard/LanguageListItemCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <CardContent className={classes.cardContent}>
+          <div>
+            {/* TODO Is name here the name of the project or language. */}
+            {/* Do I use name or display name */}
+            <Typography variant="h4">{name?.value}</Typography>
```
CarsonF, line 50: > Is name here the name of the project or language.

The language. This card has nothing to do with projects.

> Do I use name or display name

Both. `name` is the real name of the language. `displayName` is a "fake" name used when the user doesn't have permission to see the real name. Kinda like a nickname/username/gamertag.

So use `name` and fallback to `displayName`.

src/components/LanguageListItemCard/LanguageListItemCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+              LabelProps={{ color: 'textSecondary' }}
+              label="ROD"
+              value={rodNumber?.value}
+            />
+          </div>
+          <div className={classes.rightContent}>
+            <Typography variant="body2" color="textSecondary">
+              Population
+            </Typography>
+            <Typography variant="h3">{ethnologuePopulation?.value}</Typography>
```
CarsonF, line 75: We'll want to use `organizationPopulation` first and then fallback to `ethnologuePopulation`. It's tedious and I'm thinking about making a read only, `population` helper property in the API. The situation is the `ethnologuePopulation` is defined by an external organization so we can't change that data. The `organizationPopulation` is defined by us which could be more accurate. The latter is basically an override for the former.

src/components/LanguageListItemCard/LanguageListItemCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+            {/* Do I use name or display name */}
+            <Typography variant="h4">{name?.value}</Typography>
+            <Typography
+              variant="body2"
+              color="primary"
+              className={classes.location}
+            >
+              Tenron, Asia
+              {/* TODO I'm leaving this hard coded for the moment since I'm not sure */}
+              {/* if this is supposed to be a list of languages, one language, or no language */}
```
CarsonF, line 58: This is a location. We just discovered yesterday that Languages need a list of them. So we are waiting for that to be implemented, but also the designs only show one which is tricky since there actually could be multiple.

Because of this, for now I would just leave it out completely.

src/components/KeyValProp/KeyValProp.tsx
jmc637: FieldDisplay? Maybe that sounds too much like it belongs in a form

src/components/KeyValProp/KeyValProp.tsx
jmc637: Changed to ReactNode

src/components/LanguageListItemCard/LanguageListItemCard.tsx
jmc637: Changed to correct path

src/components/LanguageListItemCard/LanguageListItemCard.tsx
jmc637: Changed to use name and falls back to displayName.

src/components/LanguageListItemCard/LanguageListItemCard.tsx
jmc637: Tries to use organization population and falls back to ethnologue population.

src/components/LanguageListItemCard/LanguageListItemCard.tsx
jmc637: Leaving it out for now

jmc637:
Changes were made for above comments and skeleton component added in for when component is still loading.

src/components/KeyValProp/KeyValProp.tsx
CarsonF: Yeah I agree. Field is associated with forms right now. We do seem to be using _display_ for this kind of thing though. `DisplayProperty`. Or maybe something more indicates this is a simple or terse UI component. What about `DisplaySimpleProperty`?

jmc637:
Hey, this is rebased and is done similar to the project list item card.

CarsonF:
Thanks @jmc637!

PR #183, opened 2020-05-14
Project List Loading State

aaronkchsu:
Loading state for projectlist and projectlistitemcard

### DIFF
--- a/src/components/ProjectListItemCard/ProjectListItemCard.stories.tsx
+++ b/src/components/ProjectListItemCard/ProjectListItemCard.stories.tsx
@@ -9,51 +9,53 @@ export default { title: 'Components' };
 export const ProjectListItemCard = () => (
   <Box display="flex">
     <PIC
-      id="123"
-      createdAt="12/12/20"
-      type={select('Type', ['Translation', 'Internship'], 'Internship')}
-      status={select(
-        'status',
-        ['InDevelopment', 'Pending', 'Active', 'Stopped', 'Finished'],
-        'Active'
-      )}
-      sensitivity={select('sensitivity', ['High', 'Low', 'Medium'], 'High')}
-      modifiedAt="12/12/20"
-      deptId={{ value: text('deptId', '1234567') }}
-      estimatedSubmission={{
-        value: DateTime.fromMillis(date('estimatedSubmission')).toLocaleString(
-          DateTime.DATE_SHORT
-        ),
-      }}
-      step={{
-        value: select(
-          'ProjectStep',
-          ['Active', 'Rejected', 'Suspended'],
+      project={{
+        id: '123',
+        createdAt: '12/12/20',
+        type: select('Type', ['Translation', 'Internship'], 'Internship'),
+        status: select(
+          'status',
+          ['InDevelopment', 'Pending', 'Active', 'Stopped', 'Finished'],
           'Active'
         ),
-      }}
-      name={{ value: text('name', 'Project A') }}
-      location={{
-        value: {
-          __typename: 'Country',
-          name: { value: 'Texas' },
-          region: {
-            value: {
-              name: {
-                value: 'United States',
-              },
-              zone: {
-                value: {
-                  name: {
-                    value: 'Americas',
+        sensitivity: select('sensitivity', ['High', 'Low', 'Medium'], 'High'),
+        modifiedAt: '12/12/20',
+        deptId: { value: text('deptId', '1234567') },
+        estimatedSubmission: {
+          value: DateTime.fromMillis(
+            date('estimatedSubmission')
+          ).toLocaleString(DateTime.DATE_SHORT),
+        },
+        step: {
+          value: select(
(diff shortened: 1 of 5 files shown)


src/components/ProjectListItemCard/ProjectListItemCard.stories.tsx
```diff
@@ -54,6 +54,7 @@ export const ProjectListItemCard = () => (
         },
       }}
       // engagements={{ total: number('Engagements', 123) }}
+      loading={boolean('Loading State', true)}
```
CarsonF, line 57:
```suggestion
      loading={boolean('Loading State', false)}
```

src/scenes/Projects/ProjectList.tsx
```diff
@@ -43,11 +40,17 @@ export const ProjectList: FC = () => {
...
-        ))}
+        {data
+          ? data?.projects.items.map((item) => (
+              <Grid item key={item.id}>
+                <ProjectListItemCard loading={loading} {...item} />
+              </Grid>
+            ))
+          : _.times(5).map((index) => (
+              <Grid item key={index}>
+                <ProjectListItemCard loading={loading} {...undefined} />
```
CarsonF, line 51: This is why I thought it would be better to have a single data prop. Because those "required" fields aren't required if it's loading, but if we aren't loading then they are required. So if we don't have the data object then we are loading, else we have data.
This is a hack around TS that could easily be avoided.

src/scenes/Projects/ProjectList.tsx
```diff
@@ -1,4 +1,5 @@
 import { Button, Grid, makeStyles, Typography } from '@material-ui/core';
+import _ from 'lodash';
```
CarsonF, line 2:
```suggestion
import { times } from 'lodash';
```

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -70,13 +72,17 @@ export const ProjectListItemCard: FC<ProjectListItemCardProps> = (props) => {
...
         <div className={classes.media}>
-          <Picture
-            source={pic}
-            fit="cover"
-            width={300}
-            height={200}
-            onError={nextPic}
-          />
+          {props.loading ? (
+            <Skeleton variant="rect" height={200} animation="wave" />
```
CarsonF, line 76: Let's leave of the `animation="wave"` from every usage. If we want use that animation we can do it once in the theme, but I think the default pulsing is fine for now.

src/scenes/Projects/ProjectList.tsx
```diff
@@ -43,11 +40,17 @@ export const ProjectList: FC = () => {
...
-            <ProjectListItemCard {...item} />
-          </Grid>
-        ))}
+        {data
+          ? data?.projects.items.map((item) => (
+              <Grid item key={item.id}>
+                <ProjectListItemCard loading={loading} {...item} />
+              </Grid>
+            ))
+          : _.times(5).map((index) => (
```
CarsonF, line 49: This is weird. Either we are loading or we are not (at least until we get into infinite scroll). The condition should be on `loading` not whether we have data.

aaronkchsu: This was my fault.

src/components/ProjectListItemCard/ProjectListItemCard.tsx
```diff
@@ -56,27 +57,40 @@ const useStyles = makeStyles(({ breakpoints, spacing }) => {
   };
 });
 
-export type ProjectListItemCardProps = ProjectListItemFragment & {
+export interface ProjectListItemCardProps {
+  project?: ProjectListItemFragment;
   className?: string;
-};
+  loading?: boolean;
```
CarsonF, line 63: See now both are optional and we have to check for twice for each property. What we really want is one or the other. So you've either given the `project` or you haven't which means we are loading.
```tsx
const loading = !project;
```

aaronkchsu: ok, got it. The reason why I did like that is because you mentioned about loading prop in ticket description.

CarsonF:
@aaronkchsu The card looked like this while loading
loading before

Which doesn't look very much like the actual card with data
with data

With my changes it looks a lot closer
with my changes

I know this is the first one we've done and there wasn't a design to go off of, so no problem. But in the future this is what I'm looking for.

aaronkchsu:
> @aaronkchsu The card looked like this while loading
> loading before
> 
> Which doesn't look very much like the actual card with data
> with data
> 
> With my changes it looks a lot closer
> with my changes
> 
> I know this is the first one we've done and there wasn't a design to go off of, so no problem. But in the future this is what I'm looking for.

The sensitivity state is different with I am done?

aaronkchsu:
The sensitivity is different with what I am done? I can see only that.

PR #194, opened 2020-05-15
Simple Language List

### LINKED ISSUE #184: Simple Language List
Depends on #178 #181 

## Requirements

- [ ] `Languages` component acting as sub-router (and parent styling - maybe none)
  - `Organizations` could be used as an example.
- [ ] `LanguageList` component connecting API to a list of `LanguageListItemCard`s.
  - The list component/route should be at `/languages`.
  - List does not need to handle pagination, sorting, or filtering.
  - [ ] Page Title
  - [ ] Stub _Sort Options_ and _Filter Options_ buttons.
          Use the `outline` variant
  - [ ] Display list total
- Can use ProjectList as an example

aaronkchsu:
Closes #184

### DIFF
--- a/src/scenes/Languages/List/LanguageList.tsx
+++ b/src/scenes/Languages/List/LanguageList.tsx
@@ -1,0 +1,66 @@
+import { Button, Grid, makeStyles, Typography } from '@material-ui/core';
+import { times } from 'lodash';
+import React, { FC } from 'react';
+import { Language } from '../../../api';
+import { LanguageListItemCard } from '../../../components/LanguageListItemCard';
+import { SortButtonDialog, useSort } from '../../../components/Sort';
+import { useLanguagesQuery } from './languages.generated';
+import { LanguageSortOptions } from './LanguageSortOptions';
+
+const useStyles = makeStyles(({ spacing }) => ({
+  root: {
+    flex: 1,
+    overflowY: 'scroll',
+    padding: spacing(4),
+  },
+  options: {
+    margin: spacing(3, 0),
+  },
+}));
+
+export const LanguageList: FC = () => {
+  const sort = useSort<Language>();
+
+  const { loading, data } = useLanguagesQuery({
+    variables: {
+      input: {
+        ...sort.value,
+      },
+    },
+  });
+  const classes = useStyles();
+
+  return (
+    <div className={classes.root}>
+      <Typography variant="h2" paragraph>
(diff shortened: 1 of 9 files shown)


src/scenes/Languages/List/LanguageList.tsx
```diff
@@ -0,0 +1,66 @@
...
+        ...sort.value,
+      },
+    },
+  });
+  const classes = useStyles();
+
+  return (
+    <div className={classes.root}>
+      <Typography variant="h2" paragraph>
+        My Languages
```
CarsonF, line 36: Technically I don't think this is "my" languages. Project list needs the "my" filter, but it isn't implemented by the API yet. Let's leave it off for now.

```suggestion
        Languages
```

src/scenes/Languages/List/LanguageList.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <SortButtonDialog {...sort}>
+            <LanguageSortOptions />
+          </SortButtonDialog>
+        </Grid>
+        <Grid item>
+          <Button variant="outlined">Filter Options</Button>
+        </Grid>
+      </Grid>
+      <Typography variant="h3" paragraph>
+        {data?.languages.total} Langauges
```
CarsonF, line 49:
```suggestion
        {data?.languages.total} Languages
```

src/scenes/Languages/List/LanguageList.tsx
```diff
@@ -0,0 +1,66 @@
...
+      </Typography>
+      <Grid container direction="column" spacing={2}>
+        {loading
+          ? times(5).map((index) => (
```
CarsonF, line 53: Since these cards are shorter let's increase the number of placeholders to 10 to fill up more of the screen.
```suggestion
          ? times(10).map((index) => (
```

src/scenes/Root/Root.tsx
```diff
@@ -30,6 +31,7 @@ export const Root = () => {
       <Route path="/" element={<Home />} />
       <Route path="/organizations/*" element={<Organizations />} />
       <Route path="/projects/*" element={<Projects />} />
+      <Route path="/Languages/*" element={<Languages />} />
```
CarsonF, line 34: URLs are always lowercase.
```suggestion
      <Route path="/languages/*" element={<Languages />} />
```

PR #195, opened 2020-05-15
"near operation" for graphql-codegen

### LINKED ISSUE #182: Implement "near operation file" for graphql code gen
## Problem 
The current config has some downsides:
- The single `graphql.generated.tsx` file is getting bigger and bigger. This makes it harder to reason about and the linter gets slower and slower for every operation added & changed.
- It's hard to find the graphql definition from TS. You say "go to" in your editor to the hook/interface and it takes you to `graphql.generated.tsx` every time and you are still left wondering where the graphql src is.

## Solution
With "near operation file" configuration, every graphql file/"operation" has it's own generated file next to it. This:
- Keeps the files small, in turn keeping linting fast.
- Helps with dev navigation as the gql src is right next to the hook/interface.

https://graphql-code-generator.com/docs/presets/near-operation-file
https://graphql-code-generator.com/docs/generated-config/near-operation-file
## Expectations

- [ ] A `src/api/schema.generated.tsx` file containing our TS schema without operations
- [ ] Every other graphql file gets a `*.generated.tsx` file next to it.
  - `components/ProjectListItem/ProjectListItem.graphql`
  - `components/ProjectListItem/ProjectListItem.generated.tsx`
- [ ] If there are any gotchas found in implementation they are communicated back to me 😄

richardgirges:
Closes #182

* `src/api/schema.generated.ts` now contains TS GQL schema with no operations
* `src/**/*.generated.tsx` are generated TS GQL operations that will live next to their counterpart `.graphql` files


codegen.yml
```diff
@@ -15,11 +15,24 @@ generates:
           content: export const possibleTypes = result.possibleTypes;
     config:
       apolloClientVersion: 3
-  ./src/api/graphql.generated.tsx:
+  ./src/api/graphql.generated.ts:
```
CarsonF, line 18:
```suggestion
  ./src/api/schema.generated.ts:
```

codegen.yml
```diff
@@ -15,11 +15,24 @@ generates:
...
+      reactApolloVersion: 3
+      scalars:
+        Date: string
+        DateTime: string
+      skipTypename: false
+      enumsAsTypes: true
+  ./src/:
+    preset: near-operation-file
+    presetConfig:
+      extension: .generated.tsx
```
CarsonF, line 31: I don't think there's an JSX in these files, right? We could change it `.ts` and limit our eslint extensions.

src/api/fragments/location.generated.tsx
```diff
@@ -0,0 +1,106 @@
+/* eslint-disable import/no-duplicates, @typescript-eslint/no-empty-interface */
```
CarsonF, line 1: This fragements folder may not make a lot of sense, esp now that less is in the api/ folder.
We don't need to change it now. But it would be good in the future to co-locate it with the `displayLocation` function and maybe move that somewhere else. At least a folder within api/ if not moving to components/.

richardgirges: Noted, makes sense!

richardgirges:
@CarsonF updates pushed! Renamed all straggling `queries.graphql` files, changed `.generated.tsx` to `.generated.ts`, and also eliminated the `DevTest` scene as there is plenty of example code now.

PR #206, opened 2020-05-26
budget overview card

aaronkchsu:
I am correctly not sure where viewLabel, UpdatedAt, and title are coming from.

### DIFF
--- a/src/components/BudgetOverviewCard/BudgetOverviewCard.tsx
+++ b/src/components/BudgetOverviewCard/BudgetOverviewCard.tsx
@@ -1,0 +1,36 @@
+import { AccountBalance } from '@material-ui/icons';
+import * as React from 'react';
+import { FC } from 'react';
+import { FieldOverviewCard } from '../FieldOverviewCard';
+import { useCurrencyFormatter } from '../Formatters/useCurrencyFormatter';
+import { BudgetOverviewFragment } from './BudgetOverview.generated';
+
+export interface BudgetOverviewCardProps {
+  budget?: BudgetOverviewFragment;
+  className?: string;
+}
+
+export const BudgetOverviewCard: FC<BudgetOverviewCardProps> = ({
+  className,
+  budget,
+}) => {
+  const formatCurrency = useCurrencyFormatter();
+
+  return (
+    <FieldOverviewCard
+      className={className}
+      title="Project Budget"
+      viewLabel="Budget Details"
+      data={
+        budget
+          ? {
+              updatedAt: budget?.createdAt,
+              value: formatCurrency(budget.total),
+              to: `budget`,
+            }
+          : undefined
+      }
+      icon={AccountBalance}
+    />
+  );
+};
--- a/src/components/Formatters/useCurrencyFormatter.tsx
+++ b/src/components/Formatters/useCurrencyFormatter.tsx
@@ -1,0 +1,10 @@
+export const useCurrencyFormatter = (options?: Intl.NumberFormatOptions) => {
+  const formatter = new Intl.NumberFormat(undefined, {
+    style: 'currency',
+    currency: 'USD',
+    maximumFractionDigits: 0,
+    minimumFractionDigits: 0,
+    ...options,
+  });
+  return (value: number) => formatter.format(value);
+};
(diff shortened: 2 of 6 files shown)


src/components/BudgetOverviewCard/BudgetOverviewCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+  budget?: BudgetItemFragment;
+  className?: string;
+}
+
+export const BudgetOverviewCard: FC<BudgetItemCardProps> = ({
+  className,
+  budget,
+}) => {
+  const currencyFormatter = useCurrencyFormatter();
+  const amount = sumBy(budget?.records, (record) => record.amount.value ?? 0);
```
CarsonF, line 19: Why not use the `total` property?

src/components/BudgetOverviewCard/BudgetOverviewCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  return (
+    <FieldOverviewCard
+      className={className}
+      title="Project Budget"
+      viewLabel="Budget history"
```
CarsonF, line 25: I know designs said this, but we won't be showing history on this page for initial release.
```suggestion
      viewLabel="Budget Details"
```

src/components/BudgetOverviewCard/BudgetOverviewCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+      data={
+        budget
+          ? {
+              updatedAt: budget?.createdAt,
+              value: currencyFormatter(amount),
+              to: `/${budget?.id}`,
```
CarsonF, line 31:
```suggestion
              to: `budget`,
```
The relative path should take us to `/project/{id}/budget`

src/components/Formatters/useCurrencyFormatter.tsx
```diff
@@ -0,0 +1,18 @@
+export enum CurrencypPosition {
+  BACK = 'de-DE',
+  FRONT = 'ja-JP',
+}
+
+export const useCurrencyFormatter = (
+  options?: Intl.NumberFormatOptions & { currencyFormat?: CurrencypPosition }
+) => {
+  const formatter = new Intl.NumberFormat(
+    options?.currencyFormat ?? CurrencypPosition.FRONT,
```
CarsonF, line 10: This argument is the locale to format in. We are leaving it `undefined` to use the current locale. In the future we might pull this from context, which is why this logic is in a hook.
```suggestion
    undefined,
```

src/components/Formatters/useCurrencyFormatter.tsx
```diff
@@ -0,0 +1,18 @@
...
+    {
+      style: options?.style ?? 'currency',
+      currency: options?.currency ?? 'USD',
```
CarsonF, line 13: Spreading `options` below this will override this value if it's given in options so there's no need to try to use it here.
```suggestion
      style: 'currency',
      currency: 'USD',
```

src/components/Formatters/useCurrencyFormatter.tsx
```diff
@@ -0,0 +1,18 @@
...
+      ...options,
```
CarsonF, line 14: Can you also add these options as defaults?:
```tsx
  maximumFractionDigits: 0, 
  minimumFractionDigits: 0, 
```
We don't want to show cents by default.

src/components/BudgetOverviewCard/BudgetItem.graphql
```diff
@@ -0,0 +1,14 @@
...
+  status
+  total
+  records {
+    fiscalYear {
+      value
+    }
+    amount {
+      value
+    }
+  }
```
CarsonF, line 13: We aren't using the `records` now so there's no need for them here.

src/components/BudgetOverviewCard/BudgetItem.graphql
```diff
@@ -0,0 +1,14 @@
+fragment BudgetItem on Budget {
```
CarsonF, line 1: Could you also rename this? This isn't a list item so I didn't want to get that confused. It really is just an overview of the budget.
```suggestion
fragment BudgetOverview on Budget {
```
And it makes sense for the fragment to be similar to the component name.

PR #216, opened 2020-05-31
Project Filter Options Dialog

### LINKED ISSUE #157: Project Filter Options Dialog
~Depends on #155 #109 #156~

Screen Shot 2020-05-05 at 11 05 44 AM
https://www.figma.com/file/IfEmcJeuNY6ucgjYctlKB7/Cord-Field?node-id=259%3A1125

## Requirements
- [x] Needs to use `DialogForm`
- [x] Cancel button should be a text button labeled "Reset Filters". Patrick agrees that is a better label than "Clear" since nothing is being removed from screen.
  - [ ] Determine if it reset form and leave dialog open or submit form with default values and close dialog.
- [x] Submit button should be labeled "Apply Filters". We aren't going to worry about dynamically fetching results to show total dynamically. It's too much work for little value.
- [x] Close button and clicking off the dialog change nothing.
- [ ] Submitting form actually applies the filters to the API, but that doesn't need to be hooked up in this ticket.

richardgirges:
Closes #157 

@CarsonF I had a question here regarding the acceptance criteria:

When the "Reset Filters" button is pressed, the `onClose` listener fires. I don't have access to the form field values within that context. How do I determine if the form was reset? I figured maybe I should be using some sort've connector or hook for retrieving the form values?

#### Question # 2:
Am I supposed to encapsulate the `onSubmit` and `onClose` methods? Or are those methods supposed to be exposed to the user of this component? Since one of the future requirements is to execute the API directly from here, I assume we don't actually want to expose `onSubmit` on the Props level, and perhaps I should be using something like `Omit<ProjectFilterOptionsDialogProps, 'onClose' | 'onSubmit'>`. But this is a heavy-handed assumption, so I wanted your input.


CarsonF:
> When the "Reset Filters" button is pressed, the `onClose` listener fires. I don't have access to the form field values within that context. How do I determine if the form was reset? I figured maybe I should be using some sort've connector or hook for retrieving the form values?

Probably the easiest way would be to add the form object as the second parameter to the `onClose` callback in `DialogForm`. This would be inline with `onSubmit`. That would allow you to wrap the `onClose` callback passed into `ProjectFilterOptionsDialog` and check for `reason === 'cancel'` and if so call `form.reset({})`.

-----

> Am I supposed to encapsulate the `onSubmit` and `onClose` methods? Or are those methods supposed to be exposed to the user of this component? Since one of the future requirements is to execute the API directly from here, I assume we don't actually want to expose `onSubmit` on the Props level, and perhaps I should be using something like `Omit<ProjectFilterOptionsDialogProps, 'onClose' | 'onSubmit'>`. But this is a heavy-handed assumption, so I wanted your input.

Try taking a look at `SortButtonDialog`, which does a couple things:
- Combines the dialog and button. This allows it to manage the dialog open state itself (_uncontrolled_, instead of _controlled_). This is nice because it's one less thing to manage at a higher level.
- It leaves the actual filter fields to be defined by children. This makes the component reusable in all lists.
- The sorting state (`value` and `onChange`) are passed in. But those are also abstracted separately with a `useSort` hook. This allows the `ProjectList` component to orchestrate the state, without having to get into the details.

Those 3 things could be applied here as well.

src/components/ProjectFilterOptionsDialog/ProjectFilterOptionsDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  console.log('props', props);
+
+  const initialValues: FormValues = props.initialValues
+    ? props.initialValues
+    : {
+        status: 'ShowAll',
+        sensitivity: 'ShowAll',
+        clusters: false,
+      };
```
CarsonF, line 43: Final Form has `initialValues`, but also `defaultValue` on each field. For this we actually want default values - the values used if there wasn't one previously. Where `initialValues` is actually the previous values when the form/dialog was opened.

src/components/ProjectFilterOptionsDialog/ProjectFilterOptionsDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+        fullWidth: true,
+        maxWidth: 'xs',
+      }}
+      title="Filter Options"
+      closeLabel="Reset Filters"
+      submitLabel="Apply Filters"
+      {...props}
+      initialValues={initialValues}
+    >
+      <RadioField name="status" label="Status">
```
CarsonF, line 57: It's not clear from designs but we actually want these to be checkboxes instead. `CheckboxesField`

src/components/ProjectFilterOptionsDialog/ProjectFilterOptionsDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <div className={classes.checkboxGrid}>
+          <Box width="50%">
```
CarsonF, line 59: `CheckboxesField` exposes the `row` prop. See if that helps for this.

src/components/ProjectFilterOptionsDialog/ProjectFilterOptionsDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <RadioOption label="Show All" value="ShowAll" />
```
CarsonF, line 60: `CheckboxOption` has a `default` that can be used instead of a `value`. Since this is a filter, the default value shows everything.

src/components/ProjectFilterOptionsDialog/ProjectFilterOptionsDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+            'Finished',
+            'Active',
+          ] as ProjectStatus[]).map((status: ProjectStatus) => (
+            <Box width="50%" key={status}>
+              <RadioOption label={displayStatus(status)} value={status} />
+            </Box>
+          ))}
+        </div>
+      </RadioField>
+      <RadioField name="sensitivity" label="Sensitivity">
```
CarsonF, line 75: Same here with checkboxes.

richardgirges:
@CarsonF I have some Q's regarding these points:

> It leaves the actual filter fields to be defined by children. This makes the component reusable in all lists.

I'm confused on this point because I thought this component was meant for Projects filter. Are you suggesting abstracting this component into a more generic "ListFilterDialog" or something like that?

> The sorting state (value and onChange) are passed in. But those are also abstracted separately with a useSort hook. This allows the ProjectList component to orchestrate the state, without having to get into the details.

Also need more clarification on this point, I might not be getting the bigger picture. Should the entire form field state be abstracted out into a hook? `useSort` appears to utilize router URL state for managing it's state. Whereas the DialogForm uses `form` state.

src/components/ProjectFilterOptionsDialog/ProjectFilterOptionsDialog.tsx
richardgirges: Checkboxes means multiple options can be selected right? Is there a way to make it so that all but one are selected in the event that `ShowAll` is selected?

CarsonF:
> Are you suggesting abstracting this component into a more generic "ListFilterDialog" or something like that?

Yes

> Should the entire form field state be abstracted out into a hook? `useSort` appears to utilize router URL state for managing it's state. Whereas the DialogForm uses `form` state.

That's what I was thinking. We could probably hide the fact that this is actually using a form.
```
value    -> initialValues
onChange <- onSubmit
```

src/components/ProjectFilterOptionsDialog/ProjectFilterOptionsDialog.tsx
richardgirges: Had to keep the `<Box width="50%">` but was able to eliminate `<div className={classes.checkboxGrid}>` thanks to `row` prop. Much cleaner 👍

richardgirges:
`FilterButtonDialog` component, `useFilter` hook, and `ProjectFilterOptions` all done. Ready for re-review @CarsonF

CarsonF:
Cool. Just on my way out. I'll check it out tonight.

src/components/Filter/useFilter.ts
```diff
@@ -0,0 +1,22 @@
...
+export const useFilter = <T>() => {
+  const navigate = useNavigate();
+  const location = useLocation();
+  const values = qs.parse(location.search, {
+    ignoreQueryPrefix: true,
+  }) as Partial<T>;
+
+  const onChange = (values: Partial<T>) => {
+    const queryVals = qs.stringify(values, {
+      addQueryPrefix: true,
```
CarsonF, line 13: How do lists stringify? It might be nicer to have
```tsx
arrayFormat: 'comma',
```

richardgirges: 10-4!

richardgirges: Just pushed up this update

PR #218, opened 2020-06-03
Feature/209 project overview scene

jmc637:


jmc637:
Hi, this is the initial pr for project overview scene. Keep getting errors from api about not being able to query certain fields. I took a screen shot of what it looks like with dummy data.
Screen Shot 2020-06-03 at 3 10 49 AM

src/components/ProjectMembersSummary/ProjectMembersSummary.tsx
```diff
@@ -20,6 +20,7 @@ export const ProjectMembersSummary: FC<ProjectMembersSummaryProps> = ({
 
   return (
     <MemberListSummary
+      total={members?.total}
```
jmc637, line 23: Changed this because using length might not be accurate could be more than the default query amount and it could get paginated to be shorter. I can put the couple files effected by this in a separate pr if it's easier to look at.

src/scenes/Projects/Overview/ProjectOverview.graphql
```diff
@@ -0,0 +1,58 @@
...
+      partnerships {
+        ...PartnershipSummary
+      }
+      engagements {
+        items {
+          ...LanguageEngagementListItem
+        }
+      }
+    }
+    #    ...on InternshipProject {
```
jmc637, line 44: Will Project overview be shown for InternshipProject's also?

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -0,0 +1,154 @@
...
+              budget={project?.budget?.value}
+            />
+            {/* TODO This should be file card when it and api are finished */}
+            <BudgetOverviewCard budget={project?.budget?.value} />
+          </div>
+          <CardGroup>
+            <ProjectMembersSummary members={project?.team} />
+            <PartnershipSummary partnerships={project?.partnerships} />
+          </CardGroup>
+          {languageEngagementSection(project as TranslationProject)}
```
jmc637, line 149: Having difficulty finding a better way to do this. Don't like changing type but not sure how else to do it since it can be either Translation or Internship.

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -0,0 +1,154 @@
...
+                aria-label="upload picture"
+                component="span"
+              >
+                <Publish />
+              </IconButton>
+            </label>
+            <Typography variant="h4">Upload Files</Typography>
+          </div>
+          <div className={classes.chips}>
+            <Chip variant="outlined" label="Mandoria, Asia" />
```
jmc637, line 133: Missed this, I'll switch this out to be non hardcoded today

src/components/ProjectMembersSummary/ProjectMembersSummary.tsx
CarsonF: Yeah that's a good catch. It's fine in this PR. I would prefer in a separate commit; commits should be changing a single logical thing.

src/scenes/Projects/Overview/ProjectOverview.graphql
```diff
@@ -0,0 +1,58 @@
...
+    location {
+      value {
+        ...DisplayPlace
+        region {
+          value {
+            name {
+              value
+            }
+          }
+        }
```
CarsonF, line 16:
```suggestion
        ...DisplayLocation
```

src/scenes/Root/Root.tsx
```diff
@@ -31,6 +32,7 @@ export const Root = () => {
       <Route path="/" element={<Home />} />
       <Route path="/organizations/*" element={<Organizations />} />
       <Route path="/projects/*" element={<Projects />} />
+      <Route path="/projects/:projectId" element={<ProjectOverview />} />
```
CarsonF, line 35: This should be declared in [`Projects`](https://github.com/SeedCompany/cord-field/blob/a11aa3df971e406f2f95e256171492981bc3aa13/src/scenes/Projects/Projects.tsx#L10) instead

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  return (
+    <div className={classes.root}>
+      {error ? (
+        <Typography variant="h4">Error fetching Project</Typography>
+      ) : (
+        <>
+          <Breadcrumbs>
+            <Breadcrumb to="/projects">Project</Breadcrumb>
+            <Breadcrumb to={pathname}>{project?.name?.value}</Breadcrumb>
```
CarsonF, line 111: It would be better to just define the path here instead of listening to route changes.
```suggestion
            <Breadcrumb to={`/projects/${id}`}>{project?.name?.value}</Breadcrumb>
```

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -0,0 +1,154 @@
...
+    color: palette.common.white,
+    '&:hover': {
+      backgroundColor: fade(palette.error.main, 0.5),
+    },
+  },
+}));
+
+export const ProjectOverview: FC = () => {
+  const classes = useStyles();
+  const { data, error } = useProjectOverviewQuery();
```
CarsonF, line 71: Missing pulling id from path and passing in here.

src/scenes/Projects/Overview/ProjectOverview.graphql
```diff
@@ -0,0 +1,58 @@
...
+    mouEnd {
+      value
+    }
+    status
+    budget {
+      value {
+        ...BudgetOverview
+      }
+    }
+    ... on TranslationProject {
```
CarsonF, line 31: You should be able to define these props without specifying which type (`... on X`) since they exist on the `Project` interface as well

CarsonF: Oh :facepalm: I screwed up that interface on API. Just pushed a fix.

src/scenes/Projects/Overview/ProjectOverview.tsx
CarsonF: Both types of projects have an engagemnet section so I wouldn't try to typecast here. When you do need to check for specifics:
```tsx
project.__typename === 'TranslationProject'
```
It would probably be easiest to just do that again with each list item
```tsx
        {project?.engagements?.items?.map((engagement) =>
          engagement.__typename === 'LanguageEngagement' ? (
            <LanguageEngagementListItemCard {...engagement} />
          ) : engagement.__typename === 'InternshipEngagement' ? (
            <InternshipEngagementListItemCard {...engagement} />
          ) : null
        )}
```

CarsonF:
Oh also missing the 3 buttons below title: location, date range, status

CarsonF:
`FilesOverviewCard` and `InternshipEngagementCard` have been merged. If you rebase on v3 you can use them.

src/components/ProjectMembersSummary/ProjectMembersSummary.tsx
jmc637: I cleaned up the commit history so this is in a separate commit

src/scenes/Projects/Overview/ProjectOverview.graphql
jmc637: This is uncommented and fixed to use updated Project interface

src/scenes/Projects/Overview/ProjectOverview.tsx
jmc637: This is changed this to not have type casting and instead uses typeName for things that are different

src/scenes/Projects/Overview/ProjectOverview.tsx
jmc637: Changed out to not be hardcoded

src/scenes/Projects/Overview/ProjectOverview.graphql
jmc637: Uses DisplayLocation instead of DisplayPlace

src/scenes/Root/Root.tsx
jmc637: Moved into Projects component

src/scenes/Projects/Overview/ProjectOverview.tsx
jmc637: Path now has id from params

src/scenes/Projects/Overview/ProjectOverview.tsx
jmc637: Woops... I'm passing this in now. Getting an error from the request about some fields not existing but it looks api related

jmc637: Screen Shot 2020-06-04 at 7 46 30 PM

src/scenes/Projects/Overview/ProjectOverview.graphql
jmc637: The fragments are fixed to use the correct interface.

jmc637:
The 3 buttons are added in and FilesOverviewCard and InternshipEngagement card are being used now.

PR #230, opened 2020-06-09
People/User List

### LINKED ISSUE #220: People/User List
Follow the suit of the other list scenes to create user list
- Refer as "users" in code and "people" in UI
- There's an existing `PersonCard` that can be used but it probably needs to be tweaked.
  - I'd prefer it to accept a data from the shape of a graphql fragment.
  - It also needs a loading state
  - Follow suit of the other `*ListItemCard` components
- List filters can be omitted
- Sorting can have fields `firstName`, `lastName`

glassblowerscat:
Closes #220 

- Rename `PersonCard` to `UserListItemCard` to match other similar card components
- Update `UserListItemCard ` to accept `user` prop of type `User` instead of multiple individual user-attribute props, to match other similar card components
- Update styling of `UserListItemCard` to match Figma design and add loading state and `stories` file
- Create `UserListItem` fragment for `UserListItemCard`
- Create `Users` scene and add to `Root` routes
- Create `UserList`, `UserSortOptions`, and `stories` file for `UserSortOptions`
- Create query for `UserList`, using `UserListItem` fragment

We were repeating the same `<div className={classes.root}>` in all our top-level routes (Projects, Languages, People, Partners), with the exact same properties, so we also create a `ListContainer` component and replace the wrapper `<div>` with it in the above-mentioned `<Scene>List` components.


glassblowerscat:
@richardgirges @CarsonF This will be ready to go when I can replace some hard-coded fields with queryable data that current throws errors when I run the query. Those fields are `avatarLetters` and `organizations -> name`. The design shows a role attribute for the user for the organization ("Field Coordinator" is what the designs show for all the users), but this attribute doesn't appear to exist in the API. Should this be removed?

src/components/ListContainer/ListContainer.tsx
```diff
@@ -0,0 +1,16 @@
+import { makeStyles } from '@material-ui/core';
+import React, { FC } from 'react';
+
+const useStyles = makeStyles(({ spacing }) => ({
+  root: {
+    flex: 1,
+    overflowY: 'scroll',
+    padding: spacing(4),
+    maxWidth: '853px',
```
CarsonF, line 9: This seems really arbitrary. What about one of the `theme.breakpoints.values.{x}`? (theme is the arg passed into this function)

src/scenes/Users/List/UserList.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <UserSortOptions />
+          </SortButtonDialog>
+        </Grid>
+      </Grid>
+      <Typography variant="h3" paragraph>
+        {data?.users.total} People
+      </Typography>
+      <Grid container spacing={3}>
+        {listOrPlaceholders(data?.users.items, 5).map((item, index) => (
+          <Grid item key={item?.id ?? index} xs={12} md={6} lg={4} xl={3}>
```
CarsonF, line 49: There seemed to be a conflict with the Grid and flexbox. Chrome was adding more spacing than desired - I think between loading skeletons and actual data. I think I moved away from this with project list (d6f896e6713d532aebd6e3fcf107f4c4d28a0ab0) and organization list (54ef429719a5cd4ae774aac41d42e6cc55199058).

Do you notice that weirdness as well?

src/scenes/Users/List/UserList.tsx
```diff
@@ -0,0 +1,56 @@
...
+      </Typography>
+      <Grid container spacing={1} className={classes.options}>
+        <Grid item>
+          <SortButtonDialog {...sort}>
+            <UserSortOptions />
+          </SortButtonDialog>
+        </Grid>
+      </Grid>
+      <Typography variant="h3" paragraph>
+        {data?.users.total} People
```
CarsonF, line 45: Hmm I know the other lists don't have this but it would be probably good to have a skeleton here while loading. Maybe 2-3ch width.

src/scenes/Users/List/UserSortOptions.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { SortOption, SortOptionProps } from '../../../components/Sort';
+
+// Alias component to define generic once
+const Option = SortOption as ComponentType<SortOptionProps<User>>;
+
+export const UserSortOptions = () => (
+  <>
+    <Option
+      default
+      value="displayFirstName"
```
CarsonF, line 13: Yeah this is tricky since there is real name and display name. We don't want to sort based on real name if they don't have access to read that name. @michaelmarshall What do you think about sorting user's based on 2 of the 4 name properties?

src/components/UserListItemCard/UserListItemCard.tsx
```diff
@@ -5,119 +5,104 @@ import {
...
         </Typography>
         <Typography color="textSecondary" variant="caption">
-          <span
-            className={clsx(classes.statusCircle, {
-              [classes.activeColor]: active,
-              [classes.notActiveColor]: !active,
-            })}
-          />
-          {active ? 'Active' : 'Inactive'}
+          {!user ? <Skeleton variant="text" width="10ch" /> : role}
```
CarsonF, line 96: Since role names are pretty wide variety of lengths, I think it would be better to have the width be a percentage instead of a certain number of characters. Typically we only use `ch` measurements when we know the actual value is going to be close that number of characters.

(same for org name and full name)

src/components/UserListItemCard/UserListItemCard.tsx
```diff
@@ -5,119 +5,104 @@ import {
...
-    borderRadius: '50%',
-    display: 'inline-block',
-  },
   cardActions: {
     display: 'flex',
     flexDirection: 'column',
     alignItems: 'center',
     padding: spacing(2),
-    border: `1px solid ${palette.grey[300]}`,
+    borderTop: `1px solid ${palette.grey[300]}`,
```
CarsonF, line 41:
```suggestion
    borderTop: `1px solid ${palette.divider}`,
```

src/components/UserListItemCard/UserListItem.graphql
```diff
@@ -0,0 +1,10 @@
+fragment UserListItem on User {
+  id
+  fullName
+  displayFirstName {
+    value
+  }
+  displayLastName {
+    value
+  }
```
CarsonF, line 9: These two properties can be removed right since they are not used?

src/components/ListContainer/ListContainer.tsx
glassblowerscat: Good call; I totally didn't think about using the theme. I just did the math from the Figma design 🙃

src/scenes/Users/List/UserList.tsx
glassblowerscat: I'm not seeing any difference in width between the grid in loading state and the grid with data... but maybe I'm not looking for the right thing? I felt like some sort of grid display was needed here because these cards are so much narrower than the Projects/Languages/Partners cards. But I notice that there's a second People Card UI depicted in the Figma design that's landscape orientation; do you want to go that direction instead?

src/scenes/Users/List/UserList.tsx
glassblowerscat: 👍

src/components/UserListItemCard/UserListItemCard.tsx
glassblowerscat: The problem here was that the enclosing element has no `width` property; its width is determined by its contents. So setting the `Skeleton` width to a percentage results in a `Skeleton` that's only a few pixels wide

src/components/UserListItemCard/UserListItemCard.tsx
glassblowerscat: 👍

src/components/UserListItemCard/UserListItem.graphql
glassblowerscat: That depends on how we decide to sort, right? Currently we're sorting by these values, but if you and Michael want to go with `realFirstName` and `realLastName` instead we'd need to switch to that

src/components/UserListItemCard/UserListItemCard.tsx
CarsonF: I think we've been flexing the cards internally. This one is probably the odd one out since it's so different.

src/scenes/Users/List/UserList.tsx
CarsonF: Yeah I would've gone with the landscape one at first too. Esp since most of the emphasis is on the picture which we don't have yet. You to you, either way I don't think this list scene is a big priority.

src/components/UserListItemCard/UserListItem.graphql
CarsonF: Sorting is entirely separate from the returned props. I don't your sorting generic uses a generated TS interface either (which it doesn't need to).

src/components/UserListItemCard/UserListItemCard.tsx
glassblowerscat: I guess that's another good argument for using the landscape-orientation card instead

src/scenes/Users/List/UserSortOptions.tsx
glassblowerscat: @CarsonF @michaelmarshall Did we ever get a ruling on this?

PR #232, opened 2020-06-09
Fix for listitem link to include extended paths

jmc637:
This is to fix cases where extended paths don't match. Ex. Link with "/projects" path doesn't appear active if user is on "/projects/SOME_ID_HERE". I think this is safe to do because I don't think there's a case where you'd want an extended path to not match.

### DIFF
--- a/src/api/schema.generated.ts
+++ b/src/api/schema.generated.ts
@@ -673,7 +673,7 @@ export type Directory = FileNode &
      * This can be used to populate a path-like UI,
      * without having to fetch each parent serially.
      */
-    parents: FileNode[];
+    parents: Directory[];
     /** The user who created this directory */
     createdBy: User;
     /** Return the file nodes of this directory */
@@ -862,7 +862,7 @@ export type File = FileNode &
      * This can be used to populate a path-like UI,
      * without having to fetch each parent serially.
      */
-    parents: FileNode[];
+    parents: Directory[];
     /** The user who uploaded the first version of this file */
     createdBy: User;
     mimeType: Scalars['String'];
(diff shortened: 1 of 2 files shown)


src/components/Routing/ListItemLink.tsx
```diff
@@ -30,7 +30,7 @@ interface ExternalProps extends BaseProps {
 export const ListItemLink = forwardRef<HTMLAnchorElement, ListItemLinkProps>(
   ({ external, to, children, ...props }, ref) => {
     const path = typeof to === 'string' ? to : { path: to.pathname! };
-    const active = useMatch(path);
+    const active = useMatch(`${path}/*`);
```
CarsonF, line 33: I believe match has an `end` option. It would be better to use that than assume things about the path string.

CarsonF:
@jmc637 Update?

src/components/Routing/ListItemLink.tsx
jmc637: This is a weird case where we actually want to match the beginning of the string instead of the end in the sidebar. Instead of passing in multiple booleans for if we want to match the beginning or end we can either add in some prop called pathToMatch that can be used to tell if the paths match or just pass in whether the component is active in from the outside.

CarsonF: Does 
```tsx
    const path = typeof to === 'string' ? to : to.pathname!;
    const active = useMatch({ path, end: false });
```
do the same thing as what you have above? Does it make the sidebar selection work correctly?

CarsonF: To be clearer, `end` is true by default. We could probably default to false here and add a `exact` prop that would set end to true.

CarsonF:
I fixed with #260. Thanks for the initiative though!

PR #233, opened 2020-06-09
Refetch after mutation (CreateProjects -> Projects List update)

richardgirges:
Here's an example of how we can update the Projects list page after executing a mutation.

Seeing as how the `Projects` list query has sort and filter variables being passed into it, it's difficult to invalidate the cache or update it manually. We don't want to start abstracting filter variables into global state just so that we can deduce how to invalidate or update cached Projects list results. 

So the easiest thing to do here is refetch existing `Projects` queries. Luckily, Apollo handles this smoothly:
* We add `refetchQueries` param to the mutation. This tells Apollo to refetch any query with the name "Projects"
  * **IMPORTANT:** Apollo will find all queries of the name "Projects" and automatically refetch them using whatever variables they were previously fetched with!
  * In the event we wanted to pass in our own variables, we're free to do that as well, e.g.:
    * `refetchQueries: [{ query: ProjectsDocument, variables: { ... } }]`
* We set `awaitRefetchQueries` to `true` so that our mutation will resolve after the refetch is complete, making for a smooth UX

### DIFF
--- a/src/scenes/Projects/Create/CreateProject.tsx
+++ b/src/scenes/Projects/Create/CreateProject.tsx
@@ -14,6 +14,7 @@ export const CreateProject = (props: Except<Props, 'onSubmit'>) => {
   const submit: Props['onSubmit'] = async (input) => {
     const res = await createProject({
       variables: { input },
+      refetchQueries: ['ProjectList'],
     });
     const project = res.data!.createProject.project;
 
--- a/src/scenes/Projects/List/ProjectList.tsx
+++ b/src/scenes/Projects/List/ProjectList.tsx
@@ -4,7 +4,7 @@ import { Project } from '../../../api';
 import { ProjectListItemCard } from '../../../components/ProjectListItemCard';
 import { SortButtonDialog, useSort } from '../../../components/Sort';
 import { listOrPlaceholders } from '../../../util';
-import { useProjectsQuery } from './projects.generated';
+import { useProjectListQuery } from './projects.generated';
 import { ProjectSortOptions } from './ProjectSortOptions';
 
 const useStyles = makeStyles(({ spacing }) => ({
@@ -24,7 +24,7 @@ const useStyles = makeStyles(({ spacing }) => ({
 export const ProjectList: FC = () => {
   const sort = useSort<Project>();
 
-  const { data } = useProjectsQuery({
+  const { data } = useProjectListQuery({
     variables: {
       input: {
         ...sort.value,
--- a/src/scenes/Projects/List/projects.generated.ts
+++ b/src/scenes/Projects/List/projects.generated.ts
@@ -9,11 +9,11 @@ import {
 } from '../../../components/ProjectListItemCard/ProjectListItem.generated';
 import { ProjectListItemFragmentDoc } from '../../../components/ProjectListItemCard/ProjectListItem.generated';
 
-export interface ProjectsQueryVariables {
+export interface ProjectListQueryVariables {
   input: Types.ProjectListInput;
 }
 
-export type ProjectsQuery = { __typename?: 'Query' } & {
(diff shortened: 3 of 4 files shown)


richardgirges:
One other thing to note: the use of a string for naming the `Projects` query to be refetched is important here. This is what tells Apollo to blindly refetch all cached variables of the same name. 

We cannot pass `ProjectsDocument` into `refetchQueries` unless we were prepared to supply the query's variables as well.

CarsonF:
I don't love the magic string here for the query name. But I do love how easy this is.

Since these query names are more relevant now we should be more careful with them.

Maybe we should rename `Projects` to `ProjectList`.

I think there there's with codegen to select these query names and give them a type. Maybe that could help with magic strings?

Is waiting for list refresh really what we want? I think it makes sense if you are on the project list scene, but any other scene on the app is it worth it? Couldn't we just be quick here and then show the list loading if the user navigates there?

glassblowerscat:
> I don't love the magic string here for the query name. But I do love how easy this is.

You're right that this can get unwieldy if we find ourselves in a place where a mutation needs to refetch several different queries—keeping track of all those different query names can get tedious—but it seems like this application might not have complex UI-state dependencies like that. So we might never run into this problem?

> Is waiting for list refresh really what we want? I think it makes sense if you are on the project list scene, but any other scene on the app is it worth it? Couldn't we just be quick here and then show the list loading if the user navigates there?

`refetchQueries` actually won't refetch queries from other scenes because they'll be inactive at the time. So we will still need to decide how we want to handle scenes with UI that depend on recently updated db values. Richard proposed using `fetchPolicy: 'network-only'`, which tells the Apollo client to always request data through the network interface. That would definitely work and be relatively straightforward to handle with, as you say, loading UI.

glassblowerscat:
I'm personally a fan of `fetchPolicy: 'cache-and-network'`, but I admit that I'm not sure how that would look in practice for this app. It's possible that the user might visit a page and briefly see the previous UI state retrieved from the cache before the network request returns with the updated data. If so, that might not be desirable for our user experience.

richardgirges:
@CarsonF Renamed query `Projects` to `ProjectList`, and eliminated `awaitRefetchQueries`. The `loading` prop for the query gets set to true during refetch so we end up in a loading state anyway.

I don't see anything in the codegen, typescript and typescript-operations plugin that allows us to export consts with query names. Also poked around the `generated` files and don't see anything that gives us what we want here.

This will probably warrant additional investigation if we really don't want the magic string, or possibly writing our own codegen plugin. Either way, I strongly believe this is too much added value to abandon simply because of a magic string; the alternative approaches to cache invalidation will not be easy nor intuitive.

> refetchQueries actually won't refetch queries from other scenes because they'll be inactive at the time. So we will still need to decide how we want to handle scenes with UI that depend on recently updated db values. Richard proposed using fetchPolicy: 'network-only', which tells the Apollo client to always request data through the network interface. That would definitely work and be relatively straightforward to handle with, as you say, loading UI.

@glassblowerscat actually Apollo _will_ refetch queries from other scenes if they've been previously cached. So we may be able to get away with not doing `network-only` for now.

richardgirges:
Created this ticket to address the magic string issue: https://github.com/SeedCompany/cord-field/issues/234 - lemme know what you guys think

glassblowerscat:
Hm, I must have misread the docs about `refetchQueries`. I just know trying to update UI on subsequent pages after a mutation has always been a tremendous pain.

richardgirges:
@glassblowerscat most def. That's why this `refetchQueries` API is so great. It didn't use to allow you to refetch cached queries that had variables passed in but they seem to have resolved that in the past year or so. Super powerful!

richardgirges:
Word!

PR #235, opened 2020-06-10
Feature/207 project member card

jmc637:


src/api/displayUserName.ts
```diff
@@ -0,0 +1,6 @@
+import { NameFragment } from './fragments/name.generated';
+
+export const displayUserName = (user: NameFragment) =>
```
CarsonF, line 3: We have `User.fullName` now, which is basically the same thing but server side.

src/api/displayRole.ts
```diff
@@ -0,0 +1,7 @@
+import { startCase } from 'lodash';
+import { SecuredRoles } from './schema.generated';
+
+export const displayRole = (securedRoles: Partial<SecuredRoles>) =>
```
CarsonF, line 4: Split `displayRole` out of this and rename this one to `displayRoles`

src/components/ProjectMemberCard/ProjectMemberCard.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { randomPersonPic } from '../MemberListSummary/MemberListSummary.stories';
+import { ProjectMemberCard as ProjectMemberCardComponent } from './ProjectMemberCard';
+
+export default { title: 'Components' };
+
+const roles: Role[] = ['Consultant', 'Development'];
+
+const projectMember = {
+  id: '0',
+  createdAt: CalendarDate.fromJSDate(new Date()),
```
CarsonF, line 14:
```suggestion
  createdAt: DateTime.local(),
```

src/components/ProjectMemberCard/ProjectMemberCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+    height: spacing(7),
+    marginRight: spacing(2),
+  },
+  memberInfo: {
+    flexGrow: 1,
+  },
+}));
+
+export interface ProjectMemberCardProps {
+  imageSource?: string;
```
CarsonF, line 39: Same for this as coming from user. But we won't have it for initial release. Need to fallback to avatarLetters

src/components/ProjectMemberCard/ProjectMemberCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const classes = useStyles();
+  const dateFormatter = useDateFormatter();
+
+  const name = projectMember?.user.value
+    ? displayUserName(projectMember.user.value)
+    : '';
+  const rolesString = projectMember?.roles
+    ? displayRole(projectMember.roles)
+    : '';
+  const createdAtString = projectMember?.createdAt
```
CarsonF, line 59: `createdAt` is a `DateTime` - use the `useDateTimeFormatter`.

Also if you want to update those to accept and return nullable that would probably be useful too.

src/components/ProjectMemberCard/ProjectMemberCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+            )}
+          </Typography>
+        </div>
+      </CardContent>
+      <CardActions className={classes.cardActions}>
+        <ButtonLink
+          disabled={!projectMember}
+          size="large"
+          color="primary"
+          to={`/${projectMember?.id}/edit`}
```
CarsonF, line 98: This button will just bring up a dialog. So this can be replaced with a normal `Button`. Also `onEdit` callback passed in and connected to `onClick` here

src/components/ProjectMemberCard/ProjectMemberCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+            ) : (
+              rolesString
+            )}
+          </Typography>
+        </div>
+      </CardContent>
+      <CardActions className={classes.cardActions}>
+        <ButtonLink
+          disabled={!projectMember}
+          size="large"
```
CarsonF, line 96: I don't think it should be large. Editting isn't the normal user path forward here.

PR #236, opened 2020-06-11
Update Register User for new API

### LINKED ISSUE #224: Update Register User for new API
Following https://github.com/SeedCompany/cord-api-v3/issues/444

Our register user code needs to be updated for API changes. I think for this it's just changing the gql mutation. Input is the same.

glassblowerscat:
Closes #224 

**NOTE:** Requires `cord-api-v3` branch `create-person`, which has conflicts with `master`. This should not be merged until a PR is ready to be merged for `create-person`. It also seems like the API may still be a little in flux in this branch, so we should generate typings again for this PR after the API has been solidified.

- Update `Register` mutation to match API changes and re-generate typings
- Update `Register` component with new values to match new API/typings


glassblowerscat:
@CarsonF It seems like this mutation is working so far as to create the new person with all the correct information, but the return from the mutation has all the fields as `null`. Is this intentional? The schema says that the mutation should be returning a `CreatePersonOutput`

glassblowerscat:
Blocked until `create-person` API branch has a merged PR

CarsonF:
https://github.com/SeedCompany/cord-api-v3/pull/561

PR #237, opened 2020-06-11
Issue 222 create language

frankyma:
**Technical Notes:**
Create Language functionality with adding the language option under create showing the `CreateLanguage` modal.  Implementation matching create Org/Project.  A few TODOs:

1. The `TextField` component coerces number inputs to strings used by the gql mutation, even if the MUI input type is number.  I added in code to coerce to number on the `CreateLanguage` component but that's a short term solution.  Not sure if needs to create a new number input component or expand TextField to enforce numbers

2. Making the TextField `type="number"` adds a spinner on the UI.  This is native to html. We'll have to remove with the theme overrides (possibly) or with the custom input mentioned above
Screen Shot 2020-06-11 at 10 15 47 AM

**Testing:**
Open the Languages Scene. Click create new language, fill in the info in the form.  You should see a success snackbar and the new language populate on the scene.

### DIFF
--- a/src/scenes/Languages/Create/CreateLangaugeForm.tsx
+++ b/src/scenes/Languages/Create/CreateLangaugeForm.tsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import { CreateLanguageInput } from '../../../api';
+import {
+  DialogForm,
+  DialogFormProps,
+} from '../../../components/Dialog/DialogForm';
+import { SubmitError, TextField } from '../../../components/form';
+
+export type CreateLanguageFormProps = DialogFormProps<CreateLanguageInput>;
+
+export const CreateLanguageForm = (props: CreateLanguageFormProps) => (
+  <DialogForm
+    DialogProps={{
+      fullWidth: true,
+      maxWidth: 'xs',
+    }}
+    {...props}
+    title="Create Language"
+  >
+    <SubmitError />
+    <TextField
+      name="language.name"
+      label="Name"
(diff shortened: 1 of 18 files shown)


src/scenes/Languages/Create/CreateLangaugeForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+      label="Name"
+      placeholder="Enter language"
+      autoFocus
+      required
+    />
+    <TextField
+      name="language.displayName"
+      label="Display Name"
+      placeholder="Enter Display Name"
+      autoFocus
```
CarsonF, line 32: We don't want to autofocus every field - only one. Typically the first one.

frankyma: Changed

frankyma:
> Yeah we'll want to create a `NumberField` component.
> 
> I think we can do `type="numeric"` which does the same thing without the up down buttons, but I think it kinda makes sense here. I would be fine leaving it.

I don't see `type="numeric"` as an accepted type: https://www.w3schools.com/html/html_form_input_types.asp.  I'll keep it at type **number** if you say it's fine.

CarsonF:
Ah it's actually `inputMode`. I've got a little bit of time right now. Let me take a stab at a number field

richardgirges:
Gonna merge this, created a ticket to track work for `NumberField` component: https://github.com/SeedCompany/cord-field/issues/242

PR #243, opened 2020-06-15
FE Issue 227 partnerships list

frankyma:
Create Project -> Partnerships list view, including partnerships card component and storybook view.  Couldn't find a design for this scene/component so laid out a simple text card that shows a few Partnership fields.  Let me know if it's missing data.  The issue ticket also mentions a breadcrumb and add button but I couldn't find that on the other list scenes.

`EditPartnership` will be implemented separately

**Testing**
Create some partnerships for a project (need to do this on the gql playground as the FE isn't implemented).  Then go to the projects/{id}/partnerships and see the partnerships.  The sub header will say 0 Partners because that's a BE bug.
Screen Shot 2020-06-15 at 10 44 59 AM


CarsonF:
Just doing a first pass here based on the picture (super helpful 👍 )

- This scene is similar to Team Members List, so we can use that as an example
  - We want breadcrumbs at the top of the scene
  - I don't think we need to display the total here since this is a _short_ list
    - "short" being less than 10 items. So we skip pagination, sorting, and filtering for these
  - We'll want to have the "Add Partnership" button as well.
    - That could probably just be before the list.
    - It should also be conditionally hidden based on `project.partnerships.canCreate`.
    - Hmm since it's conditionally shown we need to figure out what the loading state is. Maybe hidden by default? I would like to try to prevent "jumps" between loading state and actual data. Open to suggestions.
For the ListItem:
- Remove the "Organization Name: " prefix. The actual name is enough
- Move _types_ to right below that. I don't think this needs the label either as the values are unique enough to be self-explanatory. 
- Move _createdAt_ to bottom right, like the other list item cards
- I think the designs are also removing the ":pencil2: &nbsp; Edit Info" text so we could go ahead and do that here too.

Ok I'll circle back to review code later today

frankyma:
> Just doing a first pass here based on the picture (super helpful 👍 )
> 
> * This scene is similar to Team Members List, so we can use that as an example
>   
>   * We want breadcrumbs at the top of the scene
>   * I don't think we need to display the total here since this is a _short_ list
>     
>     * "short" being less than 10 items. So we skip pagination, sorting, and filtering for these
>   * We'll want to have the "Add Partnership" button as well.
>     
>     * That could probably just be before the list.
>     * It should also be conditionally hidden based on `project.partnerships.canCreate`.
>     * Hmm since it's conditionally shown we need to figure out what the loading state is. Maybe hidden by default? I would like to try to prevent "jumps" between loading state and actual data. Open to suggestions.
>       For the ListItem:
> * Remove the "Organization Name: " prefix. The actual name is enough
> * Move _types_ to right below that. I don't think this needs the label either as the values are unique enough to be self-explanatory.
> * Move _createdAt_ to bottom right, like the other list item cards
> * I think the designs are also removing the "✏️   Edit Info" text so we could go ahead and do that here too.
> 
> Ok I'll circle back to review code later today

Made the requested changes.  The CardActions section spacing doesn't look perfect, mostly because the edit button sizing.  I didn't want to spend too much time making it pixel perfect because we don't have designs for what the final card looks like.

PR #244, opened 2020-06-16
FE issue 227 edit partnerships

frankyma:
**DO NOT MERGE**. This is branched off https://github.com/SeedCompany/cord-field/tree/feature-FE-227-Partnerships_List

Edit partnership feature on the project partnerships. 

**Tech Notes**
`DialogForm` to show the edit modal.  With radios for statuses and checkboxes for types.  I'm not sure what the statuses and types mean so let me know if I'm missing anything.

The `DialogForm`'s `initialValues` prop doesn't pass down to the radio field, only the types.  We'll need to add that to see the initial values selected.  Imo it's better UX.  `RadioField` has a `initialValue` prop but it's not working

Screen Shot 2020-06-17 at 8 47 51 AM

**Testing**
Create Project partnerships in the gql playground if you haven't already.  Click the edit icon on a Partnership Card, see the modal open.  Change the statuses/types and click submit.  You should see the partnership reflect the new status.  Delete deletes the partnership and closes the modal.

### DIFF
--- a/src/api/displayEnums.ts
+++ b/src/api/displayEnums.ts
@@ -19,3 +19,9 @@ export const displayEngagementStatus = displayEnum<EngagementStatus>();
 export const displayRole = displayEnum<Role>();
 export const displayRoles = (roles: Role[]) =>
   roles.map(displayRole).join(', ');
+
+export const PartnershipStatuses: PartnershipAgreementStatus[] = [
+  'NotAttached',
+  'AwaitingSignature',
+  'Signed',
+];
--- a/src/scenes/Partnerships/Edit/EditPartnership.tsx
+++ b/src/scenes/Partnerships/Edit/EditPartnership.tsx
@@ -1,4 +1,4 @@
-import { Box, Button } from '@material-ui/core';
+import { Box } from '@material-ui/core';
 import React, { FC } from 'react';
 import { Except } from 'type-fest';
 import {
@@ -17,6 +17,8 @@ import {
   CheckboxOption,
   RadioField,
   RadioOption,
+  SubmitAction,
+  SubmitButton,
   SubmitError,
 } from '../../../components/form';
 import {
@@ -39,16 +41,6 @@ export const EditPartnership: FC<EditPartnershipProps> = ({
   const [updatePartnership] = useUpdatePartnershipMutation();
   const [deletePartnership] = useDeletePartnershipMutation();
 
-  const onDeleteClick = () => {
-    const { id } = partnership;
-    deletePartnership({
-      variables: { input: id },
-      refetchQueries: [GQLOperations.Query.ProjectPartnerships],
-    });
-    //TODO: correctly handle closing the dialog.  props.onClose has a TS errors
-    // props.onClose && props.onClose('cancel');
-  };
-
   const radioOptions = PartnershipStatuses.map((status) => (
     <RadioOption
       key={status}
@@ -58,20 +50,37 @@ export const EditPartnership: FC<EditPartnershipProps> = ({
   ));
 
   return (
-    <DialogForm<UpdatePartnershipInput>
+    <DialogForm<UpdatePartnershipInput & SubmitAction<'delete'>>
       DialogProps={{
         fullWidth: true,
         maxWidth: 'xs',
       }}
       {...props}
-      onSubmit={(input) => {
-        updatePartnership({
+      onSubmit={async (input) => {
+        const refetchQueries = [GQLOperations.Query.ProjectPartnerships];
+        if (input.submitAction === 'delete') {
+          await deletePartnership({
+            variables: { input: input.partnership.id },
+            refetchQueries,
+          });
+          return;
+        }
+        await updatePartnership({
           variables: { input },
-          refetchQueries: [GQLOperations.Query.ProjectPartnerships],
+          refetchQueries,
         });
       }}
       title={`Edit Partnership with ${partnership.organization.name.value}`}
-      leftAction={<Button onClick={onDeleteClick}>Delete</Button>}
+      leftAction={
+        <SubmitButton
+          action="delete"
+          color="error"
+          fullWidth={false}
+          variant="text"
+        >
+          Delete
+        </SubmitButton>
+      }
       initialValues={{
         partnership: {
           id: partnership.id,
--- a/src/scenes/Partnerships/Edit/EditPartnership.tsx
+++ b/src/scenes/Partnerships/Edit/EditPartnership.tsx
@@ -1,0 +1,119 @@
+import { Box, Button } from '@material-ui/core';
+import React, { FC } from 'react';
+import { Except } from 'type-fest';
+import {
+  displayPartnershipStatus,
+  GQLOperations,
+  PartnershipStatuses,
+  PartnershipType,
+  UpdatePartnershipInput,
+} from '../../../api';
+import {
+  DialogForm,
+  DialogFormProps,
+} from '../../../components/Dialog/DialogForm';
+import {
+  CheckboxesField,
+  CheckboxOption,
+  RadioField,
+  RadioOption,
+  SubmitError,
+} from '../../../components/form';
+import {
+  EditPartnershipFragment,
+  useDeletePartnershipMutation,
+  useUpdatePartnershipMutation,
+} from './EditPartnership.generated';
+
+type EditPartnershipProps = Except<
+  DialogFormProps<UpdatePartnershipInput>,
+  'onSubmit' | 'initialValues'
+> & {
+  partnership: EditPartnershipFragment;
+};
+
+export const EditPartnership: FC<EditPartnershipProps> = ({
+  partnership,
+  ...props
+}) => {
+  const [updatePartnership] = useUpdatePartnershipMutation();
+  const [deletePartnership] = useDeletePartnershipMutation();
+
+  const onDeleteClick = () => {
+    const { id } = partnership;
+    deletePartnership({
+      variables: { input: id },
+      refetchQueries: [GQLOperations.Query.ProjectPartnerships],
+    });
+    //TODO: correctly handle closing the dialog.  props.onClose has a TS errors
+    // props.onClose && props.onClose('cancel');
+  };
+
+  const radioOptions = PartnershipStatuses.map((status) => (
+    <RadioOption
+      key={status}
+      label={displayPartnershipStatus(status)}
+      value={status}
+    />
+  ));
+
+  return (
+    <DialogForm<UpdatePartnershipInput>
+      DialogProps={{
+        fullWidth: true,
+        maxWidth: 'xs',
+      }}
+      {...props}
+      onSubmit={(input) => {
+        updatePartnership({
+          variables: { input },
+          refetchQueries: [GQLOperations.Query.ProjectPartnerships],
+        });
+      }}
+      title={`Edit Partnership with ${partnership.organization.name.value}`}
+      leftAction={<Button onClick={onDeleteClick}>Delete</Button>}
+      initialValues={{
+        partnership: {
+          id: partnership.id,
+          agreementStatus: partnership.agreementStatus.value,
+          mouStatus: partnership.mouStatus.value,
+          types: partnership.types.value,
+        },
+      }}
+    >
+      <SubmitError />
+      <CheckboxesField
+        name="partnership.types"
+        label="Types"
+        row
+        disabled={!partnership.types.canEdit}
+      >
+        {([
+          'Managing',
+          'Funding',
+          'Impact',
+          'Technical',
+          'Resource',
+        ] as PartnershipType[]).map((type: PartnershipType) => (
+          <Box width="50%" key={type}>
+            <CheckboxOption key={type} label={type} value={type} />
+          </Box>
+        ))}
+      </CheckboxesField>
+      <RadioField
+        name="partnership.agreementStatus"
+        label="Agreement Status"
+        disabled={!partnership.agreementStatus.canEdit}
+      >
+        {radioOptions}
+      </RadioField>
+      <RadioField
+        name="partnership.mouStatus"
+        label="Mou Status"
+        disabled={!partnership.mouStatus.canEdit}
+      >
+        {radioOptions}
+      </RadioField>
+    </DialogForm>
+  );
+};
--- a/src/api/operations.generated.ts
+++ b/src/api/operations.generated.ts
@@ -9,6 +9,8 @@ export const GQLOperations = {
     ProjectOverview: 'ProjectOverview',
   },
   Mutation: {
+    UpdatePartnership: 'UpdatePartnership',
+    DeletePartnership: 'DeletePartnership',
     ForgotPassword: 'ForgotPassword',
     Login: 'Login',
     Logout: 'Logout',
@@ -26,6 +28,7 @@ export const GQLOperations = {
     DisplayZone: 'DisplayZone',
     DisplayPlace: 'DisplayPlace',
     BudgetOverview: 'BudgetOverview',
+    EditPartnership: 'EditPartnership',
     InternshipEngagementListItem: 'InternshipEngagementListItem',
     LanguageEngagementListItem: 'LanguageEngagementListItem',
     LanguageListItem: 'LanguageListItem',
--- a/src/components/PartnershipCard/PartnershipCard.stories.tsx
+++ b/src/components/PartnershipCard/PartnershipCard.stories.tsx
@@ -1,6 +1,7 @@
 import { action } from '@storybook/addon-actions';
 import { select, text } from '@storybook/addon-knobs';
 import React from 'react';
+import { PartnershipStatuses } from '../../api';
 import { csv } from '../../util';
 import { dateTime } from '../knobs.stories';
 import { PartnershipCard } from './PartnershipCard';
@@ -19,18 +20,10 @@ export const WithData = () => {
       value: csv(text('Types', 'Managing, Funding')),
     },
     mouStatus: {
-      value: select(
-        'Mou Status',
-        ['NotAttached', 'AwaitingSignature', 'Signed'],
-        'NotAttached'
-      ),
+      value: select('Mou Status', PartnershipStatuses, 'NotAttached'),
     },
     agreementStatus: {
-      value: select(
-        'Mou Status',
-        ['NotAttached', 'AwaitingSignature', 'Signed'],
-        'NotAttached'
-      ),
+      value: select('Mou Status', PartnershipStatuses, 'NotAttached'),
     },
   };
 
--- a/src/scenes/Partnerships/Edit/EditPartnership.generated.ts
+++ b/src/scenes/Partnerships/Edit/EditPartnership.generated.ts
@@ -1,0 +1,173 @@
+/* eslint-disable import/no-duplicates, @typescript-eslint/no-empty-interface */
+import * as ApolloReactCommon from '@apollo/client';
+import * as ApolloReactHooks from '@apollo/client';
+import gql from 'graphql-tag';
+import * as Types from '../../../api/schema.generated';
+import { PartnershipCardFragment } from '../../../components/PartnershipCard/PartnershipCard.generated';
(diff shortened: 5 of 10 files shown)


src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -10,11 +24,89 @@ interface EditPartnershipProps {
...
+    await updatePartnership({
+      variables: { input: inputWithId },
+      refetchQueries: ['ProjectPartnerships'],
+    });
+    enqueueSnackbar(
+      `Updated Partnership with ${partnership.organization.name.value}`,
+      {
+        variant: 'success',
+      }
+    );
```
CarsonF, line 46: I'm not sure snackbars are that useful here. We have them on global create dialogs since those can be invoked anywhere and you don't immediately see the results on screen. With both edit and delete here the changes should be reflected in the list and immediately visible to the user.
What do you think?

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -10,11 +24,89 @@ interface EditPartnershipProps {
...
 export const EditPartnership: FC<EditPartnershipProps> = ({
   state,
-  // partnership,
+  partnership,
 }) => {
+  const [updatePartnership] = useUpdatePartnershipMutation();
+  const [deletePartnership] = useDeletePartnershipMutation();
+  const { enqueueSnackbar } = useSnackbar();
+
+  const submit = async (input: UpdatePartnershipInput) => {
```
CarsonF, line 33: We should be careful conflating form state with other interfaces. For instance, `input.id` is valid to TS but it doesn't exist. I would be more comfortable with a separate form state interface that matches how the fields are defined and used here.

I wish there was a way to infer this automatically but given the nature of how the fields are defined it's impossible.

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -10,11 +24,89 @@ interface EditPartnershipProps {
...
+    state.onClose();
+    enqueueSnackbar(
+      `Deleted Partnership with ${partnership.organization.name.value}`,
+      {
+        variant: 'success',
+      }
+    );
+  };
+
+  // TODO: have DialogForm pass down initialValues to the radio options (currently passed to Checkboxes)
```
CarsonF, line 64: I'll play around here and see if I can reproduce

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -10,11 +24,89 @@ interface EditPartnershipProps {
...
+      DialogProps={{
+        fullWidth: true,
+        maxWidth: 'xs',
+      }}
+      title="Edit Partnership"
+      leftAction={<Button onClick={onDeleteClick}>Delete</Button>}
+      // initialValues={initialValues}
+    >
+      <SubmitError />
+      <RadioField name="partnership.agreementStatus" label="Agreement Status">
```
CarsonF, line 87: These fields should also be disabled based on their `canEdit` property

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -10,11 +24,89 @@ interface EditPartnershipProps {
...
+        <RadioOption label="Not Attached" value="NotAttached" />
+        <RadioOption label="Awaiting Signature" value="AwaitingSignature" />
+        <RadioOption label="Signed" value="Signed" />
+      </RadioField>
+      <RadioField name="partnership.mouStatus" label="Mou Status">
+        <RadioOption label="Not Attached" value="NotAttached" />
+        <RadioOption label="Awaiting Signature" value="AwaitingSignature" />
+        <RadioOption label="Signed" value="Signed" />
+      </RadioField>
+      <CheckboxesField name="partnership.types" label="Partnership Type" row>
```
CarsonF, line 97:
```suggestion
      <CheckboxesField name="partnership.types" label="Partnership Types" row>
```
or maybe even just "Types".

Can you also move this to the top? It's the most important field.

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <DialogForm
+      onSubmit={submit}
+      {...state}
+      DialogProps={{
+        fullWidth: true,
+        maxWidth: 'xs',
+      }}
+      title="Edit Partnership"
```
CarsonF, line 82: Oh it might be good to include the org name in this title

.eslintrc.yml
```diff
@@ -23,3 +23,8 @@ rules:
         - '@material-ui/core/*'
         - '@material-ui/icons/*'
         - lodash/*
+  no-console:
+    - warn
+    - allow:
+        - warn
+        - error
```
CarsonF, line 30: Let's decide and do this separately from this PR

src/components/EditPartnership/EditPartnership.tsx
frankyma: agree, removed

src/components/EditPartnership/EditPartnership.tsx
frankyma: Changed to use `UpdatePartnership` without ID

src/components/EditPartnership/EditPartnership.tsx
frankyma: Moved types up and disabled if `canEdit: false` although it doesn't work as expected on the radio fields.  The radio fields are still able to be toggled

.eslintrc.yml
frankyma: removed

src/components/EditPartnership/EditPartnership.tsx
frankyma: added

src/components/EditPartnership/EditPartnership.tsx
frankyma: Passed through the initialValues, you'll see the checkboxes working but not the dialog

frankyma:
@CarsonF ready for re-review

src/components/EditPartnership/EditPartnership.tsx
frankyma: and disabled if `canEdit: false` although it doesn't work as expected on the radio fields.  The radio fields are still able to be toggled

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,115 @@
...
+  SubmitError,
+} from '../form';
+import { PartnershipCardFragment } from '../PartnershipCard/PartnershipCard.generated';
+import {
+  useDeletePartnershipMutation,
+  useUpdatePartnershipMutation,
+} from './EditPartnership.generated';
+
+interface EditPartnershipProps {
+  state: DialogState;
```
CarsonF, line 21: This should just forward on the `DialogForm` props. The `useDialog` props are a subset of this, but this component shouldn't care about that (how it's used).

See what I did with `EditOrganization`:
https://github.com/SeedCompany/cord-field/blob/a42a18dbe55ddf3f5933162662b0ca3e863f5587/src/scenes/Organizations/Edit/EditOrganization.tsx#L12-L17

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,115 @@
...
+    });
+  };
+
+  const onDeleteClick = async () => {
+    const { id } = partnership;
+    await deletePartnership({
+      variables: { input: id },
+      refetchQueries: ['ProjectPartnerships'],
+    });
+    state.onClose();
```
CarsonF, line 52: Hmm this is a new use case. We need to show this delete in progress state in the UI. This happens automatically with the form submit handler, but not with this handler. Let me think about the best way to handle this.

src/components/PartnershipCard/PartnershipCard.graphql
```diff
@@ -3,14 +3,18 @@ fragment PartnershipCard on Partnership {
   createdAt
   types {
     value
+    canEdit
```
CarsonF, line 6: This fragment should _only_ be used for the `PartnershipCard` and what it needs to display. It should be able to change this at will without affecting anything else.

The details/edit scenes should use a different fragment for the needs of that scene.

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,115 @@
...
+    partnership: {
+      agreementStatus: partnership.agreementStatus.value,
+      mouStatus: partnership.mouStatus.value,
+      types: partnership.types.value,
+    },
+  };
+
+  return (
+    <DialogForm
+      onSubmit={submit}
```
CarsonF, line 65: Take a look at `EditOrganization` and how I passed the generic to the component and put onSubmit and initialValues inline so I didn't have to explicitly type those variables. 
https://github.com/SeedCompany/cord-field/blob/a42a18dbe55ddf3f5933162662b0ca3e863f5587/src/scenes/Organizations/Edit/EditOrganization.tsx#L23
https://github.com/SeedCompany/cord-field/blob/a42a18dbe55ddf3f5933162662b0ca3e863f5587/src/scenes/Organizations/Edit/EditOrganization.tsx#L31-L41

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,115 @@
...
+export const EditPartnership: FC<EditPartnershipProps> = ({
+  state,
+  partnership,
+}) => {
+  const [updatePartnership] = useUpdatePartnershipMutation();
+  const [deletePartnership] = useDeletePartnershipMutation();
+
+  const submit: DialogFormProps<DialogInput>['onSubmit'] = async (input) => {
+    const inputWithId = {
+      partnership: { ...input.partnership, id: partnership.id },
```
CarsonF, line 38: So since we talked about this I found out (maybe via @jmc637) that we can pass ID into the `initialValues` and get it back out on submit - basically like a hidden field. This seems to be the best way to go about it as it makes types easier but still doesn't add any overhead for defining a field component for the ID.

I also do this in `EditOrganization`.

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,115 @@
...
+    };
+    await updatePartnership({
+      variables: { input: inputWithId },
+      refetchQueries: ['ProjectPartnerships'],
```
CarsonF, line 42: This should use the `GQLOperations` object instead of raw strings

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,115 @@
...
+    });
+  };
+
+  const onDeleteClick = async () => {
+    const { id } = partnership;
+    await deletePartnership({
+      variables: { input: id },
+      refetchQueries: ['ProjectPartnerships'],
```
CarsonF, line 50: This should use the `GQLOperations` object instead of raw strings

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const inputWithId = {
+      partnership: { ...input.partnership, id: partnership.id },
+    };
+    await updatePartnership({
+      variables: { input: inputWithId },
+      refetchQueries: ['ProjectPartnerships'],
+    });
+  };
+
+  const onDeleteClick = async () => {
```
CarsonF, line 46: Hmm this is a new use case. We need to show this delete in progress state in the UI. This happens automatically with the form submit handler, but not with this handler. Let me think about the best way to handle this.

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,115 @@
...
+      </CheckboxesField>
+      {/* TODO: fix missing initial valus and disabled not working on RadioField */}
+      <RadioField
+        name="partnership.agreementStatus"
+        label="Agreement Status"
+        disabled={!partnership.agreementStatus.canEdit}
+      >
+        <RadioOption label="Not Attached" value="NotAttached" />
+        <RadioOption label="Awaiting Signature" value="AwaitingSignature" />
+        <RadioOption label="Signed" value="Signed" />
```
CarsonF, line 102: These labels should be defined via the `displayPartnershipStatus` function. Could probably loop through like you are doing with partnership type. Also you could move it to a variable so you don't have to do it again below.

src/components/Dialog/useDialog.tsx
```diff
@@ -1,13 +1,19 @@
 import { useCallback, useMemo, useState } from 'react';
 
+export interface DialogState {
+  open: boolean;
+  onClose: () => void;
+  onExited: () => void;
+}
```
CarsonF, line 7: This file should be reverted. See comment on `EditPartnershipProps`

src/components/EditPartnership/EditPartnership.tsx
```diff
@@ -0,0 +1,115 @@
...
+          'Impact',
+          'Technical',
+          'Resource',
+        ] as PartnershipType[]).map((type: PartnershipType) => (
+          <Box width="50%" key={type}>
+            <CheckboxOption key={type} label={type} value={type} />
+          </Box>
+        ))}
+      </CheckboxesField>
+      {/* TODO: fix missing initial valus and disabled not working on RadioField */}
```
CarsonF, line 94: Ok I'll fix this.

src/components/EditPartnership/EditPartnership.tsx
frankyma: Fixed

src/components/EditPartnership/EditPartnership.tsx
frankyma: Ok, I've commented out the onClose since with the prop fix above it has a TS error.  I added a TODO for you to come back to this.

CarsonF: You need to pass the close reason right?

src/components/PartnershipCard/PartnershipCard.graphql
frankyma: Understood, removed

src/components/EditPartnership/EditPartnership.tsx
frankyma: Changed

src/components/Dialog/useDialog.tsx
frankyma: reverted

src/components/EditPartnership/EditPartnership.tsx
frankyma: got it

src/components/EditPartnership/EditPartnership.tsx
frankyma: Used `displayPartnershipStatus` and moved out to variable

src/components/EditPartnership/EditPartnership.tsx
frankyma: Fixed

src/components/EditPartnership/EditPartnership.tsx
frankyma: Fixed

src/components/EditPartnership/EditPartnership.tsx
frankyma: Close reason and a `Form` object as second argument
Screen Shot 2020-06-22 at 12 01 34 PM

frankyma:
@CarsonF ready

CarsonF:
@frankyma Can you move `EditPartnership`? It should be in `scenes/Partnerships`. It's not a re-useable component.

frankyma:
> @frankyma Can you move `EditPartnership`? It should be in `scenes/Partnerships`. It's not a re-useable component.

@CarsonF yes sorry missed that, pushed.

src/api/displayEnums.ts
```diff
@@ -19,3 +19,9 @@ export const displayEngagementStatus = displayEnum<EngagementStatus>();
 export const displayRole = displayEnum<Role>();
 export const displayRoles = (roles: Role[]) =>
   roles.map(displayRole).join(', ');
+
+export const PartnershipStatuses: PartnershipAgreementStatus[] = [
+  'AwaitingSignature',
+  'NotAttached',
```
CarsonF, line 25: It's a workflow, and not attached is the starting point.
```suggestion
  'NotAttached',
  'AwaitingSignature',
```

src/scenes/Partnerships/Edit/EditPartnership.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const radioOptions = PartnershipStatuses.map((status) => (
+    <RadioOption label={displayPartnershipStatus(status)} value={status} />
+  ));
+
+  return (
+    <DialogForm<UpdatePartnershipInput>
+      //TODO: fix the form flashing back to initialValues
+      onSubmit={(input) => {
+        updatePartnership({
```
CarsonF, line 60: You need to await the `updatePartnership` call. Does that fix the initial values flashing back?

src/scenes/Partnerships/Edit/EditPartnership.tsx
```diff
@@ -0,0 +1,117 @@
...
+          'Impact',
+          'Technical',
+          'Resource',
+        ] as PartnershipType[]).map((type: PartnershipType) => (
+          <Box width="50%" key={type}>
+            <CheckboxOption key={type} label={type} value={type} />
+          </Box>
+        ))}
+      </CheckboxesField>
+      {/* TODO: fix missing initial valus and disabled not working on RadioField */}
```
CarsonF, line 100: This is fixed in `v3`

src/scenes/Partnerships/Edit/EditPartnership.tsx
```diff
@@ -0,0 +1,117 @@
...
+  return (
+    <DialogForm<UpdatePartnershipInput>
+      //TODO: fix the form flashing back to initialValues
+      onSubmit={(input) => {
+        updatePartnership({
+          variables: { input },
+          refetchQueries: [GQLOperations.Query.ProjectPartnerships],
+        });
+      }}
+      {...props}
```
CarsonF, line 65: Ok just FYI here. The order in which props are spread and what comes before and after determines the default values and overrides. The logic that this order defines should match the TS interface. 
This component's props exclude `onSubmit` and `initialValues` from our `DialogFormProps`, because they are handled here. So those should come after the spreading of `props`. Even if you pass them in they should do nothing because we enforce them here.
The opposite is true of passing in `DialogProps`. Those are just defaults, so we don't exclude them from our props interface and they should come before the props spread so if/when they are passed in they are correctly applied.

Does that make sense?

src/api/displayEnums.ts
frankyma: fixed

src/scenes/Partnerships/Edit/EditPartnership.tsx
frankyma: confirmed fixed

src/scenes/Partnerships/Edit/EditPartnership.tsx
frankyma: Yes it does, thanks for the explanation.  Changed.

src/scenes/Partnerships/Edit/EditPartnership.tsx
frankyma: No that doesn't fix. here's the code I tried
```
 onSubmit={async (input) => {
        await updatePartnership({
          variables: { input },
          refetchQueries: [GQLOperations.Query.ProjectPartnerships],
        });
      }}
```

src/components/EditPartnership/EditPartnership.tsx
CarsonF: Fixed via custom submit action

src/scenes/Partnerships/Edit/EditPartnership.tsx
CarsonF: Hmm ok. Let's solve it in a separate pass. This will probably be happening everywhere.

CarsonF:
Thanks @frankyma!

Issue #225, opened 2020-06-05
Currency Field


CarsonF:
A new form field that is for currency.

Previously implemented on corporate website [NumberFormatInput](https://github.com/SeedCompany/website-v3/blob/824eb75b69896ce53cf50a6fd8164eabb8694771/src/components/Shared/Common/NumberFormatInput.tsx), [CurrencyInput](https://github.com/SeedCompany/website-v3/blob/47a7b11ee4e2dc1c813d9ef50eb772eefc75adf2/src/components/Donate/DonateCard/CurrencyInput.tsx), [Amount](https://github.com/SeedCompany/website-v3/blob/d6851b50729a4b9533601742fe7a7f74056f91ea/src/components/Donate/DonateCard/Amount.tsx#L105).
I know it's private :/ It's using `react-number-format` library. Feel free to use that or another implementation. I've seen some hook versions as well. I think the MUI date picker library uses one.

It should be just be able to be extended from our `TextField`, just like Email or Password Fields.

CarsonF:
@richardgirges Maybe you can pick this up since you have access to those links?

richardgirges:
Sounds good @CarsonF !

CarsonF:
Closed with #246

PR #246, opened 2020-06-17
Formatted Text & Number Fields

### LINKED ISSUE #242: NumberField component
Per discussion here: https://github.com/SeedCompany/cord-field/pull/237#pullrequestreview-429194085

Create a `NumberField` component

CarsonF:
Closes #242 #225 

- Created `FormattedTextField` as base for formatted input
- Created `NumberField` based on this that handles integers, floats, negatives, prefixes/suffixes.
  Turns out this is really hard lol. It's not perfect and there are still some interactions I don't like. I still felt like this was a better base than `react-number-format` library. Happy to be proven wrong and replace it.
- Created `YearField` thin wrapper around `NumberField` for years.
- Updated `EmailField` to lowercase input as emails are case-insensitive. Added option to keep sensitivity like when creating account. The API should handle this insensitivity anyways, this is just for looks.
- Added stories for everything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "react-router-dom": "^6.0.0-alpha.5",
     "react-scripts": "^3.4.1",
     "react-use": "^14.1.1",
+    "rifm": "^0.12.0",
     "ts-essentials": "^6.0.4",
     "validator": "^13.0.0"
   },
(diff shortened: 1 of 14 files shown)


richardgirges:
Solid!

PR #250, opened 2020-06-18
GraphQL named operations codegen plugin

### LINKED ISSUE #234: Codegen GraphQL query const strings
We want graphql-codegen to export a const string representing every single GraphQL Query Operation name.

* Investigate any potential plugins that may provide this functionality out of the box
* Investigate potentially rolling our own custom plugin for producing this behavior

Writing your own Codegen plugin: https://graphql-code-generator.com/docs/custom-codegen/write-your-plugin

## Additional Details:

As an example, the following query:

```graphql
query FooList {
  foo {
    id
    name
  }
}
```

...should allow you to import a `const` by the name of `FooListOperation` or something of that nature:

```typescript
import { FooListQuery } from './foo.generated';

console.log(FooListQuery); // "FooList"
```

This will allow us to have type-safe operation names to pass into our mutations for cache-invalidation upon refetching:

```typescript
    const res = await createFoo({
      variables: { input },
      refetchQueries: [FooListQuery],
    });
```

richardgirges:
Closes #234


codegen.yml
```diff
@@ -23,12 +23,14 @@ generates:
   ./src/api/schema.generated.ts:
     plugins:
       - typescript
+      - named-operations-object
```
CarsonF, line 26: Could we put this in operations.generated.ts? It should change based on client side usage instead of API changes.

richardgirges: Good idea, I'll update right now

PR #251, opened 2020-06-18
Give FF field type of checkbox for SwitchField

blakekrammes:
- MUI uses a checkbox for its Switch component so we need to tell FF to use `type: 'checkbox'` for the input so it can map the checked state correctly
  - I think `value` was being converted to a string, but `checked` will remain a boolean

### DIFF
--- a/src/components/form/SwitchField.tsx
+++ b/src/components/form/SwitchField.tsx
@@ -32,7 +32,11 @@ export const SwitchField: FC<SwitchFieldProps> = ({
   ...props
 }) => {
   const name = useFieldName(nameProp);
-  const { input, meta, rest } = useField(name, { defaultValue, ...props });
+  const { input, meta, rest } = useField(name, {
+    defaultValue,
+    ...props,
+    type: 'checkbox',
+  });
   const disabled = disabledProp ?? meta.submitting;
   const ref = useFocusOnEnabled<HTMLInputElement>(meta, disabled);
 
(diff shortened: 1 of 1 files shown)


CarsonF:
@richardgirges and I found an hour ago that the string value was given to FF from the url due to url logic. So IMO that's where the real problem is. I'm working on that right now.

But this could be another safety layer so it's not a bad idea.

src/components/form/SwitchField.tsx
```diff
@@ -32,7 +32,11 @@ export const SwitchField: FC<SwitchFieldProps> = ({
   ...props
 }) => {
   const name = useFieldName(nameProp);
-  const { input, meta, rest } = useField(name, { defaultValue, ...props });
+  const { input, meta, rest } = useField(name, {
+    defaultValue,
+    type: 'checkbox',
+    ...props,
```
CarsonF, line 38: We should probably enforce this type right?
```suggestion
    ...props,
    type: 'checkbox',
```

blakekrammes:
Yeah. I was going to mention that the form was being messed up by values parsed from the url.

src/components/form/SwitchField.tsx
blakekrammes: Yar.

blakekrammes:
For some reason, `form.reset()` doesn't restore the initial values, which in this case is an empty object I think.

CarsonF:
Ah I'll fix that. It is restoring the initial values. The initial values are the values that the dialog is opened with, not the actual "default values".

PR #252, opened 2020-06-18
Create story for DisplaySimpleProperty component

blakekrammes:


src/components/DisplaySimpleProperty/DisplaySimpleProperty.stories.tsx
```diff
@@ -0,0 +1,50 @@
+import { boolean, select, text } from '@storybook/addon-knobs';
+import * as React from 'react';
+import { DisplaySimpleProperty as DSP } from './DisplaySimpleProperty';
+
+export default {
+  title: 'Components',
+};
+
+export const Property = () => (
```
CarsonF, line 9: `Components/Property` will be the path. We want it to match the component name

PR #256, opened 2020-06-18
Cord field 226 budget scene

### LINKED ISSUE #226: Budget Scene
- [x] url: `/projects/{id}/budget`
- [ ] data pulled from `project.budget`
- [x] I'm guessing an MUI table would be best here
  - https://material-table.com/ might be a good resource here to allow for inline editing
  - Only the amount column is possibly editable. Each `BudgetRecord` amount should check `record.amount.canEdit` to allow editing.
  - Each record (row) will need to be saved individually
  - [x] This could be a separate `RecordsTable` component in the scene folder
- The budget has a `total` property but it might be better to calculate based on record amounts since they are editable here. Another example of a "side effect" from the API. Or open to using total working with Apollo cache. If it would help if the API returned the budget from that mutation then we can make that happen as well.
- We have a `useCurrencyFormatter` hook too, which will probably be useful.

Here's screenshot from v2 as an idea:
Screen Shot 2020-06-05 at 4 56 42 PM
It should have column headers, fiscal year displayed better, amount formatted and right aligned.

glassblowerscat:
Closes #226 

- Create `ProjectBudget` component in `scenes/Projects/Budget` and wire up to project budget data
- Create `BudgetRecords` fragment, `ProjectBudget` query, and `UpdateProjectBudgetRecords` mutation
- Update `material-table` to `1.69.0` from `1.60.0`. This new version has a feature for editing individual cells, instead of entire rows, which is a nicer user experience. 
- Update patch for `material-table` to work with new `Icon` types and patch two bugs:
  1. In the `MTableEditField` component, the `aria-label` attribute was being passed incorrectly.
  2. Several components were improperly passing an unsupported `scrollWidth` property as an attribute. A fix for this has already been committed in the repo but just not been released yet
(Issues were opened in the library's repo to address to other minor bugs found during development.)
- Update our `Table` component to accept the standard version of `material-table`'s `isEditable` and `onRowUpdate` props instead of our own custom version. We really weren't gaining much from this, and it just makes it more brittle

## Plumbing changes to `material-table`
- In an earlier commit for this PR we had not updated to 1.69.0 because it has a dependency on a version of `jspdf` with a bug that throws a warning in the dev environment. It turned out the version of `jspdf` in 1.68.0 had an even worse bug due to a dependency on `file-saver` that it was trying to retrieve from a  specific commit on GitHub that threw an error during installation, breaking in our CI. So we are updating to 1.69.0 and using the workaround of  overriding the version of `jspdf` in `resolutions`. This should be temporary; `material-table` has already patched this themselves but just hasn't  released the patch.
- We had previously patched `material-table` to change the typings for the `icons` prop and make them more friendly so we didn't have to use the ugly and tedious `forwardRef` required by the official typings. Since we are now patching the library **again**, it seems like this isn't worth the effort. This isn't a bug; it's how `material-table` intends to be used. We should probably just live with it instead of putting in the effort to patch it every time just to make our own code that we almost never touch look nicer


src/scenes/Projects/Budget/ProjectBudget.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const columns = [
+    {
+      title: 'ID',
+      field: 'id',
+      hidden: true,
+    },
+    {
+      title: 'Organization',
+      field: 'organization',
+      editable: 'never' as const,
```
CarsonF, line 82: FYI you could do `as const` on the root columns list definition instead of each time here. It applies to all inner declarations.

src/scenes/Projects/Budget/ProjectBudget.tsx
```diff
@@ -0,0 +1,176 @@
...
+      ) : (
+        <>
+          {loading ? (
+            <Skeleton variant="text" width="20%" />
+          ) : (
+            <Breadcrumbs>
+              <Breadcrumb to="/projects">Projects</Breadcrumb>
+              <Breadcrumb to={`/projects/${projectId}`}>
+                {projectName}
+              </Breadcrumb>
```
CarsonF, line 125:
```suggestion
              <ProjectBreadcrumb data={data?.project}  />
              <Breadcrumb to=".">Budget</Breadcrumb>
```

src/scenes/Projects/Budget/ProjectBudget.tsx
```diff
@@ -0,0 +1,176 @@
...
+                        const input = {
+                          budgetRecord: {
+                            id: data.id,
+                            amount: Number(newAmount),
+                          },
+                        };
+                        await updateBudgetRecord({
+                          variables: { input },
+                          refetchQueries: [GQLOperations.Query.ProjectBudget],
+                        });
```
CarsonF, line 165: I believe this won't update the total until the query is refetched "some time later". Could we either:
- calculate the total ourselves based on the records which should update immediately in cache
- or write to the apollo cache directly the new total. I guess that still requires us to calculate it first.

src/scenes/Projects/Budget/ProjectBudget.tsx
```diff
@@ -0,0 +1,176 @@
...
+                <Skeleton variant="text" width="20%" />
+                <Skeleton variant="text" width="10%" />
+              </>
+            ) : (
+              <>
+                <Typography variant="h2">
+                  {data?.project.name.value} Budget
+                </Typography>
+                {budget && (
+                  <Typography variant="body1" className={classes.total}>
```
CarsonF, line 140: Could we use a standard variant here instead of a manual `em` value? Maybe h3/h4?

CarsonF (commented):
Haven't pulled it down yet but this is what I'm seeing

src/scenes/Projects/Budget/ProjectBudget.tsx
glassblowerscat: Refetches for live queries run immediately upon successful completion of the mutation; this is the whole reason for their existence. The only reason it would take a noticeably long time to update the data would be because of network, server, or db latency. If we're worried about that, then the next option would generally be updating the cache using the returned data. This would **not** require us to calculate it, provided we request the new amount in the returned mutation data. But it is sort of a pain, which is why we have been sticking with refetches so far.

CarsonF: I'm worried about API performance yes.
If we didn't want to calculate it we could add a `total` property to the `UpdateBudgetRecordOutput` output type. However, I don't think Apollo Cache would connect that to `Budget.total` automatically so we'd have to write it manually.

I think calculating it ourselves it honestly probably the easiest. The `total` from the API was really there as a helper for list or summary views that weren't getting all the records.
It's only a one-liner...
https://github.com/SeedCompany/cord-api-v3/blob/a32c02431f71febad76d8f89debea17f770488a8/src/components/budget/budget.resolver.ts#L54

Thoughts?

glassblowerscat: Yeah, fair enough. It does seem simpler to just calculate the total rather than writing a second value back into the cache alongside the updated amount

CarsonF (approved):
@glassblowerscat I made....some...changes lol. Mostly just tweaking styling and abstracting styling to Table / theme.

I changed the header so it looks like the rest of the sub-pages. I know it's a little weird that budget is duplicated in breadcrumbs and header, but I'd rather be consistent for now.

Here's what it looks like now:
Screen Shot 2020-09-08 at 1 48 46 PM


If you wouldn't mind looking over my commits, I'll merge.

CarsonF:
I'm still not happy with the edit styling. It's jumpy and input looks bad (small font, no formatting, etc). I'd rather get this in though and revisit.

CarsonF:
I forgot until now, but I wanted to look at the sizing. The max width is pretty wide for this content, and I'd like to shrink it.

glassblowerscat:
Looks good to me!

PR #257, opened 2020-06-19
New Query Param Handling

CarsonF:
This adds a hook (creator) that creates a type-safe getter/setter for query parameters via a defined schema. 

It also updates sorting & filtering to use this new hook.

## `makeQueryHandler`
```tsx
const useColorState = makeQueryHandler({
  name: StringParam,
  red: NumberParam,
  green: NumberParam,
  blue: NumberParam,
  favorite: BooleanParam(),
});

export const MyComponent = () => {
  const [color, setColor] = useColorState();
  color.name // string | undefined
  color.red // number | undefined

  ...
};
```

This is a hook created outside of a component, which means the config/schema cannot be changed at runtime. This is more performant and saves us a lot of effort with checking if the schema has changed.

Multiple hooks can be used at once. The setter only changes the values
_defined_ by the hook's schema. The setter itself does not allow partial sets.
Basically it works just like `useState`.

By default, the setter replaces the current history entry instead of pushing
a new one. This default is the opposite of react-router's defaults, but I
think it makes more sense here for query parameter handling.

Each value of the schema (i.e. `StringParam`) is an object that defines `encode` & `decode` functions.
This makes it easy to have custom handling and abstract it.

### Under the hood
This integrates `react-router` v6 with the [`serialize-query-params`](https://github.com/pbeshai/serialize-query-params) library.
This is essentially what [`use-query-params`](https://github.com/pbeshai/use-query-params) library does, but it isn't updated
for RR v6. RR v6 includes more OOTB in this area so only thin wrapper is needed.

### Naming?

Maybe `makeQueryParams` would be better? Would stay inline with the `makeStyles/useStyles` convention.

## Filtering

Previously we had a single generic `useFilter` hook. The goal of it was to abstract the complexity of query param handling into a getter/setter.
But it had problems because it didn't know the types of each parameter so had some trouble decoding booleans, etc. It also didn't know which parameters it should keep track of so the setter blindly replaced all query params. This overwrote the other params managed separately like sorting.

This new solution I believe still meets the goal of abstracting the complexity away. As seen below it's short and pretty much only concerned with "business logic". In the diff it's actually the same lines of code because it replaces a previously defined interface.

```tsx
export const useProjectFilters = makeQueryHandler({
  status: EnumListParam<ProjectStatus>(),
  sensitivity: EnumListParam<Sensitivity>(),
  clusters: BooleanParam(),
});
```

## Sorting

The down side of defining the hook as a variable is that generics can't be used since variables can't be generic. This comes into play with the `useSort` hook which is generic. It may not be that big of deal as wrapping in another function like I've done isn't terrible.
```tsx
export const useSort = <T>() => {
  const [value, onChange] = useSortHandler();
  // Adapt output to have generic for sort value
  return {
    value: value as Partial<SortValue<T>>,
    onChange: onChange as (next: Partial<SortValue<T>>) => void,
  };
};
```


richardgirges:
Mind blown! Super clean. Great job! I don't know why but I like `makeQueryHandler` better for the name.

PR #263, opened 2020-06-23
FE 228 User Detail and Edit User

frankyma:
The User Detail and Edit User Dialog scenes.
Screen Shot 2020-06-23 at 11 19 17 AM
Screen Shot 2020-06-23 at 11 19 08 AM

**Technical Notes**

- Used the same fragment for UserDetail and EditUser (like implemented in Organizations).  I thought that was ok because these two scenes are paired.  Let me know if that understanding is incorrect.

- chose not to put in the initialValues for the dialog

- email is a not supported to edit based on the EditUserInput schema.  Let me know if that's a mistake

**Testing**
Go to /users/{userId} and see your user details.  Click the pencil icon to edit, and on submit see the changes.

### DIFF
--- a/src/api/operations.generated.ts
+++ b/src/api/operations.generated.ts
@@ -48,5 +48,7 @@ export const GQLOperations = {
     orgDetails: 'orgDetails',
     EditPartnership: 'EditPartnership',
     userDetails: 'userDetails',
+    UserForm: 'UserForm',
+    ss: 'ss',
   },
 };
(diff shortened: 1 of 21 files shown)


frankyma:
@CarsonF you can start reviewing this if you'd like, the only blocker is the phone number input on the edit dialog

frankyma:
@CarsonF changed the `numberField` to `textField` and rebased v3 so all functionality should work now.  Removed the **blocked** tag

PR #264, opened 2020-06-23
FE 262 Error and Not Found components

frankyma:
Created Error component to view Error and Not Found.  

**Technical notes**

- Combined the two views into one component because they're so similar but can be broken into 2 if needed.

- Need to pass in navigate back as a prop for the back button.  Was going to put in the navigate back through the history object but `useHistory` is not available for router v6: https://github.com/ReactTraining/react-router/issues/7189.

**Testing:**
See the views in storybook.
Screen Shot 2020-06-23 at 1 04 49 PM
Screen Shot 2020-06-23 at 1 04 43 PM

### DIFF
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -1,3 +1,36 @@
+import { Button, makeStyles, Typography } from '@material-ui/core';
 import React from 'react';
 
-export const Error = () => <div>This is the Error component</div>;
+const useStyles = makeStyles(({ spacing }) => ({
+  container: {
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'flex-start',
+    '& h3': {
+      fontWeight: 'bold',
+    },
+  },
+  button: {
+    height: spacing(4),
+    width: spacing(16),
+    marginTop: spacing(2),
+  },
+}));
+
+export const Error = () => {
+  const classes = useStyles();
+
+  return (
+    <div className={classes.container}>
+      <Typography>Oops, Sorry.</Typography>
+      <Typography variant="h3">ERROR</Typography>
+      <Button
+        variant="contained"
+        color="secondary"
+        classes={{ root: classes.button }}
+      >
+        Back
+      </Button>
+    </div>
+  );
+};
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -1,8 +1,8 @@
 import { Button, makeStyles, Typography } from '@material-ui/core';
 import React, { FC } from 'react';
+import { useNavigate } from 'react-router';
 
 interface ErrorProps {
-  navigateBack: (event: React.MouseEvent<HTMLButtonElement>) => void;
   variant?: 'Error' | '404';
 }
 
@@ -26,8 +26,9 @@ const useStyles = makeStyles(({ spacing }) => ({
   },
 }));
 
-export const Error: FC<ErrorProps> = ({ navigateBack, variant = 'Error' }) => {
+export const Error: FC<ErrorProps> = ({ variant = 'Error' }) => {
   const classes = useStyles();
+  const navigate = useNavigate();
 
   const errorMessage =
     variant === 'Error' ? (
@@ -47,7 +48,7 @@ export const Error: FC<ErrorProps> = ({ navigateBack, variant = 'Error' }) => {
       <Typography>Oops, Sorry.</Typography>
       {errorMessage}
       <Button
-        onClick={navigateBack}
+        onClick={() => navigate(-1)}
         variant="contained"
         color="secondary"
         classes={{ root: classes.button }}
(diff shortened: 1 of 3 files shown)


src/components/Error/Error.tsx
```diff
@@ -0,0 +1,59 @@
...
+    marginRight: spacing(1),
+  },
+  button: {
+    height: spacing(4),
+    width: spacing(16),
+    marginTop: spacing(2),
+  },
+}));
+
+export const Error: FC<ErrorProps> = ({ navigateBack, variant = 'Error' }) => {
```
CarsonF, line 29:
```suggestion
export const Error: FC<ErrorProps> = ({ variant = 'Error' }) => {
const navigate = useNavigate();
// navigate(-1);
```

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,59 @@
...
+  errorType: {
+    display: 'flex',
+  },
+  boldText: {
+    fontWeight: 'bold',
+    marginRight: spacing(1),
+  },
+  button: {
+    height: spacing(4),
+    width: spacing(16),
```
CarsonF, line 24: Let's not change the width and height manually. Let's stick to `Button.size` prop options

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,59 @@
+import { Button, makeStyles, Typography } from '@material-ui/core';
+import React, { FC } from 'react';
+
+interface ErrorProps {
+  navigateBack: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  variant?: 'Error' | '404';
```
CarsonF, line 6: I think it would be good to pass in the error. 
```suggestion
  error?: ApolloError;
```
We could determine the error type based off this and possibly show more error information (maybe just in dev).

Similar `ApolloError` handling is done here https://github.com/SeedCompany/cord-field/blob/2191e6167629aa0c49b19272268a475ea66eceb4/src/api/ApolloProvider.tsx#L59-L65

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const errorMessage =
+    variant === 'Error' ? (
+      <Typography variant="h3" className={classes.boldText}>
+        ERROR
+      </Typography>
+    ) : (
+      <div className={classes.errorType}>
+        <Typography variant="h3" className={classes.boldText}>
+          404
+        </Typography>
```
CarsonF, line 41: I know this is from the designs but it rubs me the wrong way. There's no reason to artificially create this error code. The not found message is more than enough.

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,59 @@
...
+  container: {
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'flex-start',
+  },
+  errorType: {
+    display: 'flex',
+  },
+  boldText: {
+    fontWeight: 'bold',
```
CarsonF, line 19: If we must bold, let's do it this way:
```suggestion
    fontWeight: typography.weight.bold,
```

Otherwise let's just try to stick to the typography variants already defined.

CarsonF (dismissed):
Sorry for throwing in more requirements here. Sometimes I have to see it to make the connection.

src/components/Error/Error.tsx
frankyma: changed

src/components/Error/Error.tsx
frankyma: Sure, changed back to default medium size.  It's a little different than the wide button in the design but I think it's fine
Screen Shot 2020-06-23 at 4 31 46 PM
Screen Shot 2020-06-23 at 4 31 38 PM

src/components/Error/Error.tsx
frankyma: agree, removed

src/components/Error/Error.tsx
frankyma: changed to `typography.weight.bold`

src/components/Error/Error.tsx
frankyma: passed through the error instance but will leave out the specific error handling for now

frankyma:
@CarsonF ready for re-review

src/components/Error/Error.tsx
CarsonF: Yeah that's fine. Maybe size large would be good since it's the only button on screen?

frankyma: here's large, doesn't look good imo
Screen Shot 2020-06-24 at 3 18 50 PM

CarsonF: Yep agreed

glassblowerscat (changes requested):
Design:

image

Story:

image

Did we change our minds about this, or did it just happen as a result of passing through the error?

CarsonF:
Yeah I'm ok only loosely matching designs here. I didn't want to display http status codes which shouldn't mean anything to users and we would fake anyways.

glassblowerscat (approved):

glassblowerscat:
@CarsonF I think Frank made your requested changes here; do you want to take another look?

PR #266, opened 2020-06-23
Picture stories

blakekrammes:
Still one thing I want to figure out:

It seems like `objectFit: 'scale-down'` is not being implemented in the `Picture` component, but maybe I'm missing something. There's a problem when I use `'scale-down'`, the background that overlays and darkens the image doesn't map correctly spatially.


CarsonF:
Closing until ready or at least being worked on.

PR #268, opened 2020-06-24
Multiple Form Submit Actions

CarsonF:
This allows us to submit forms different ways. For example, _update_ and _delete_. And it allows FF to still manage the submit progress.

```tsx
<Form>
  onSubmit={(data) => {
    data.submitAction === 'delete' ? doDelete() : doUpdate();
  }}
>
  {({ handleSubmit }) => (
    <form onSubmit={handleSubmit}>
      <SubmitButton />
      <SubmitButton action="delete">Delete</SubmitButton>
    </form>
  )}
</Form>
```

### DIFF
--- a/src/components/form/SubmitButton.stories.tsx
+++ b/src/components/form/SubmitButton.stories.tsx
@@ -1,12 +1,14 @@
+import { Box } from '@material-ui/core';
 import { boolean, select, text } from '@storybook/addon-knobs';
+import { useSnackbar } from 'notistack';
 import React from 'react';
 import { Form } from 'react-final-form';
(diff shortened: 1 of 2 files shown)


CarsonF:
One caveat I just thought of is that the form still has to be valid. So you couldn't remove a required value and then submit a delete action. Maybe this edge case can be handled later.

richardgirges:
> One caveat I just thought of is that the form still has to be valid. So you couldn't remove a required value and then submit a delete action. Maybe this edge case can be handled later.

Definitely sounds like an edge case to me.

This is neat!

PR #304, opened 2020-06-30
[FE 300] Feature: engagement details

frankyma:
This is the engagement details scenes for the language and internship engagements.  The scenes are read-only.

**Technical notes:**

- They are both consumed by the `Engagement` component, queries the engagement and renders the respective scenes.  Loading and error WIP
- The `status` and modifiedAt fields are broken in the BE so have placeholders for that
- I copied over `PersonCard` to create the `MentorCard` because they are slightly different.  The MentorCard has an edit button where the person card has a to: `ButtonLink`.  also the MentorCard design doesn't have the status and slightly different sizing to match the `CeremonyCard`
- The `CeremonyCard` has slightly different display (like Certification date) than the design to use the existing formatter, let me know if you'd like that changed
- I downloaded two icons off figma for the `FieldOverviewCard` but can't seem to size/center them exactly, I can use some help in that.

**The next TODOs** 

1. create the loading an error scenes
2. the edit dialog and mutations to update the engagements
3. storybooks for the cards.
4. The new cards themselves may need loading/no-permission states also.

Since this PR is already so big I think it's best to break that up.

**Testing:**
I used the test suite data because creating new engagement data was erroring.  Then you might have to update the engagements to create data like the ceremony dates.  Load the engagement on /peojects/{projectId}/engagements/{engagementId}

Internship
Screen Shot 2020-06-30 at 12 01 32 PM

Language
Screen Shot 2020-06-30 at 11 59 01 AM


frankyma:
@CarsonF Made the fixes and additions as discussed today:

1. Change the redacted text to cannot read language name or person name
2. Show redacted on the breadcrumb if don't have read permission
3. remove the /projects path on breadcrumbs
4. Accept empty values on `FieldOverviewCardProps`. I passed through a string instead of a node like you mentioned, because it's more consistent with the value, easier to apply markup, and simpler for storybook. This is what it looks like
Screen Shot 2020-07-01 at 11 50 21 AM

5. Fixed the grid spacing to 3
6. On LanguageEngagement add the "Upload PnP" button
7. Use the Material icon for speech bubble con communications complete date (it does look more bold compared to the others)
8. change the 'Visual' methodology category to mui play icon
9. Fix the cardactions borders
10. Added storybook components for Methodology and Ceremony Card. 

 I did run into issues with the methodologies cards: 
On the component itself I couldn't find an easy markup for the list of methodology icons, especially taken into account overflow.  I want to give you a shot before I apply my hacky css
Screen Shot 2020-07-01 at 3 21 43 PM

Also on the storybook I didn't know the best way to add a knob to select multiple values for these methodologies. The typescript also tripped me up.

Next I will work on the loading/redacted scenes and cards, and finally the edit actions.

frankyma:
Added the loading scene on `EngagementDetailLoading`: 
Screen Shot 2020-07-02 at 12 34 32 PM

frankyma:
Added fix to show `status` and `modifiedAt` after BE fixes.  Not sure what the best way to deal with a Ceremony with `null` date values: 
Screen Shot 2020-07-06 at 11 24 54 AM

PR #315, opened 2020-07-07
Upload Manager and Partial Upload Handling

glassblowerscat:
Creates three main components:

- A `Dropzone` for dragging and dropping files to upload (or selecting them through the system file manager)
- An `UploadManager` for viewing the progress of uploaded files
- An `UploadItem`, which shows the state and completion percentage of an uploaded file

For the Dropzone, we are currently not supporting a model in which the user "stages" files and can upload files in multiple passes before submitting all staged files for upload. Any file selected by the user will be uploaded immediately. We are also not limiting the Dropzone to accept only certain file types.

The Upload Manager opens automatically when an upload is started. It can be collapsed into the toolbar, repositioned on screen by dragging the toolbar, and closed—which makes a button appear in the Profile Toolbar to reopen it. 

The Upload Item shows the percentage of the file that is uploaded in a progress bar, along with the name of the file. If an error occurs, an error message will be shown underneath the progress bar; otherwise the text below the progress bar shows messaging to indicate what stage the upload is at.

This also temporarily adds an `/upload` route to the `Root` router. This scene just has a Dropzone on it to allow testing of the upload process.

I think we have a few unanswered questions about some of the behaviors. For example: collapsing the Upload Manager currently collapses it "up" into the toolbar, rather than "down". But I figured these things might be better left for after the designers have done a pass at this.


glassblowerscat:
Superseded by #340; closing

PR #321, opened 2020-07-08
FE 309 list scenes scroll

frankyma:
Fixed the sizing of the scene container and only scroll the list of items.  Added a `div` around the list of items

Screen Shot 2020-07-07 at 6 17 15 PM

**Testing**
Open any list scene with items that overflows the page.  The items themselves scroll but not the headers.


CarsonF (dismissed):
Padding looks weird both for scroll bar and for bottom of list being cut off.
Screen Shot 2020-07-09 at 6 09 24 PM

frankyma:
Removed the bottom and right padding on the outside container.  Kept the inner list padding to show card shadow.

CarsonF:
Closing for now since no work is being done. Reopen if you make changes.

frankyma:
@CarsonF I thought this was pending review?  I don't remember change requests

CarsonF:
Sorry I thought we talked about it on slack/zoom. The scroll bar was spaced away from the edge of the window if I remember correctly. I think there might have been something else with the scrolling and bottom of the window, but you might have fixed it.

The project details scene is styled correctly if you want to match that.

frankyma:
I remember where we left off.  I fixed the scroll issues but couldn't align the cards, I asked you for some help but you were busy at the time.  I fixed the alignment now with a simple negative margin.

Here's the final view:
ezgif com-video-to-gif (1)


By the way this isn't styled the same as project details because the list item cards themselves scroll. It's one container level deeper than the detail scenes.  For this case I'm using overflow: 'hidden' on the parent container and overflow: 'auto' on the child list container.

frankyma:
Also changed all the overfills to 'auto' from 'scroll' to avoid this (this is on v3 remote)
Screen Shot 2020-09-08 at 3 45 48 PM

src/components/ContentContainer/ContentContainer.tsx
```diff
@@ -1,12 +1,13 @@
...
+const useStyles = makeStyles(({ spacing }) => ({
   root: {
     flex: 1,
-    overflowY: 'scroll',
-    padding: theme.spacing(4),
-    maxWidth: theme.breakpoints.values.md,
+    overflow: 'hidden',
+    padding: spacing(4, 0, 0, 4),
+    display: 'flex',
+    flexDirection: 'column',
```
glassblowerscat, line 10: This seems effective for the problem we're trying to solve, but I think we need to add some styling to `ProjectFilesList` to counteract this:

image

glassblowerscat: It's supposed to look like this:

image

glassblowerscat: Could just be as simple as wrapping the table in a `Grid`

src/scenes/Languages/List/LanguageList.tsx
```diff
@@ -14,6 +14,14 @@ const useStyles = makeStyles(({ spacing }) => ({
...
     margin: spacing(3, 0),
   },
+  listContainer: {
+    overflow: 'auto',
+    marginLeft: spacing(-2),
+    padding: spacing(2),
+  },
+  item: {
+    marginBottom: spacing(2),
+  },
```
glassblowerscat, line 24: We repeat this a lot. Can we break this out into a simple `ListScrollContainer` or similar? It might be worth moving that new component and `ContentContainer` into a new `components/Layout` directory; what do you think?

glassblowerscat (changes requested):

src/components/ContentContainer/ContentContainer.tsx
frankyma: I just gave the table some padding on the right and bottom
Screen Shot 2020-09-18 at 1 54 21 PM (2).  Don't think we need Grid for this since it's just a single item.  Let me know what you think

src/scenes/Languages/List/LanguageList.tsx
frankyma: good idea, implemented.

src/components/ContentContainer/ContentContainer.tsx
glassblowerscat: What I was thinking with the `Grid` was that we ought to keep it at that 2/3-ish width it was at before... thought I remembered that being something @CarsonF and I discussed, but maybe not. Care to weigh in, Carson?

glassblowerscat (approved):

PR #324, opened 2020-07-08
FE 323 unique member key

frankyma:
Fixed unique key react bug.  The on the project overview page, the project partnerships were using the organization name as the unique mapping key.

Testing:
Create two partnerships with the same projectId and organizationID (can be done in playground).  Then go to that project overview page on the browser, you should see the two partnerships with the same org name.  In the devtools console you should **not** see the React same key warning.

### DIFF
--- a/src/components/ProjectMembersSummary/ProjectMembersSummary.tsx
+++ b/src/components/ProjectMembersSummary/ProjectMembersSummary.tsx
@@ -11,13 +11,15 @@ export interface ProjectMembersSummaryProps {
 export const ProjectMembersSummary: FC<ProjectMembersSummaryProps> = ({
   members,
 }) => {
-  const summarizedMembers = members?.items.map(
-    ({ user }): MemberSummaryItem => ({
-      avatarLetters: user?.value?.avatarLetters ?? undefined,
-      label: user?.value?.firstName ?? '',
-      id: user?.value?.id || '',
-    })
-  );
+  const summarizedMembers = members?.items
+    .filter(({ user }) => user.canRead)
+    .map(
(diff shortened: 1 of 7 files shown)


CarsonF:
Organization names are unique. A single organization cannot be in two different partnerships on a project, it's unique as well. Therefore name should be fine.

I know that the API has a bug right now that currently returns duplicates. https://github.com/SeedCompany/cord-api-v3/issues/691

This still might be worth while though for the cases where there isn't unique labels - like user's first name.

frankyma:
> Organization names are unique. A single organization cannot be in two different partnerships on a project, it's unique as well. Therefore name should be fine.

I was actually able to create two partnerships on the same project using the playground, so we'll have to make sure that doesn't happen on the FE.

>This still might be worth while though for the cases where there isn't unique labels - like user's first name.

I agree it's just better practice to use the uid as the key instead of a label

src/components/ProjectMembersSummary/ProjectMembersSummary.tsx
```diff
@@ -15,6 +15,7 @@ export const ProjectMembersSummary: FC<ProjectMembersSummaryProps> = ({
     ({ user }): MemberSummaryItem => ({
       avatarLetters: user?.value?.avatarLetters ?? undefined,
       label: user?.value?.firstName ?? '',
+      id: user?.value?.id || '',
```
CarsonF, line 18: The member list should probably be filtered to only the members that the current user has access to read.

frankyma: Added the filter

PR #328, opened 2020-07-08
Bugfix FE 311 validate name fields on register

frankyma:
1. Match the text field names with the api field names
2. Show display first name and display last name fields
3. Enforce minimum of 2 chars for name fields

Testing:
On register, must have 2 or more characters for all the name fields
Screen Shot 2020-07-08 at 3 10 35 PM

### DIFF
--- a/src/scenes/Authentication/Register/Register.tsx
+++ b/src/scenes/Authentication/Register/Register.tsx
@@ -21,17 +21,12 @@ export const Register = (props: Except<Props, 'onSubmit'>) => {
     }
   }, [navigate, session, sessionLoading, success]);
 
-  const submit: Props['onSubmit'] = async (input) => {
+  const submit: Props['onSubmit'] = async ({ confirmPassword, ...input }) => {
     try {
       const { data } = await register({
         variables: {
           input: {
-            displayFirstName: input.firstName,
-            displayLastName: input.lastName,
-            realFirstName: input.firstName,
-            realLastName: input.lastName,
-            email: input.email,
-            password: input.password,
+            ...input,
             timezone: DateTime.local().zone.name,
           },
         },
--- a/src/scenes/Authentication/Register/RegisterForm.tsx
+++ b/src/scenes/Authentication/Register/RegisterForm.tsx
@@ -2,6 +2,7 @@ import { makeStyles, Typography } from '@material-ui/core';
 import { Decorator, Mutator } from 'final-form';
 import React from 'react';
 import { Form, FormProps } from 'react-final-form';
+import { RegisterInput } from '../../../api';
 import {
   blurOnSubmit,
   EmailField,
@@ -38,18 +39,12 @@ const useStyles = makeStyles(({ spacing }) => ({
   },
 }));
 
-interface RegisterInput {
-  realFirstName: string;
-  realLastName: string;
-  displayFirstName: string;
-  displayLastName: string;
-  email: string;
-  password: string;
+interface RegisterInputFields extends RegisterInput {
   confirmPassword: string;
 }
 
 export type RegisterFormProps = Pick<
-  FormProps<RegisterInput>,
+  FormProps<RegisterInputFields>,
   'onSubmit' | 'initialValues'
 > & { className?: string };
 
@@ -127,7 +122,7 @@ const passwordMatching = ({
   password,
   confirmPassword,
   ..._otherVals
-}: RegisterInput) => {
+}: RegisterInputFields) => {
   return password && confirmPassword && password !== confirmPassword
     ? {
         password: 'Passwords must match',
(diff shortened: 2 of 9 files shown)


src/scenes/Authentication/Register/RegisterForm.tsx
```diff
@@ -105,17 +119,32 @@ export const RegisterForm = ({ className, ...props }: RegisterFormProps) => {
...
+  if (!realFirstName || realFirstName.length < 2)
+    errors.realFirstName = 'First Name must be longer than 2 characters';
+  if (!realLastName || realLastName.length < 2)
+    errors.realLastName = 'Last Name must be longer than 2 characters';
+  if (!displayFirstName || displayFirstName.length < 2)
+    errors.displayFirstName =
+      'Display First Name must be longer than 2 characters';
+  if (!displayLastName || displayLastName.length < 2)
+    errors.displayLastName =
+      'Display Last Name must be longer than 2 characters';
```
CarsonF, line 146: These should be done with field validators. See functions in this file and their usages https://github.com/SeedCompany/cord-field/blob/636c73f880d1dd864319aae3c6640db4ae3eb10e/src/components/form/validators.ts

src/scenes/Authentication/Register/RegisterForm.tsx
```diff
@@ -66,26 +68,38 @@ export const RegisterForm = ({ className, ...props }: RegisterFormProps) => {
...
               <TextField
-                name="lastName"
+                name="realLastName"
                 label="Last Name"
                 placeholder="Enter Last Name"
                 required
               />
+              <TextField
+                name="displayFirstName"
+                label="Display First Name"
```
CarsonF, line 93: Let's call it _Public_ instead of _Display_, just like create/update user does.

src/scenes/Authentication/Register/RegisterForm.tsx
```diff
@@ -38,8 +38,10 @@ const useStyles = makeStyles(({ spacing }) => ({
 }));
 
 interface RegisterInput {
```
CarsonF, line 40: This interface could be pulled from the schema or at least extend from it since we need to add `confirmPassword`.

frankyma:
> Let's sync the real and display name fields which the form decorators, just like create/update user does.

Decorator added

src/scenes/Authentication/Register/RegisterForm.tsx
frankyma: Done

src/scenes/Authentication/Register/RegisterForm.tsx
frankyma: Changed

src/scenes/Authentication/Register/RegisterForm.tsx
frankyma: Pulled from schema

src/components/form/validators.ts
```diff
@@ -37,3 +37,6 @@ export const min = (min: number, error?: string) => (
 export const max = (max: number, error?: string) => (
   val: number | null | undefined
 ) => (val != null && val > max ? error ?? 'Value is above maximum' : undefined);
+
+export const twoOrMoreChars = (value: string) =>
+  !value || value.length >= 2 ? undefined : 'Must be 2 or more characters';
```
CarsonF, line 42: How about
```suggestion
export const minLength = (min = 2) => (value: string) =>
  !value || value.length >= min ? undefined : `Must be ${min} or more characters`;
```

src/scenes/Authentication/Register/Register.tsx
```diff
@@ -26,10 +26,10 @@ export const Register = (props: Except<Props, 'onSubmit'>) => {
...
         variables: {
           input: {
-            displayFirstName: input.firstName,
-            displayLastName: input.lastName,
-            realFirstName: input.firstName,
-            realLastName: input.lastName,
+            displayFirstName: input.realFirstName,
+            displayLastName: input.realLastName,
+            realFirstName: input.displayFirstName,
+            realLastName: input.displayLastName,
```
CarsonF, line 32: Hmm 👀

frankyma: I used a spread operator but the input had the confirmPassword so I got lazy

src/components/form/validators.ts
frankyma: done, good idea

src/scenes/Authentication/Register/Register.tsx
CarsonF: Well I was attempting to point out that you've flipped the display & real fields.

But I think spreading would be better anyways.
```tsx
  const submit: Props['onSubmit'] = async ({ confirmPassword, ...input }) => {
    try {
      const { data } = await register({
        variables: {
          input: {
            ...input,
            timezone: DateTime.local().zone.name,
          },
        },
      });
```

frankyma: This is why you're the tech lead.  Fixed

PR #330, opened 2020-07-09
Project file/folder list and routing

### LINKED ISSUE #295: File List
- [x] url: /projects/{projectId}/files
- [x] Needs a `parent` query parameter to control the current sub directory (if any)
- [x] I would imagine sub directories are represented by `Breadcrumb`s and can be retrieved via `directory.parents`
- [x] File table implemented with #297 

Screenshot from v2
Screen Shot 2020-07-07 at 2 55 46 PM

glassblowerscat:
Resolves #295 

- Creates a `ProjectFilesList` component, which displays a table with the contents of the directory currently being viewed
- Implements `/projects/:projectId/folders/:folderId` routing scheme for routing the user to different folder views based on their selection of a given folder in the UI. The `:folderId` is checked to make sure it is a folder associated with the `:projectId`
- Shows breadcrumbs for the current folder and its parents, up to the root "Files" page for the project

This also re-implements the `Table` component created in PR #256, which is still blocked and in draft mode. 

This PR is just to handle the directory querying/navigation/routing; more work remains to be done on the file table itself. We have not implemented file previews, downloads, history, or deletions. We also need to decide whether we are using a single-click (open immediately) or double-click (select row then open) behavior for the individual table rows. We also need to wire up the file upload/new folder UI shown here:

86835698-17fce880-c062-11ea-9aa7-cfd60db22e9c


glassblowerscat:
Superseded by #340; closing

PR #333, opened 2020-07-10
Bugfix FE 329 language schema breaking changes create language

frankyma:
Updated the Create Language functionality in response to the schema change

**Technical Notes**
Copied over the `UserForm` implementation of the `SecuredField`s.
Not sure what the order, label and description wording of the fields should be, please let me know if that should be changed.

**Testing**
Create language form looks like screenshot and works (You may get a generic error message but if you query the list you'll see the language was created

There's validators on most fields for min chars and number/character inputs

Screen Shot 2020-07-10 at 12 13 32 PM


CarsonF:
Looks like this now

Screen Shot 2020-07-16 at 5 31 15 PM

PR #336, opened 2020-07-14
Bugfix: FE 334 edit user crash

frankyma:
App was crashing because the UserForm didn't wrap onSubmit in an async await, so the function resolved immediately and the error was never caught by the try catch in the DialogForm (instead caught by CRA)

Hide the email field on UserForm during edit user.  Showing but disabling is more involved because on the EditUser component the email field is not recognized by typescript because it's reading the input type as `UpdateUserInput`.

### DIFF
--- a/src/scenes/Users/Detail/UserDetail.generated.ts
+++ b/src/scenes/Users/Detail/UserDetail.generated.ts
@@ -3,8 +3,11 @@ import * as ApolloReactCommon from '@apollo/client';
 import * as ApolloReactHooks from '@apollo/client';
 import gql from 'graphql-tag';
 import * as Types from '../../../api/schema.generated';
-import { UserFormFragment } from '../UserForm/UserForm.generated';
-import { UserFormFragmentDoc } from '../UserForm/UserForm.generated';
+import { SsFragment, UserFormFragment } from '../UserForm/UserForm.generated';
+import {
+  SsFragmentDoc,
+  UserFormFragmentDoc,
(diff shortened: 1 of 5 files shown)


frankyma:
> What about pulling `email` out of input on update?
> 
> ```tsx
> onSubmit={async ({ email, ...input }) => {}}
> ```

I tried that but get `Property 'email' does not exist on type 'UpdateUserInput' because it's reading input as `UpdateUserInput` from `<UserForm<UpdateUserInput>`

> Alternatively you could check the `initialValues` prop in `UserForm` we only have it on update not create so that could be used instead of adding another prop.

I can't get TS to give me the `initialValues?.person.email` value `Property 'user' does not exist on type 'NonNullable<T>'.ts(2339)`

CarsonF:
> I tried that but get `Property 'email' does not exist on type 'UpdateUserInput'` because it's reading input as `UpdateUserInput` from `<UserForm<UpdateUserInput>`

You could define the form values with that field
```tsx
<UserForm<CreatePersonInput & { person: { email: string } }>
```

That's probably easiest.

Also I'm just realizing that email is optional on creating a user so we actually do need it to be updatable to some extent.

frankyma:
> ```tsx
> <UserForm<CreatePersonInput & { person: { email: string } }>
> ```

Good call, applied as suggested.  I also updated the fragments and queries so the UI on the detail and formscenes  updates after the mutation.

PR #337, opened 2020-07-14
Autocomplete field and OrganizationAutocomplete

blakekrammes:
Autocomplete component and implementation with the Organization list API.


src/components/OrganizationAutocomplete/OrganizationAutocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+>(
+  props: OrganizationAutocompleteProps<T, Multiple, DisableClearable, FreeSolo>
+) {
+  const [value, setValue] = useState('');
+  const [options, setOptions] = useState<any[]>([]);
+  const [selectedOptions, setSelectedOptions] = useState<any[]>([]);
+  const [searching, setSearching] = useState(false);
+
+  // if we don't want to show any initial options, it'd be nice not to have to call the api before input
+  const { refetch } = useOrganizationAutocompleteQuery({
```
CarsonF, line 31: What about the lazy hook?
```suggestion
  const [fetch, result] = useOrganizationAutocompleteLazyQuery({
```

src/components/OrganizationAutocomplete/OrganizationAutocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+        },
+      });
+      setOptions(
+        data?.organizations?.items.map((item) => item.name.value) || []
+      );
+      setSearching(false);
+    },
+    500,
+    [searching, value]
+  );
```
CarsonF, line 58: It would be better to have this debounce in the AutocompleteField, at least as an option. That way every wrapper doesn't have to implement it.

src/components/OrganizationAutocomplete/OrganizationAutocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <AutocompleteField
+      {...props}
+      hideNoOptionsDiv={!searching && value === ''}
+      options={[...options, ...selectedOptions]}
+      filterSelectedOptions
+      loading={searching}
+      // we do this and set filterSelectedOptions to true ^ to avoid an error in the console about selected options not appearing in the list
+      onOptionSelect={(options: any[]) => {
+        setSelectedOptions(options);
+      }}
```
CarsonF, line 70: I think that's what the `getOptionSelected` prop is for.

src/components/OrganizationAutocomplete/OrganizationAutocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+      const { data } = await refetch({
+        input: {
+          filter: {
+            name: value,
+          },
+        },
+      });
+      setOptions(
+        data?.organizations?.items.map((item) => item.name.value) || []
+      );
```
CarsonF, line 53: We want the options to be the objects, that way we can get the ID from it.

src/components/OrganizationAutocomplete/OrganizationAutocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+        if (trimmed.length <= 1) {
+          setOptions([]);
+        }
+        if (trimmed === '') {
+          setSearching(false);
+        } else {
+          setSearching(true);
+        }
+        setValue(val);
+      }}
```
CarsonF, line 84: This should also probably be apart of the AutocompleteField so every wrapper doesn't have to implement this.

src/components/OrganizationAutocomplete/OrganizationAutocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+  T,
+  Multiple extends boolean | undefined,
+  DisableClearable extends boolean | undefined,
+  FreeSolo extends boolean | undefined
+>(
+  props: OrganizationAutocompleteProps<T, Multiple, DisableClearable, FreeSolo>
+) {
+  const [value, setValue] = useState('');
+  const [options, setOptions] = useState<any[]>([]);
+  const [selectedOptions, setSelectedOptions] = useState<any[]>([]);
```
CarsonF, line 27: `any` tisk tisk

src/components/form/AutocompleteField.tsx
```diff
@@ -0,0 +1,125 @@
...
+  T,
+  Multiple extends boolean | undefined,
+  DisableClearable extends boolean | undefined,
+  FreeSolo extends boolean | undefined
+> = Except<
+  FieldConfig<Value<T, Multiple, DisableClearable, FreeSolo>>,
+  'multiple'
+> & {
+  name?: string;
+  // https://github.com/mui-org/material-ui/issues/19354 -- workaround for setting multiple, freeSolo and disableClearable in a wrapped Autocomplete
```
CarsonF, line 24: That issue has been fixed...

src/components/form/AutocompleteField.tsx
```diff
@@ -0,0 +1,125 @@
...
+  isMultiple?: boolean;
+  isFreeSolo?: boolean;
+  isDisableClearable?: boolean;
```
CarsonF, line 27: I think we should try to stick to the props MUI provides to reduce confusion

src/components/form/AutocompleteField.tsx
```diff
@@ -0,0 +1,125 @@
...
+      multiple={isMultiple}
+      defaultValue={defaultValue}
+      freeSolo={isFreeSolo}
+      disableClearable={isDisableClearable}
+      {...autocompleteProps}
+      classes={{
+        input: classes.autocompleteInput,
+        // hide the "No Options" div when we haven't typed anything yet. Doesn't make sense to show this if we haven't given any input
+        // unless we want to populate an autocomplete with options on initial render any input is given
+        noOptions: hideNoOptionsDiv && classes.noOptionsDiv,
```
CarsonF, line 93: There's got to be a better way than this. The arrow also changes directions indicating that the popup is open, but it's "not" because have it hidden.

src/components/form/AutocompleteField.tsx
```diff
@@ -0,0 +1,125 @@
...
+      }}
+      renderTags={(values: T[], getTagProps) =>
+        values.map((option: T, index: number) => (
+          <Chip variant="outlined" label={option} {...getTagProps({ index })} />
```
CarsonF, line 97: outlined?

We should probably let this be overridable too.

src/components/form/AutocompleteField.tsx
```diff
@@ -0,0 +1,125 @@
...
+        ))
+      }
+      onChange={(_, values: any[], __) => {
+        input.onChange(values);
+        if (!onOptionSelect) return;
+        onOptionSelect(values);
```
CarsonF, line 103:
```suggestion
        onOptionSelect?.(values);
```

src/components/OrganizationAutocomplete/OrganizationAutocomplete.tsx
blakekrammes: What are we using the id for? And I think the options passed to Autocomplete have to be strings.

CarsonF: The IDs are what are sent to the API. Nope they don't have to be strings. Several props are functions given a `T` generic object. In this case `T` will be `Organization`.

blakekrammes: By options do you mean the literal `options` prop?

Issue #297, opened 2020-06-25
File List Table Card


michaelmarshall:
- [ ] Displays a directory's children
- I imagine it being in a card
- I think we talked about removing the `type` column and just adding a icon before the file name based on the `node.category`.

Screenshot from v2
![Screen Shot 2020-07-07 at 2 55 46 PM](https://user-images.githubusercontent.com/932566/86835698-17fce880-c062-11ea-9aa7-cfd60db22e9c.png)

glassblowerscat:
Popup menu contents

![image](https://user-images.githubusercontent.com/728777/87209688-cdd56a80-c2e0-11ea-8634-bbf8a1add129.png)

PR #340, opened 2020-07-16
Files Scene

### LINKED ISSUE #295: File List

glassblowerscat:
Resolves #295 and #297 

# Overview

Implements file uploading and a "Project Files" scene that displays a table of files associated with the project. We use a `/projects/:projectId/files/:folderId` routing scheme for routing the user to different folder views based on their selection of a given folder in the UI. The `:folderId` is checked to make sure it is a folder associated with the `:projectId`

We show breadcrumbs for the current folder and its parents, up to the root "Files" page for the project. This also re-implements the `Table` component created in PR #256, which is still blocked and in draft mode. 

## Project Files Scene

The Project Files scene has the following features:

- "Upload files" and "New folder" buttons that open dialogs to perform the respective actions
- Clicking the name of a folder opens the folder using the routing scheme as above. Clicking a file does nothing at the moment. We will implement file previews in a future PR
- Each file/folder row in the table has a "more" button with actions that can be performed for that type of item. All these actions are implemented, although changes will probably need to be made after feedback from the design team:
  - Folders:
    - Rename
    - Delete
  - Files:
    - Rename
    - Download
    - Delete
    - View History
    - Create New Version
- The version history for a file displays the following:
  - Date of version
  - User who created version
  - File name of version
  - Icon representing `category` of version
  - Download button for version
  
## Uploads

File uploads are implemented through an `Dropzone` field component and an `UploadManager` for viewing the progress of uploaded files. We do not currently offer the option to rename a file during the upload process; files are uploaded immediately upon being dragged into the dropzone.

The Upload Manager opens automatically when an upload is started. It can be collapsed into the toolbar, repositioned on screen by dragging the toolbar, and closed—which makes a button appear in the Profile Toolbar to reopen it. 

The Upload Item shows the percentage of the file that is uploaded in a progress bar, along with the name of the file. If an error occurs, an error message will be shown underneath the progress bar; otherwise the text below the progress bar shows messaging to indicate what stage the upload is at.

I think we have a few unanswered questions about some of the behaviors. For example: collapsing the Upload Manager currently collapses it down into the bottom-right corner and makes it no longer draggable. But I figured these things might be better left for after the designers have done a pass at this.


src/components/Upload/Reducer/uploadInitialState.ts
```diff
@@ -0,0 +1,3 @@
+export const initialState = {
```
CarsonF, line 1:
```suggestion
export const initialState: UploadState = {
```

src/components/Upload/Reducer/uploadTypings.ts
```diff
@@ -0,0 +1,76 @@
...
+  uploading: boolean;
+}
+
+export interface UploadState {
+  submittedFiles: UploadFile[];
+}
+
+interface FileAction {
+  queueId: UploadFile['queueId'];
+}
```
CarsonF, line 32:
```suggestion
interface FileAction extends Pick<UploadFile, 'queueId'> {
}
```

src/components/Upload/UploadContext.tsx
```diff
@@ -0,0 +1,225 @@
+import React, {
```
CarsonF, line 1: There's a lot going on in this file. Can it be broken up more? UploadManager state/context and the xhr request seem like easy picks to move to a different file.

src/components/Upload/UploadContext.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+interface UploadManagerContextValue {
+  isManagerOpen: boolean;
+  setIsManagerOpen: (isOpen: boolean) => void;
+}
+
+const initialUploadManagerContext = {
+  isManagerOpen: false,
+  setIsManagerOpen: () => null,
+};
```
CarsonF, line 43: Personally I don't like having to define the types again after defining the initial state. What I like to do is use the initial state to define the shape.
```suggestion
const initialUploadManagerContext = {
  isManagerOpen: false,
  setIsManagerOpen: (isOpen: boolean) => {
    // noop
  },
};
type UploadManagerContextValue = typeof initialUploadManagerContext;
```
Better still since the interface/type is only passed to the `createContext` function, it can just inferred the type from the initial state shape.

Really the trick is to make sure the initial state props declare the types correctly. Functions usually don't match so I've done this before too:
```tsx
  setIsManagerOpen: noop as (isOpen: boolean) => void
```

src/components/Upload/UploadContext.tsx
```diff
@@ -0,0 +1,225 @@
...
+          } else {
+            handleFileUploadCompleteError(xhr.statusText, queueId);
+          }
+        }
+      };
+
+      xhr.upload.onprogress = handleFileProgress.bind(null, queueId);
+      xhr.upload.onerror = () => handleUploadError(queueId);
+      xhr.open('PUT', url);
+      xhr.setRequestHeader('Access-Control-Allow-Headers', '*');
```
CarsonF, line 163: This is a response header, and does nothing in requests.
```suggestion
```

src/components/Upload/UploadFilesForm.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { Except } from 'type-fest';
+import { DialogForm, DialogFormProps } from '../Dialog/DialogForm';
+import { DropzoneField, SubmitError } from '../form';
+import { UploadCallback, useUpload } from '../Upload';
+
+export type UploadFilesFormProps = DialogFormProps<{ files: File[] }> & {
+  callback: UploadCallback;
```
CarsonF, line 8: I think this prop could be named better. Following the on* convention would be nice but I can't think of a name that fits. `finalizeUpload` seems nice.

src/components/Upload/UploadFilesForm.tsx
```diff
@@ -0,0 +1,42 @@
...
+  multiple?: boolean;
+  title?: string;
```
CarsonF, line 10: DialogFromProps should already have this prop defined.
```suggestion
```

src/components/Upload/UploadFilesForm.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  return (
+    <DialogForm
+      DialogProps={{
+        fullWidth: true,
+        maxWidth: 'xs',
+      }}
+      {...props}
+      onSubmit={onSubmit}
+      title={title}
```
CarsonF, line 36: To go along with the change above you can just set default here and props can still override it.
```suggestion
      title="Upload Files"
      {...props}
      onSubmit={onSubmit}
```

src/components/Upload/UploadItem.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const progressLabel = error
+    ? error.message
+    : !uploadId
+    ? 'Initializing'
+    : percentCompleted === 100
+    ? 'Completed'
+    : 'Uploading';
+
+  return (
+    <Box
```
CarsonF, line 58: I still don't want to use `Box` since it creates new css with every mount. I think `<Grid>` should get you pretty close to what you are already doing here.

src/components/Upload/UploadItems.tsx
```diff
@@ -0,0 +1,49 @@
+import { Box, makeStyles, Typography } from '@material-ui/core';
+import React, { FC } from 'react';
+import { UploadFile, UploadState } from './Reducer/uploadTypings';
+import { UploadItem } from './UploadItem';
+
+const useStyles = makeStyles(({ palette }) => ({
+  noUploadsText: {
+    color: palette.grey[400],
```
CarsonF, line 8:
```suggestion
    color: palette.action.disabled,
```

src/components/Upload/UploadManager.tsx
```diff
@@ -0,0 +1,235 @@
...
+  root: {
+    pointerEvents: 'none',
+  },
+  paper: {
+    margin: 0,
+    pointerEvents: 'auto',
+    position: 'absolute',
+    top: 0,
+    left: 0,
+    width: `${PAPER_WIDTH}px`,
```
CarsonF, line 42: FYI pixels are assumed for numbers.
```suggestion
    width: PAPER_WIDTH,
```

src/components/Upload/UploadManager.tsx
```diff
@@ -0,0 +1,235 @@
...
+  },
+  titleContainer: {
+    cursor: 'move',
+    display: 'flex',
+    justifyContent: 'space-between',
+    alignItems: 'center',
+    padding: spacing(1),
+    '&.collapsed': {
```
CarsonF, line 50: You can add an empty collapsed class here and then reference it like `&$collapsed`. [MUI](https://github.com/mui-org/material-ui/blob/50c0bbc3a2040c9945f5009a6bd56292d216f236/packages/material-ui/src/InputLabel/InputLabel.js) does this too.

src/components/Upload/UploadManager.tsx
```diff
@@ -0,0 +1,235 @@
...
+    },
+  },
+  noUploadsText: {
+    color: palette.grey[400],
+  },
+}));
+
+const initialPosition = { x: null as null | number, y: null as null | number };
+
+const PaperComponent: FC<PaperProps & { isCollapsed: boolean }> = ({
```
CarsonF, line 78: Could this be moved to another file. Maybe `DraggablePaper`?

src/components/Upload/UploadManager.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+  const { width: windowWidth, height: windowHeight } = useWindowSize();
+  const managerSize = {
+    width: (paperRef.current?.offsetWidth ?? 0) + 24,
+    height: (paperRef.current?.offsetHeight ?? 0) + 24,
+  };
+  const collapsedPosition = {
+    x: windowWidth - managerSize.width,
+    y: windowHeight - managerSize.height,
+  };
```
CarsonF, line 95: Not sure where to put this but I don't think we need a separate position for the collapsed state. Having it collapse/expand where ever it seems fine. What do you think?

src/components/Upload/UploadManager.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const [position, setPosition] = useState({ x: 0, y: 0 });
+  const [isDragging, setIsDragging] = useState(false);
+  const paperRef = useRef<HTMLDivElement>(null);
+  const positionRef = useRef(initialPosition);
+  const { isCollapsed, ...rest } = props;
+
+  const { width: windowWidth, height: windowHeight } = useWindowSize();
+  const managerSize = {
+    width: (paperRef.current?.offsetWidth ?? 0) + 24,
+    height: (paperRef.current?.offsetHeight ?? 0) + 24,
```
CarsonF, line 90: What's the 24 here for? Margin to add spacing between the edge of the screen? Maybe that could just be applied to the paper outright? That would enforce the spacing when dragging which seems nice too.

src/components/Upload/UploadManager.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const classes = useStyles();
+
+  function handleClose(event: React.MouseEvent<HTMLButtonElement>) {
+    event.stopPropagation();
+    setIsManagerOpen(false);
+  }
+
+  function handleCollapse(event: React.MouseEvent<HTMLButtonElement>) {
+    event.stopPropagation();
+    console.log('Changing collapsed');
```
CarsonF, line 204:
```suggestion
```

src/scenes/Root/Header/ProfileToolbar/ProfileToolbar.tsx
```diff
@@ -46,6 +53,11 @@ export const ProfileToolbar: FC = () => {
         <IconButton>
           <NotificationsNone />
         </IconButton>
+        {!isManagerOpen && (
+          <IconButton onClick={() => setIsManagerOpen(true)}>
+            <CloudUpload />
+          </IconButton>
+        )}
```
CarsonF, line 60: I wonder if it would be better to put this in the context menu to the right. That seems better than adding another icon here and would give one of the existing icons something to do.

src/components/Formatters/useFileSizeFormatter.ts
```diff
@@ -0,0 +1,13 @@
...
+    if (bytes === 0) {
+      return `${bytes} ${sizes[0]}`;
+    }
+    const sizeIndex = Number(Math.floor(Math.log(bytes) / Math.log(1024)));
+    return `${Math.round(bytes / Math.pow(1024, sizeIndex))} ${
+      sizes[sizeIndex]
+    }`;
+  };
+  return formatFileSize;
+};
```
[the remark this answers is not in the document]
glassblowerscat: Correct, this could just be a regular function. I thought I had seen another thing like this that had been made as a hook, so I figured we were just doing that for semantic consistency. But it could totally be a function

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,109 @@
+import { makeStyles, Typography } from '@material-ui/core';
+import { ArrowDownward, Check, Clear, Edit } from '@material-ui/icons';
+import MaterialTable, { Column, Icons } from 'material-table';
+import React, { FC, forwardRef } from 'react';
+
+export interface RowData {
+  [key: string]: any;
+}
```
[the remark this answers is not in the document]
glassblowerscat: I don't mind at all if you want to work on it, but I can also do it if you want to focus on other things

src/components/Upload/UploadContext.tsx
glassblowerscat: Oh, interesting; I've never seen that before

src/components/Upload/UploadManager.tsx
glassblowerscat: I tried this and the styles didn't take, but I'll have to look at it again... maybe I got the syntax wrong

src/components/Upload/UploadManager.tsx
glassblowerscat: I could go either way on it, so I went the easier way 🤣 If we let it stay where it is when collapsed, we have to do extra math + logic to keep track of its height. So I figured I'd let the design team tell me if they wanted that. Do you want to go ahead and do it, or keep this the way it is for now?

src/components/Upload/UploadManager.tsx
glassblowerscat: I'll have to look at this again. You're right that that's what it's for, and there was a reason I now can't remember for why I hard-coded this in here. But I'm not crazy about it

src/components/Upload/UploadManager.tsx
CarsonF: You also need to apply it with `classes.collapsed` instead of the `"collapsed"` string constant.

src/components/Upload/UploadManager.tsx
CarsonF: Is that not happening already? The height is dynamic right?

src/components/Upload/UploadManager.tsx
CarsonF: Mmm offsetWidth/Height doesn't include margin. maybe we could wrap it again and use padding so we don't have to apply it ourselves?

src/components/Upload/UploadContext.tsx
glassblowerscat: @CarsonF I tried this with the `noop` function from `ts-essentials`, and TypeScript yelled at me; were you using a different `noop` function?

CarsonF: Umm lodash. But both should work. What's the problem?

glassblowerscat: Nm, I figured it out:

`setIsManagerOpen: (_: boolean) => { return; },`

src/components/Upload/UploadManager.tsx
glassblowerscat: I remember now... TypeScript doesn't like this; it can't tell that `classes.collapsed` exists

src/components/Upload/UploadManager.tsx
glassblowerscat: Okay, we still need these values in order to keep the spacing between the edge of the screen because `margin` isn't included in `offsetWidth` or `offsetHeight`, but now it's calculated dynamically instead of hard-coded

glassblowerscat:
@CarsonF I think this is ready to review again. I believe I've done all the changes we discussed. I didn't add any features that weren't already in this branch, so no file previews, and "File History" only offers a download link, not the full set of actions. This is just the UI/UX changes we discussed. Let me know if I'm missing anything

glassblowerscat:
@CarsonF This is rebased and ready again

PR #341, opened 2020-07-16
show organizations on user details

### LINKED ISSUE #314: Display Roles & Organizations on User Detail

richardgirges:
Closes #314 

Not showing roles, as discussed over slack, since roles don't exist on the API side yet.


src/scenes/Users/Detail/UserDetail.tsx
```diff
@@ -111,6 +115,24 @@ export const UserDetail = () => {
             loading={!user}
           />
           {user ? <EditUser user={user} {...editUserState} /> : null}
+
+          <DisplayProperty
+            label="Partners"
+            value={
+              user?.organizations.items.length ? (
+                <Box mt={1}>
```
CarsonF, line 123: Let's not use Box outside of StoryBook. It creates a new css class every time it's mounted.

richardgirges: Got it, so just use div with `className` prop?

CarsonF: Yeah that would be great :)

richardgirges: Just updated to div+class instead of Box

src/scenes/Users/Detail/UserDetail.tsx
```diff
@@ -111,6 +118,24 @@ export const UserDetail = () => {
             loading={!user}
           />
           {user ? <EditUser user={user} {...editUserState} /> : null}
+
+          <DisplayProperty
+            label="Partners"
+            value={
```
CarsonF, line 124: Idk if using `DisplayProperty` is the best thing here. It's more designed for simple values instead of a list of cards. 
For example, right now the loading state is completely different.

richardgirges: Got it, updated!

glassblowerscat (approved):

Issue #344, opened 2020-07-17
Language Engagement Products


frankyma:
Create, Edit, Product Card

https://www.figma.com/file/IfEmcJeuNY6ucgjYctlKB7/Cord-Field?node-id=2848%3A13293

PR #347, opened 2020-07-22
FE #344 Create and Edit Product

frankyma:
**Technical Notes**
Create and Edit functionality.

Using radio fields for the single select as discussed, but had to allow the RadioField to allow `required=false` in the prop or else can't continue to submit the fields that are not required.

I couldn't figure out the secured fields for produces and scripture references because they are product type dependent.  Could use some help on that please.

On Edit, product type and scripture references won't be shown with previous values because there's some uncertainty on how to show those, and the return value isn't ready on the API.

The UpdateProduct endpoint isn't ready yet also added as TODO.

The styling is a little rough, would like Patrick to take a look and give some input.
Screen Shot 2020-07-30 at 5 33 17 PM

**Testing:**
need to run yarn to update the mui package, and make sure to have valid projects and language engagements ready.  Create a Product on projects/:projectId/engagements/:engagementId/create-product.  Then on success click the edit link to open the edit page, and see the values filled


frankyma:
Added the products card with a delete action, its storybook, and a link out to create product from engagement.

The card height isn't fixed, and will have mismatches if the products have some fields that exist or not,  or text  that flow to a new line.  Would like Design to input on what we should do.  Also, the card design doesn't have purpose.
Screen Shot 2020-08-03 at 3 42 24 PM

The scripture references isn't ready on the API.  It's mocked on the storybook but also not parsed like the design.  It's still a flat list rather than combined.  Waiting on the scripture pills and validations, etc to be completed

frankyma:
Found a weird bug where if the page loads on a Language Engagement Detail, and click on create product to navigate to the create page, and before the breadcrumbs load click back on the browser, the gql network calls keeps firing one after the other indefinitely.  I've confirmed that neither the Langue Engagement Detail nor the Create Product component is rerendering, so not sure what's going on there.

glassblowerscat:
@frankyma I was able to successfully create a product on my second try, but on my first try I accidentally added the exact same "Scripture Range" twice. I decided to leave it to see what would happen, and sure enough, the mutation failed. We might want to consider doing some kind of de-duplication checking, but maybe it's an edge case.

frankyma:
> @frankyma I was able to successfully create a product on my second try, but on my first try I accidentally added the exact same "Scripture Range" twice. I decided to leave it to see what would happen, and sure enough, the mutation failed. We might want to consider doing some kind of de-duplication checking, but maybe it's an edge case.

@glassblowerscat The raw input scripture range I have is just a stub because we have the scripture pills coming soon, and that has all sorts of validation built in (or so I was promised by @CarsonF )

frankyma:
Pushed new changes.

On Edit, the product type is locked, but the producible can still be changed.

The correct scripture reference flow is still pending the scripture pills, and right now the scriptures on edit is locked.  Also don't have functionality to display the default scripture references of the producibles no override.  Pending discussion with @CarsonF and Design.

Many UI changes on accordion and product card to match colors, margins, etc

frankyma:
Looks like the CI is failing because there's a TS mismatch.  Not using the latest schema yet because api master is still broken atm

frankyma:
Just rebased and resolved conflicts.  @CarsonF Are we still pending a discussion with API on how to handle scriptureRefrences?   If not on the FE we just need the scripture fields, and dynamically change the edit page's scripture field name depending on the type (direct = scriptureReference, derivative = scriptureReferenceOverride)  so the error message comes through.

CarsonF (approved):
Ok I'm happy enough with this now for a first release.

Still lot's we can improve around scripture handling.
I'm also dev-ing an `EnumField` which can replace the radio fields used here.

Thanks for all your hard work & sticking with me on this @frankyma! 🎉

PR #350, opened 2020-07-24
Thursday save feature/fe 344 product create edit

frankyma:
Older version of accordion that expands and collapses accordion based on user clicking on the toggle buttons

Previous description
`CreateProduct.tsx` is the create product scene, which renders a ff `Form`, and has a local state (created from `useReducer`) to track the state of the accordion pieces.  `CreateProduct` renders `AccordionState.tsx` inside of the form to sync the accordion state with the form state.

`AccordionState` is a renderless function, and it keeps track of the current and previous (useRef) form state, and if the right state value changed (e.g. if the product field went from **untouched** to **touched**) then it'll show the next accordion by going through the reducer.

@CarsonF let me know how you feel about this.  I was having some trouble with the typescript on the reducer pieces (you'll see an abundance of `any` for now), I hope that's just from me being green with TS but let me know if you see this being an issue.

Also I'm using MUI Accordion now but the styling is different enough from our design that I just might create our own.  It's got box shadows, colors, and a pseudo element that i'll have to change.  But I'll cross this bridge later.


frankyma:
Closing PR after discussion with Carson and Patrick to have independent accordion sections

PR #356, opened 2020-07-28
[Bugfix][FE-349] Redirect after login

frankyma:
The login component was always redirecting to home if already logged in, so now reading the `returnTo`

**Testing**
open incognito, go to /users, after login /users should be open

### DIFF
--- a/src/scenes/Authentication/Login/Login.tsx
+++ b/src/scenes/Authentication/Login/Login.tsx
@@ -14,12 +14,13 @@ export const Login = (props: Except<Props, 'onSubmit'>) => {
   const [session, sessionLoading, setUserSession] = useSession();
   const [success, setSuccess] = useState(false);
 
-  // Redirect to homepage if already logged in (and not from successful login)
+  // Redirect to homepage or returnTo url if already logged in (and not from successful login)
   useEffect(() => {
     if (!sessionLoading && session && !success) {
-      navigate('/', { replace: true });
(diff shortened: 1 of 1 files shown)


frankyma:
I believe the issue is the `Login` component submit handler calls `setUserSession` (sets the logged in user in local apollo cache using `client.writeQuery`) and `navigate` (navigates back to home with the returnTo path).  Writing to apollo cache triggers a rerender in the parent `Authentication` component.  The first rerender has `session` being `null`, causing the app to navigate to `Login` again (from the useEffect hook).  Then the `Session` custom hook gets fired, and Authentication rerenders, this time with the correct `session` object, but since the `Login` component is already mounted its own effect hook triggers that navigates back to home, ignoring the `returnTo` path.  

I'm not sure why the first rerender has the empty session, as I've confirmed that `client.writeQuery` is synchronous (again, it's writing to local not remote).  I don't fully understand the `constate` package, which is what wraps the `Session` function into  context that can be used elsewhere, but I do see that each call to `useSession` doesn't immediately cause `Session` to re-run, and thus updating the returned values (`session`, `isLoading`, etc).

The code I pushed does fix the redirect even though it calls the `navigate` twice.  it makes Login catch landing on this component while being logged in and having a redirect.  There's no bad UX, so if we want a quick fix this'll do.  Otherwise I may need some help as I don't know how much more time I'll need to fix.

glassblowerscat:
This will be fixed by #490

glassblowerscat:
Issue in question was closed by a different PR; closing

PR #357, opened 2020-07-29
Cord field 270 verses field

blakekrammes:


src/components/form/VersesField.tsx
```diff
@@ -0,0 +1,259 @@
...
+        const validated = cleaned ? validateInput(cleaned) : '';
+        if (validated === null) return;
+        input.onChange(validated);
+        setInputValue(validated);
+      }}
+      onChange={(_, value) => {
+        if (meta.error) {
+          setErrorCode('createError');
+          // it'd be nice to be able to preserve the input value if a user tries submitting when an error is present
+          // setInputValue(value[value.length - 1] as string);
```
blakekrammes, line 232: The bug I mentioned is that if there's an error and the user hits enter, the input is erased. I was manually setting the input value before, but this caused other problems with FF's state. This code is planned to be eventually refactored anyhow.

CarsonF (approved):
This is a complicated component. Lets merge since it is functional now and the interface is unlikely to change. We'll continue to iterate on it.

PR #360, opened 2020-07-31
[Bugfix][FE 318] pronounciation guide

frankyma:
Added word guide


CarsonF:
How about the LanguageForm too?

frankyma:
> How about the LanguageForm too?

You're right.  Changed...not sure about the placeholder though? `placeholder="Enter the Pronunciation Guide of Public Name"`

CarsonF:
Hmm yeah that looks weird. Maybe keep the placeholder how it was.

frankyma:
> Hmm yeah that looks weird. Maybe keep the placeholder how it was.

changed

PR #383, opened 2020-08-12
Ignore generated types

CarsonF:
This changes two parts of the gql codegen process.

## 1.) Ignoring generated files
Originally the `*.generated.ts` files were committed so that CI could validate the types.
Now that we have a hosted API, we can pull them from there.

This does mean that local breaking changes on both the API & UI will not pass in CI now, but that was rarely, if ever the case. The API should strive to not have breaking changes anyways.

### Benefits
- Reduced clutter in commits
- Reduced merge conflicts while rebasing
- The gql-codegen process is much faster now, because we don't lint or prettify the generated files now

The only down side of this is not being able to see schema changes between commits/branches. But this is minor and available in other places.

## 2.) Schema Location

Previously we required devs to symlink the `schema.grahql` file from their local API folder.
Because this:
- Allowed the frontend to use a schema modified in a working branch on the API
- Allowed IDEs to use this schema file to autocomplete/validate our graphql documents

However, symlinks are gross and this prevented using a remote schema even though you could use a remote API.

### Now

- We pull the schema from the API
- Which API is configured by `.env` files which gql-codegen now loads
- gql-codegen generates a `schema.graphql` file for our IDEs/us to use

## Caveats 

- The server process has to be running to generated types. Not a big deal.

- Schema is baked at build time which means based on API being referenced the built package could change. 
  - We do have a few generated things that are used at runtime (the fragment matcher and the type policies).
  - If anything this is more accurate as the package would be closer to what the API is serving.
  - If the API has breaking changes it could fail the build though.


glassblowerscat:
LGTM!

.github/workflows/ci.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - v3
 
+env:
+  REACT_APP_API_BASE_URL: ${{ secrets.API_BASE_URL }}
```
blakekrammes, line 9: This is the url of the hosted API and the same url as the schema url/`graphql` below, right?

CarsonF: Correct. And I just wanted to be able to change it without a commit. We don't reference the deployed envs anywhere outside of our infrastructure code.

PR #385, opened 2020-08-12
[FE-371] Show Ethnologue code on language engagement card

frankyma:
Shown same as Registry of Dialects Code if exists.  Ignoring the `canRead` property same as the other fields for now.

Screen Shot 2020-08-12 at 3 39 56 PM

### DIFF
--- a/src/components/DisplaySimpleProperty/DisplaySimpleProperty.tsx
+++ b/src/components/DisplaySimpleProperty/DisplaySimpleProperty.tsx
@@ -22,6 +22,8 @@ export const DisplaySimpleProperty = ({
   wrap,
   ...props
 }: DisplaySimplePropertyProps) => {
+  const shouldRenderElement = loading || (label && value);
+  if (!shouldRenderElement) return null;
   const property = (
     <Typography variant="body2" {...props}>
       {loading === true ? (
--- a/src/components/LanguageEngagementListItemCard/LanguageEngagementListItemCard.tsx
+++ b/src/components/LanguageEngagementListItemCard/LanguageEngagementListItemCard.tsx
@@ -81,6 +81,7 @@ export const LanguageEngagementListItemCard: FC<LanguageEngagementListItemCardPr
   const name = language?.name.value ?? language?.displayName?.value;
   const population = language?.population.value;
   const registryOfDialectsCode = language?.registryOfDialectsCode.value;
+  const ethnologueCode = language?.ethnologue.code.value;
   const endDate = getEndDate(props);
 
   return (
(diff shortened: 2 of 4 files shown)


CarsonF:
Should the code labels be the same color as the status label?

frankyma:
> Should the code labels be the same color as the status label?

We could, I saw you just render a typography for the Registry code so I did the same with Ethnologue code.  The Status uses `DisplaySimpleProperty`

Also on 2nd look, there's no reason for the new line on the Registry code with so much space on the right, i'll take a look

frankyma:
Changed to use display simple property, also don't render the right side content if `population` and `endDate` doesn't exist

Screen Shot 2020-08-12 at 5 06 39 PM

src/components/LanguageEngagementListItemCard/LanguageEngagementListItemCard.tsx
```diff
@@ -103,13 +104,18 @@ export const LanguageEngagementListItemCard: FC<LanguageEngagementListItemCardPr
...
-                  Registry Of Dialects Code: {registryOfDialectsCode}
-                </Typography>
-              </Grid>
-            ) : null}
+            <Grid item>
+              <DisplaySimpleProperty
+                label="Registry Of Dialects Code"
+                value={registryOfDialectsCode}
+              />
+            </Grid>
```
CarsonF, line 112: Can you change these to use the new `wrap` prop? This way the grid item doesn't get rendered if the value doesn't exist.
```suggestion
            <DisplaySimpleProperty
              label="Registry Of Dialects Code"
              value={registryOfDialectsCode}
              wrap={(node) => <Grid item>{node}</Grid>}
            />
```

frankyma: Done, had to change up the `DisplaySimpleProperty` wrap rending though because it was still calling the wrap if the value is empty

src/components/DisplaySimpleProperty/DisplaySimpleProperty.tsx
```diff
@@ -22,6 +22,7 @@ export const DisplaySimpleProperty = ({
   wrap,
   ...props
 }: DisplaySimplePropertyProps) => {
+  const shouldRenderElement = loading || (label && value);
```
CarsonF, line 25: Ah my bad I forgot that logic. Can you return early here with this logic?

frankyma: good call, changed

PR #388, opened 2020-08-13
Bugfix: LanguageEngagementListItemCard show population or end date

frankyma:
LanguageEngagementListItemCard should use || operator if either `population` or `endDate` available


CarsonF:
Thanks @frankyma. I fixed it with #387

PR #393, opened 2020-08-17
Global search

jmc637:


jmc637:
Hi, this is a PR for global search. I had to store the search state somewhere so I put it in the router state. I don't have a strong preference for where this is stored. Initially I tried to put it as a query param but then had to deal with syncing this with the search input so I switched it to router state since this was easier to do. If you know a different way to do this that is cleaner I can change this.

src/scenes/Root/Header/HeaderSearch/HeaderSearch.tsx
```diff
@@ -15,13 +16,30 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
...
+    navigate(
+      {
+        pathname: 'search',
+      },
+      {
+        state: {
+          searchValue: value,
+        },
+      }
+    );
```
CarsonF, line 33: > I had to store the search input value somewhere so I put it in the router state. I don't have a strong preference for where this is stored. Initially I tried to put it as a query param but then had to deal with syncing this with the search input so I switched it to router state. If you know a different way to do this that is cleaner I can change this.


Yeah you can use the `makeQueryHandler` function which produces a "setState"-like hook

```tsx
// outside of component
const useSearch = makeQueryHandler({
  search: StringParam,
});

// in component
const [val, setVal] = useSearch();
```
This hook can be used in multiple places too.

src/scenes/SearchResults/search.graphql
```diff
@@ -0,0 +1,42 @@
+query Search($input: SearchInput!) {
+  search(input: $input) {
+    items {
+      ... on Country {
+        ...DisplayCountry
+      }
```
CarsonF, line 6: We actually just figured out we don't need to specify the type here to put the fragment inside. Because the fragment defines what it is on.
```suggestion
      ...DisplayCountry
```

You could think of it like a named vs anonymous function.
```gql
... on X {}
```
```gql
myX on X {}
```

src/scenes/Root/Header/HeaderSearch/HeaderSearch.tsx
CarsonF: Here's an example in code
https://github.com/SeedCompany/cord-field/blob/a87f0ece1aef8bc5a070ca6dfa41d440f3a5260d/src/scenes/Projects/List/ProjectFilterOptions.tsx#L8-L12

jmc637: Hi, I've had difficulty with the useSearchParams functions from "react-router-dom" inside the makeQueryHandler hook. It keeps resetting the url to "/" instead of appending the url parameters to the current url. Is there a parameter I need to add to navigate options to not reset the url? I've tried to set replace to false but I think this is for url params and not the url.

src/scenes/SearchResults/search.graphql
jmc637: I'll change this to the new way have just been trying to deal with issue with url being redirected from hook.

jmc637:
Hi, I pushed the code mentioned earlier with the redirect issue

src/scenes/Root/Header/HeaderSearch/HeaderSearch.tsx
CarsonF: I was able to reproduce that @jmc637, sorry for the delay. This is a problem with react-router. I applied a workaround in our makeQueryHandler code.

PR #394, opened 2020-08-20
Create producibles

blakekrammes:


src/components/form/Lookup/Film/FilmField.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { CreateFilm } from '../../../../scenes/Engagement/LanguageEngagement/Product/Producibles/Film/CreateFilm';
+import { useFilmLookupLazyQuery } from './FilmLookup.generated';
+
+export const FilmField = LookupField.createFor<FilmLookupItem, CreateFilmInput>(
+  {
+    resource: 'Film',
+    useLookup: useFilmLookupLazyQuery,
+    // option is nullish initially
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    getOptionLabel: (option) => option?.name?.value ?? '',
```
CarsonF, line 13: Why is it called with a null? Sounds like a problem with our recent changes.

blakekrammes: Ah. Because we call `getOptionLabel` with `null` if multiple is false

```typescript
const selectedText = multiple
    ? ''
    : (getOptionLabel || identity)(field.value as T) ?? '';
```

blakekrammes: What is the purpose of identity here?

blakekrammes: Why not:

```typescript
const selectedText =
    !multiple && getOptionLabel && (field.value as T)
      ? getOptionLabel(field.value as T)
      : '';
```

src/components/form/Lookup/LookupField.tsx
```diff
@@ -64,20 +71,31 @@ export function LookupField<
...
+  freeSolo,
   defaultValue,
   disabled: disabledProp,
   useLookup,
   ChipProps,
   helperText,
   label,
   required,
+  CreateDialogForm,
+  getOptionLabel,
```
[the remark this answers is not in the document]
blakekrammes: Do you prefer doing` props.getOptionLabe`l to passing it directly? And what do you mean recheck that all changes are necessary?

src/components/form/Lookup/Film/FilmField.tsx
blakekrammes: Don't you mean this?

```typescript
const selectedText = multiple || !field.value
    ? ''
    : (getOptionLabel || identity)(field.value as T) ?? '';
```

Otherwise the problem with `option` being null in `getOptionLabel` remains. 
`getOptionLabel: (option) => option.name.value ?? ''`

And if the above is correct, `field.value` will never being nullish, but either a string literal or empty string, so the coalescing at the end is unnecessary.

src/components/form/Lookup/LookupField.tsx
```diff
@@ -135,63 +157,126 @@ export function LookupField<
...
-      // want to hide it in the dropdown if we start deleting characters
-      filterSelectedOptions={multiple || input === selectedText}
-      renderTags={(values: T[], getTagProps) =>
-        values.map((option: T, index: number) => (
-          <Chip
-            variant="outlined"
-            {...ChipProps}
-            label={props.getOptionLabel!(option)}
-            {...getTagProps({ index })}
+    <>
```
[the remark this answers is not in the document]
blakekrammes: Children don't exist on Autocomplete, so I'll do a var.

src/components/form/Lookup/Film/FilmField.tsx
blakekrammes: Although, we still need to coalesce if `name.value` doesn't exist since it's a secured string, right?

src/components/form/Lookup/LookupField.tsx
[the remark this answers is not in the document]
CarsonF: That's fine. I think calling it via `props.` helps to represent that this isn't our prop, but a prop of the underlying component we are just also using. It's also less code because you don't have to pass it down explicitly.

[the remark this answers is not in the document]
blakekrammes: True

src/components/form/Lookup/Film/FilmField.tsx
CarsonF: You're right, my bad.

blakekrammes: No problemo

src/components/form/Lookup/LookupField.tsx
```diff
@@ -135,63 +157,126 @@ export function LookupField<
...
+      <Autocomplete<T, Multiple, DisableClearable, FreeSolo>
+        disabled={disabled}
+        // FF also has multiple and defaultValue
+        multiple={multiple}
+        // if we only have one value that can and has been selected, we don't
+        // want to hide it in the dropdown if we start deleting characters
+        filterSelectedOptions={multiple || input === selectedText}
+        renderTags={(values: T[], getTagProps) =>
+          values.map((option: T, index: number) => {
+            return (
```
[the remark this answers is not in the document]
blakekrammes: What do you mean?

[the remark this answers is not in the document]
CarsonF: You added brackets to the map function that aren't needed.

CarsonF:
@blakekrammes I smoothed out most/all? the edge cases. What do you think?

blakekrammes:
Bueno!

PR #399, opened 2020-08-24
Fix S3 downloads

CarsonF:
Ported file-saver to our src. Dropped is support for IE. Added option to skip HEAD requests to check CORS.
This is not compatible with S3's signed urls, since they are signed specifically for GET requests.


CarsonF:
Bleh this doesn't apply. webpack uses minified file. I'm going to copy src into our repo (only a couple methods). I could clean it up alot (no globals) and apply changes from a bunch of their open PRs.
👍 /👎  @glassblowerscat?

glassblowerscat:
> 👍 /👎 @glassblowerscat?

Sounds good to me!

CarsonF:
@glassblowerscat How's this look to you? There's more open PRs from their library that could potentially be useful for us to adopt. Maybe after this PR.

PR #401, opened 2020-08-24
Feature/update project status

frankyma:
Update Project Step dialog.  It's just a long scroll down for now.  The new schema is from SeedCompany/cord-api-v3#1038, hence the failing CI but the code is ready for review

### DIFF
--- a/src/api/projectSteps.ts
+++ b/src/api/projectSteps.ts
@@ -1,0 +1,24 @@
+import { ProjectStatus, ProjectStep } from './schema.generated';
+
+export const projectStepToStatusMap: Record<ProjectStatus, ProjectStep[]> = {
+  InDevelopment: [
+    'EarlyConversations',
+    'PrepForConsultantEndorsement',
+    'PrepForFinancialEndorsement',
+    'FinalizingProposal',
+    'PrepForGrowthPlanEndorsement',
+    'PendingGrowthPlanEndorsement',
+  ],
+  Pending: [
+    'PendingConceptApproval',
+    'PendingConsultantEndorsement',
+    'PendingFinancialEndorsement',
+    'PendingAreaDirectorApproval',
+    'PendingRegionalDirectorApproval',
+    'PendingFinanceConfirmation',
+    'OnHoldFinanceConfirmation',
+  ],
+  Active: ['Active'],
+  Stopped: ['Suspended', 'Rejected', 'Terminated'],
+  Finished: ['DidNotDevelop', 'Completed'],
+};
--- a/src/api/projectSteps.ts
+++ b/src/api/projectSteps.ts
@@ -1,6 +1,7 @@
+import { entries, mapFromList } from '../util';
 import { ProjectStatus, ProjectStep } from './schema.generated';
 
-export const projectStepToStatusMap: Record<ProjectStatus, ProjectStep[]> = {
+export const projectStatusToStepsMap: Record<ProjectStatus, ProjectStep[]> = {
   InDevelopment: [
     'EarlyConversations',
     'PrepForConsultantEndorsement',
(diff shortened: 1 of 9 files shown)


src/api/projectSteps.ts
```diff
@@ -0,0 +1,62 @@
+enum ProjectStatus {
+  InDevelopment = 'InDevelopment',
+  Pending = 'Pending',
+  Active = 'Active',
+  Stopped = 'Stopped',
+  Finished = 'Finished',
+}
+
+enum ProjectStep {
```
CarsonF, line 9: Both of these enums can be replaced with the string literals from generated schema.

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -293,6 +311,66 @@ export const ProjectOverview: FC = () => {
           )}
         </div>
       )}
+      <DialogForm<UpdateProjectInput>
```
CarsonF, line 314: Separate component/file please

src/api/projectSteps.ts
frankyma: Changed

src/scenes/Projects/Overview/ProjectOverview.tsx
frankyma: Moved

src/scenes/Projects/Update/UpdateProjectDialog.tsx
```diff
@@ -0,0 +1,89 @@
+import { Typography } from '@material-ui/core';
```
CarsonF, line 1: Can you rename the file so it matches the component name?

src/scenes/Projects/Update/UpdateProjectDialog.tsx
```diff
@@ -0,0 +1,89 @@
...
+        },
+      }}
+      onSubmit={async (input) => {
+        await updateProject({ variables: { input } });
+      }}
+    >
+      <SubmitError />
+      <RadioField name="project.step">
+        <Typography variant="h4">In Development</Typography>
+        {projectStepToStatusMap.InDevelopment.map((step) => (
```
CarsonF, line 47: You could loop through the groups as well

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -293,6 +301,10 @@ export const ProjectOverview: FC = () => {
           )}
         </div>
       )}
+      <UpdateProjectStepDialog
```
CarsonF, line 304: I think it makes more sense to put this next to the DataButton. Dialogs use a Portal so it doesn't matter where they go in JSX.

src/scenes/Projects/Update/UpdateProjectDialog.tsx
frankyma: Done

src/scenes/Projects/Update/UpdateProjectDialog.tsx
frankyma: Fixed, much cleaner 👌

src/scenes/Projects/Overview/ProjectOverview.tsx
frankyma: I was thinking of using this dialog for the update dates and location too

CarsonF: It's called UpdateStep though?

frankyma: I renamed it `UpdateProjectDialog`

CarsonF:
@frankyma The dialog was really long due to all the step options. I changed it to an autocomplete. What do you think?

frankyma:
> @frankyma The dialog was really long due to all the step options. I changed it to an autocomplete. What do you think?

Seems like a good idea.  I was thinking we'd divide into columns but autocomplete is easier

PR #408, opened 2020-08-26
Create Project Team Members

frankyma:
Create Project Team Members

ezgif com-video-to-gif (1)

There's still a bug where if there's already an existing team member, the api call errors, but the error doesn't show on the dialog.  I think it's because on submit the dialog resets the userField to initial values so the only error shown is the `required` error.
create-member-bug

### DIFF
--- a/src/scenes/Projects/Members/Create/CreateProjectMember.tsx
+++ b/src/scenes/Projects/Members/Create/CreateProjectMember.tsx
@@ -1,3 +1,5 @@
+import { Decorator } from 'final-form';
+import onFieldChange from 'final-form-calculate';
 import React, { useMemo } from 'react';
 import { Except, Merge } from 'type-fest';
 import {
@@ -33,6 +35,15 @@ type CreateProjectMemberProps = Except<
   projectId: string;
 };
 
+const decorators: Array<Decorator<FormValues>> = [
+  ...DialogForm.defaultDecorators,
+  onFieldChange({
+    field: 'projectMember.userId',
+    isEqual: UserField.isEqual,
+    updates: { 'projectMember.roles': () => [] },
+  }),
+];
+
 export const CreateProjectMember = ({
   projectId,
   ...props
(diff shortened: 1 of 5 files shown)


src/scenes/Projects/Members/Create/CreateProjectMember.tsx
```diff
@@ -0,0 +1,138 @@
...
+}
+
+type CreateProjectMemberProps = Except<
+  DialogFormProps<DialogValues>,
+  'onSubmit' | 'initialValues'
+> & {
+  projectId: string;
+};
+
+const Roles: Role[] = [
```
frankyma, line 32: We don't have a standard of where to throw these constants.  We can either keep it in the component for 1 time uses or have a constants directory in our api/utils/etc

CarsonF: If we had a `RoleField` it would make since there ;)

I'm thinking about auto generating these lists from the api schema, so we can use them at run time. But that's for another time. It's fine here or in api/ somewhere

CarsonF:
Is that gif incomplete? It looks like that dialog is prematurely closing. If there's a duplicate error from the API on submit then the form should handle that correctly. It shouldn't be resetting to initial values, but it could be...

src/scenes/Projects/Members/Create/CreateProjectMember.tsx
frankyma: Sounds good, i'll look into the `RoleField` as I do the edit

frankyma:
> Is that gif incomplete? It looks like that dialog is prematurely closing. If there's a duplicate error from the API on submit then the form should handle that correctly. It shouldn't be resetting to initial values, but it could be...

Yes sorry the gif got cut off, uploaded the complete one.  I also uploaded the duplicate error gif

CarsonF (approved):
@frankyma I tweaked somethings. It seems to be working well for me (other than auth on roles from API). Will you give my changes a once over and maybe pull it down and test?

frankyma:
Happy path is working.  Saw an issue where if you select person A and role X, then switch to person B.  role X remains selected and can be submitted even if that role isn't available to the user.  The API doesn't give any errors neither

CarsonF:
Ah good call! Yes let's clear it on user change. Can you open an API issue to validate user roles on project member create/update?

frankyma:
Just tested, LGTM

CarsonF (approved):

PR #410, opened 2020-08-26
Edit and Delete team members

frankyma:
Edit and delete team members
ezgif com-video-to-gif (1)

### DIFF
--- a/src/scenes/Projects/Members/List/ProjectMembersList.tsx
+++ b/src/scenes/Projects/Members/List/ProjectMembersList.tsx
@@ -101,7 +101,6 @@ export const ProjectMembersList: FC = () => {
                 projectMemberId: item.id,
                 userId: item.user.value?.id || '',
                 userRoles: item.roles.value,
-                projectId,
               })
             }
           />
--- a/src/scenes/Projects/Members/List/ProjectMembersList.tsx
+++ b/src/scenes/Projects/Members/List/ProjectMembersList.tsx
@@ -61,6 +61,7 @@ export const ProjectMembersList: FC = () => {
     projectMemberId?: string;
     userId?: string;
     userRoles?: readonly Role[];
+    projectId: string;
   }>();
 
   return (
@@ -105,6 +106,7 @@ export const ProjectMembersList: FC = () => {
                 projectMemberId: item?.id,
                 userId: item?.user.value?.id,
                 userRoles: item?.roles.value,
+                projectId,
               })
             }
           />
@@ -115,6 +117,7 @@ export const ProjectMembersList: FC = () => {
         id={projectMemberProps?.projectMemberId ?? ''}
         userId={projectMemberProps?.userId ?? ''}
         userRoles={projectMemberProps?.userRoles}
+        projectId={projectId}
       />
     </div>
   );
--- a/src/scenes/Projects/Members/List/ProjectMembersList.tsx
+++ b/src/scenes/Projects/Members/List/ProjectMembersList.tsx
@@ -8,6 +8,7 @@ import { Add } from '@material-ui/icons';
 import { FC } from 'react';
 import * as React from 'react';
 import { useParams } from 'react-router-dom';
+import { Role } from '../../../../api';
 import { Breadcrumb } from '../../../../components/Breadcrumb';
 import { useDialog } from '../../../../components/Dialog';
 import { Fab } from '../../../../components/Fab';
@@ -15,6 +16,7 @@ import { ProjectBreadcrumb } from '../../../../components/ProjectBreadcrumb';
 import { ProjectMemberCard } from '../../../../components/ProjectMemberCard';
 import { listOrPlaceholders } from '../../../../util';
 import { CreateProjectMember } from '../Create/CreateProjectMember';
+import { UpdateProjectMember } from '../Update';
 import { useProjectMembersQuery } from './ProjectMembers.generated';
 
 const useStyles = makeStyles(({ spacing, breakpoints }) => ({
@@ -50,6 +52,17 @@ export const ProjectMembersList: FC = () => {
     createProjectMemberDialogState,
     openCreateProjectMemberDialog,
   ] = useDialog();
+
+  const [
+    updateProjectMemberDialogState,
+    openUpdateProjectMemberDialog,
+    projectMemberProps,
+  ] = useDialog<{
+    projectMemberId?: string;
+    userId?: string;
+    userRoles?: readonly Role[];
+  }>();
+
   return (
     <div className={classes.root}>
       <Breadcrumbs>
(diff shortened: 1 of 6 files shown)


CarsonF (commented):
On partnerships, we have the delete button in the edit dialog. We should probably be consistent one way or the other. Since deleting isn't a very common use case, I'm leaning towards in the dialog. What do you think?

frankyma:
Close call, on one hand it's an extra step to go from update -> delete, on the other hand if it's a direct delete button we might want a confirmation dialog.  I can lean towards update -> delete if it's not a common action

CarsonF:
Let's go with that then. Might help us avoid a confirmation dialog lol

frankyma:
Edit functionality moved to the update dialog

src/scenes/Projects/Members/Update/UpdateProjectMember.tsx
```diff
@@ -0,0 +1,117 @@
...
+    awaitRefetchQueries: true,
+  });
+
+  const availableRoles = data?.user.roles.value ?? [];
+  return (
+    <DialogForm<DialogValues>
+      title="Update Team Member Role"
+      closeLabel="Close"
+      submitLabel="Save"
+      onlyDirtySubmit
```
glassblowerscat, line 67:
```suggestion
```
This is `true` by default in `DialogForm`

src/scenes/Projects/Members/List/ProjectMembersList.tsx
```diff
@@ -87,9 +101,24 @@ export const ProjectMembersList: FC = () => {
             key={item?.id ?? index}
             projectMember={item}
             className={classes.item}
+            onEdit={() =>
+              openUpdateProjectMemberDialog({
+                projectMemberId: item?.id,
+                userId: item?.user.value?.id,
+                userRoles: item?.roles.value,
```
glassblowerscat, line 108: This (combined with lines 117-119) seems like it's putting the burden on `UpdateProjectMember` to handle empty values, when really those values would just make it non-functional. It would be better to make all these parameters required in the type variable we're passing to `useDialog` above, so that it matches the typings for `UpdateProjectMember`. Then we should do the proper checking inside this `.map()` to avoid rendering anything at all if `item` is `undefined`. 

We'd still have the problem of needing the `userId`, though. Could `ProjectMemberCard` be a wrapper around `FieldOverviewCard`? It was recently updated to handle `canRead` and `canEdit`, and (among other things) avoid displaying an "Edit" button if `canEdit` is `false`. You can see one example of how its used in the `FilesOverviewCard`.

src/scenes/Projects/Members/Update/UpdateProjectMember.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+type UpdateProjectMemberProps = Except<
+  DialogFormProps<DialogValues>,
+  'onSubmit' | 'initialValues'
+> & {
+  id: string;
+  userId: string;
+  userRoles?: readonly Role[];
+  projectId: string;
+};
```
glassblowerscat, line 34:
```suggestion

import { UpdateProjectMemberInput } from '../../../../api';

export interface UpdateProjectMemberFormParams {
  projectMember: UpdateProjectMemberInput['projectMember'];
  userId: string;
}

type UpdateProjectMemberProps = Except<
  DialogFormProps<UpdateProjectMemberInput>,
  'onSubmit' | 'initialValues'
> &
  UpdateProjectMemberFormParams;
```

It's safer to base these values on the actual input type the mutation will be expecting (`UpdateProjectMemberInput`). This way if it ever changes we don't have to rewrite these typings. 

Also, you can then import `UpdateProjectMemberFormParams` in `ProjectMembersList` and pass it as the type variable to `useDialog`.

src/scenes/Projects/Members/List/ProjectMembersList.tsx
```diff
@@ -50,6 +52,18 @@ export const ProjectMembersList: FC = () => {
...
+  const [
+    updateProjectMemberDialogState,
+    openUpdateProjectMemberDialog,
+    projectMemberProps,
+  ] = useDialog<{
+    projectMemberId?: string;
+    userId?: string;
+    userRoles?: readonly Role[];
+    projectId: string;
+  }>();
```
glassblowerscat, line 65: This fields should just all be required, to match `UpdateProjectMemberProps`. Then we don't have to mess with passing empty strings if they're undefined.

src/scenes/Projects/Members/List/ProjectMembersList.tsx
```diff
@@ -87,9 +101,24 @@ export const ProjectMembersList: FC = () => {
...
           />
         ))
       )}
+      <UpdateProjectMember
+        {...updateProjectMemberDialogState}
+        id={projectMemberProps?.projectMemberId ?? ''}
+        userId={projectMemberProps?.userId ?? ''}
+        userRoles={projectMemberProps?.userRoles}
+        projectId={projectId}
+      />
```
glassblowerscat, line 121: Here we would just wrap everything in 

```
{projectMemberProps &&
  ...
}
```

glassblowerscat (changes requested):

src/scenes/Projects/Members/Update/UpdateProjectMember.tsx
```diff
@@ -0,0 +1,117 @@
...
+    },
+  });
+
+  const [deleteProjectMember] = useDeleteProjectMemberMutation({
+    refetchQueries: [
+      {
+        query: ProjectMembersDocument,
+        variables: { input: projectId },
+      },
+    ],
```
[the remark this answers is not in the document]
frankyma: good to know, changed

src/scenes/Projects/Members/List/ProjectMembersList.tsx
frankyma: > Then we should do the proper checking inside this .map() to avoid rendering anything at all if item is undefined.
We still need to render to show a loading list, but I did make the props required

> Could ProjectMemberCard be a wrapper around FieldOverviewCard?
I think this is a bigger ask that I'd like to do outside of this PR if we choose to go this route.

src/scenes/Projects/Members/Update/UpdateProjectMember.tsx
frankyma: got it, removed

src/scenes/Projects/Members/Update/UpdateProjectMember.tsx
frankyma: Good idea, done

src/scenes/Projects/Members/List/ProjectMembersList.tsx
frankyma: Done

src/scenes/Projects/Members/List/ProjectMembersList.tsx
glassblowerscat: > We still need to render to show a loading list

Lol, wasn't thinking

src/scenes/Projects/Members/List/ProjectMembersList.tsx
frankyma: set to use `UpdateProjectMemberProps` as suggested good idea

glassblowerscat (approved):

PR #413, opened 2020-08-30
Spreadsheet preview merged cells

glassblowerscat:
SheetJS provides a list of merged cells in the parsed spreadsheet, but it will only integrate this data with the actual cell contents if using the `sheet_to_html` utility, which we'd rather not use for other reasons. This PR does two things, primarily:

1. Update the `SpreadsheetView` component to take the individual cell data not as a string or number but as this:

```
{
  index: number;
  spanned: false;
  rowspan: number;
  colspan: number;
  content: string | number;
}
```

2. Add a function to `ExcelPreview` to read through the list of cell merges then feed the row data provided by the `sheet_to_json` util through it to output the above type of data

We also update `CsvPreview` to provide row data in this format and do some general tidying up of typings between the three components, centralizing anything consumed by both `CsvPreview` and `ExcelPreview` in `SpreadsheetView`.

### DIFF
--- a/src/components/files/FilePreview/SpreadsheetView.tsx
+++ b/src/components/files/FilePreview/SpreadsheetView.tsx
@@ -54,18 +54,25 @@ export type ColumnData = Array<{
 }>;
 export type RowData = ReturnType<XLSX$Utils['sheet_to_json']>;
 
-export interface TableSpan {
-  startColumn: number;
-  startRow: number;
-  colspan: number;
-  rowspan: number;
+export interface TableCellSpanned {
+  index: number;
+  spanned: true;
 }
+
+export interface TableCellData {
+  index: number;
+  spanned: false;
+  rowspan: number;
+  colspan: number;
+  content: string | number;
+}
+
+export type TableRow = Array<TableCellSpanned | TableCellData>;
 
 export interface SheetData {
   name: string;
-  rows: RowData;
+  rows: TableRow[];
   columns: ColumnData;
-  spans: TableSpan[];
 }
 
 interface SpreadSheetViewProps {
@@ -73,10 +80,7 @@ interface SpreadSheetViewProps {
 }
 
 const RenderedSheet: FC<Omit<SheetData, 'name'>> = (props) => {
-  const { rows, columns, spans } = props;
-  console.log('rows', rows);
-  console.log('columns', columns);
-  console.log('spans', spans);
+  const { rows, columns } = props;
   return (
     <table>
       <tbody>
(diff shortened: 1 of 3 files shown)


src/components/files/FilePreview/SpreadsheetView.tsx
```diff
@@ -76,16 +103,22 @@ const RenderedSheet: FC<Omit<SheetData, 'name'>> = (props) => {
...
-              <td key={column.key}>{row[column.key]}</td>
-            ))}
-          </tr>
-        ))}
+        {rows.map((cells, rowIndex) => {
+          return (
+            <tr key={rowIndex}>
+              <th className="table-header">{rowIndex + 1}</th>
+              {cells.map((cell) => {
+                if (cell.spanned) return null;
```
CarsonF, line 111: Trying to follow logic...are you padding empty cells in parsing above but then just ignoring them here?
Or am I missing another reason for that?

glassblowerscat: Correct. It's HTML, so the way you render a spanned `<td>` is to not render it

glassblowerscat: Okay, now that my brain has revved up a little I realize that you were probably asking why we're bothering to keep the `spanned: true` cells around at all. It's because the same item in the array of arrays that makes up the row data returned by `sheet_to_json` might need to be marked as `spanned` by more than one cell-merge. If we just removed it from the list when processing "Merge A", but then looked for its index again later when processing "Merge B", we'd find a totally different item and mark that one as `spanned` incorrectly.

CarsonF: Ok got it. And it's important to look it up by index? Or more specifically is it necessary to map to an object? If we were processing the actual data here I would say yes. But since we are just displaying via two loops we don't care about those keys/indexes. Or...?

glassblowerscat: `sheet_to_json` gives us something like this:

```
rows: [
  ["A1", B2", C2-D2", ""],
  ["A2", "B2-B3", "C2, D2"],
  ["A3", "", "C3-D4", ""],
  ["A4", "B4", "", ""],
]
```

If all we were dealing with were merged columns, we wouldn't need to use objects for tracking the state of our processing and the resulting `rowspan`/`colspan` values. We could just pass `SpreadsheetView` the raw `sheet_to_json` data, and then in our `.map()` just always look ahead of each current index to see if the next one or more items are blank, and if so, set `colspan` to the number of blank items + 1. But trying to do that look-ahead operation for subsequent rows, particularly when more than one column might be merged along with the rows, seems like it would get very messy. It seemed like a better idea to just process all the merging data up front and then just pass the "view" component something it could render pretty simply.

glassblowerscat: (Also, when I referred to looking up a cell by "index" above, I meant not the `index` value in the object we're creating here but the index its array. What I'm saying is we can't just delete empty strings from the row-data arrays and have things turn out the same.)

Also, an empty string in the row data isn't the same as a merge. It might literally be an empty cell. We need a way to distinguish, which is why we're using the SheetJS `worksheet['!merges']` value to get the list of merged cells and then tag the values in the row data accordingly.

CarsonF: I gotcha. In other words all of this info is stored in separate lists instead of a tree structure. Then yes, I'd agree that that processing/merging should be done separately like you've done.

PR #415, opened 2020-08-31
Secured directory query update

glassblowerscat:
**NOTE:** Will fail GQL type generation until API #1070 ([https://github.com/SeedCompany/cord-api-v3/pull/1070](https://github.com/SeedCompany/cord-api-v3/pull/1070)) is merged

The API is changing the type of value returned for `rootDirectory` in the `project` query from `Directory`, to `SecuredDirectory`, which follows the pattern of other "Secured" types. 

- Update `ProjectRootDirectory` query to retrieve correct fields based on API change
- Update `FilesOverviewCard`, `ProjectOverview`, `ProjectFilesList`, and `CreateProjectDirectory` components to handle the new `canRead` value, displaying different UI and potentially displaying errors if `canRead` is `false`
- Updates `FieldOverviewCard` to accept `loading`, `redacted`, and `redactedText` props, in order to handle loading and redacted states. Because this component renders several different pieces of data, we aren't using individual `Redacted` components for each of those pieces of data. Instead, if `redacted` is `true`, we're rendering static `Skeleton` components or no data, and wrapping the entire card in a `Tooltip` with the `redactedText`

### DIFF
--- a/src/scenes/Projects/Files/CreateProjectDirectory.tsx
+++ b/src/scenes/Projects/Files/CreateProjectDirectory.tsx
@@ -17,7 +17,7 @@ export const CreateProjectDirectory = (
   props: Except<CreateProjectDirectoryProps, 'onSubmit'>
 ) => {
   const [createDirectory] = useCreateProjectDirectoryMutation();
-  const { project, directoryId, loading } = useProjectCurrentDirectory();
+  const { project, directoryId } = useProjectCurrentDirectory();
   const { enqueueSnackbar } = useSnackbar();
 
   const onSubmit: CreateProjectDirectoryProps['onSubmit'] = async (
--- a/src/scenes/Projects/Files/CreateProjectDirectory.tsx
+++ b/src/scenes/Projects/Files/CreateProjectDirectory.tsx
@@ -17,24 +17,8 @@ export const CreateProjectDirectory = (
   props: Except<CreateProjectDirectoryProps, 'onSubmit'>
 ) => {
   const [createDirectory] = useCreateProjectDirectoryMutation();
-  const {
-    project,
-    directoryId,
-    loading,
-    canRead,
-  } = useProjectCurrentDirectory();
+  const { project, directoryId, loading } = useProjectCurrentDirectory();
   const { enqueueSnackbar } = useSnackbar();
-
-  if (canRead === false) {
-    enqueueSnackbar(
-      `You don't have permission to add folders in this project`,
-      {
-        preventDuplicate: true,
-        variant: 'error',
-        autoHideDuration: 3000,
-      }
-    );
-  }
 
   const onSubmit: CreateProjectDirectoryProps['onSubmit'] = async (
     nameInput
--- a/src/scenes/Projects/Files/CreateProjectDirectory.tsx
+++ b/src/scenes/Projects/Files/CreateProjectDirectory.tsx
@@ -17,8 +17,24 @@ export const CreateProjectDirectory = (
   props: Except<CreateProjectDirectoryProps, 'onSubmit'>
 ) => {
   const [createDirectory] = useCreateProjectDirectoryMutation();
-  const { project, directoryId } = useProjectCurrentDirectory();
+  const {
+    project,
+    directoryId,
+    loading,
+    canRead,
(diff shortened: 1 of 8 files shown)


glassblowerscat:
@CarsonF I'm not wild about what I'm doing here in `FilesOverviewCard`; it seems clunky. But it's really just a more specific `FieldOverviewCard`, and I was reluctant to change that component without input from you. Do you want to update `FieldOverviewCard` to give it a "redacted" state, or do you want to ditch using it in `FilesOverviewCard` so we can implement redaction there?

CarsonF:
I think it definitely makes sense to implement redaction in FieldOverviewCard. Maybe it could be similar to DataButton? DataButton uses a render prop (children) instead of requiring formatted data to be passed in.
```diff
-children={formatText(prop.value)}
+data={prop}
+children={formatText}
```
If that makes sense. 

It handles the 4 loading, redacted, empty, data states and let's you declare what each of those are as props.
That's easy though when it's just a text value, as DataButton is. Not so much for these cards which have multiple dynamic pieces.

src/scenes/Projects/Files/CreateProjectDirectory.tsx
```diff
@@ -17,9 +17,25 @@ export const CreateProjectDirectory = (
...
-  const { project, directoryId } = useProjectCurrentDirectory();
+  const {
+    project,
+    directoryId,
+    loading,
+    canRead,
+  } = useProjectCurrentDirectory();
   const { enqueueSnackbar } = useSnackbar();
 
+  if (canRead === false) {
```
CarsonF, line 28: Is this necessary? They can't even get to this code.

src/scenes/Projects/Files/ProjectFilesList.tsx
```diff
@@ -114,16 +115,29 @@ const ProjectFilesListWrapped: FC = () => {
...
     loading: directoryLoading,
+    canRead: canReadRootDirectory,
     project,
     directoryId,
     rootDirectoryId,
   } = useProjectCurrentDirectory();
 
+  if (canReadRootDirectory === false) {
+    enqueueSnackbar(
+      `You don't have permission to view files for this project`,
```
CarsonF, line 132: Something permanent would probably be better. The card could be replaced with typography. Breadcrumbs would still be shown to allow easy back.

Another similar case: https://github.com/SeedCompany/cord-field/blob/e9310f0eb169b9ca9b03029a109afeed06370c27/src/scenes/Languages/Detail/LanguageDetail.tsx#L223-L226

src/scenes/Projects/Files/CreateProjectDirectory.tsx
glassblowerscat: Yeah, not really. But I had already implemented it before I updated to make it unreachable, so I just left it for thoroughness.

src/scenes/Projects/Files/ProjectFilesList.tsx
```diff
@@ -268,7 +270,11 @@ const ProjectFilesListWrapped: FC = () => {
     }
   };
 
-  return directoryLoading ? null : (
+  return directoryLoading ? null : canReadRootDirectory === false ? (
+    <Typography color="textSecondary">
+      You don't have permission to see the projects this language is engaged in
```
CarsonF, line 275: copy/paste error

CarsonF (commented):
Looks good, just that one error message. Ready to merge API PR?

src/scenes/Projects/Files/ProjectFilesList.tsx
glassblowerscat: Guh. Fixed

glassblowerscat:
> Ready to merge API PR?

I'm good with that

CarsonF:
API change deploying right now. Waiting a few mins for deployment to finish then we'll re-run CI

CarsonF:
@glassblowerscat failing now for a legit reason

glassblowerscat:
> @glassblowerscat failing now for a legit reason

Curses!

glassblowerscat:
> @glassblowerscat failing now for a legit reason

Fixed now

CarsonF (approved):

PR #416, opened 2020-08-31
Refactor Edit Engagement

CarsonF:
- Fields are defined as a static mapping outside of the component
  instead of a bunch if/else statements inside it
- Edit field(s) are now strict keys instead of any/magic string
- Fixed missing initial values for language engagements.
  Previously they were added manually on the defaultValue of firstScripture & lukePartnership fields.

### DIFF
--- a/src/scenes/Engagement/EditEngagement/EditEngagementDialog.tsx
+++ b/src/scenes/Engagement/EditEngagement/EditEngagementDialog.tsx
@@ -1,12 +1,12 @@
 import { pick, startCase } from 'lodash';
-import React, { FC } from 'react';
-import { Except } from 'type-fest';
+import React, { ComponentType, FC } from 'react';
+import { Except, Merge } from 'type-fest';
 import {
   displayInternPosition,
   InternshipEngagementPositionList,
   MethodologyToApproach,
-  UpdateInternshipEngagementInput,
-  UpdateLanguageEngagementInput,
+  UpdateInternshipEngagement,
+  UpdateLanguageEngagement,
 } from '../../../api';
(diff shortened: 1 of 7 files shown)


src/scenes/Engagement/EditEngagement/EditEngagementDialog.tsx
```diff
@@ -26,186 +25,172 @@ import {
...
-    | LanguageEngagementDetailFragment;
-  editValue?: string;
+type Engagement = InternshipEngagement | LanguageEngagement;
+type EngagementKeys = keyof InternshipEngagement | keyof LanguageEngagement;
+
+interface EngagementFieldProps {
+  engagement: Engagement;
+}
+
+const fieldMapping = {
```
glassblowerscat, line 43: Okay, so this is very pretty and satisfying, and the nested ternary operators below are messy and feel brittle, but if the DX is so bad on this, could you instead keep the ternaries and just create yourself an Enum instead of all the magic strings? That'd be a compromise and wouldn't feel as nice, but **would** still provide the missing strictness you're looking for.

If you're fine with the DX, then otherwise this looks good to me.

CarsonF: The code is strict. I think the DX is bad because the validation isn't on the key/value pair but on an unused variable below. But on the other hand the DX for the object mapping in general is way better than the ternaries IMO.
Let me try duplicating the keys in a string literal and see how that looks.

CarsonF: Ok pushed. I'm more happy with that anyways. Thanks for the push (ha) 👍 Lemme know what you think.

PR #427, opened 2020-09-06
Engagement Files

### LINKED ISSUE #404: PnP/Growth Plan File

glassblowerscat:
Closes #404 

This wires up our Language and Internship Engagement Detail screens to the file management API and Context already created for Project files.

Both detail screens now have a new section header and card for PNP/Growth Plan. The card UI rendered is based on both the user's permissions and whether a PNP/Growth Plan already exists for that engagement:

- If no file exists, show an `AddItemCard`, a new component that can use `react-dropzone` and a passed-in `handleFileSelect`
  - if `canEdit`, adding a file for upload will result in a new PNP/Growth Plan being uploaded
  - If `!canEdit`, uploading will be disabled and a `Tooltip` will be shown with a redaction message
- If a file exists, show an `EngagementFileCard` (another new component)
  - If `canRead`, show basic info and a `FileActionsMenu` containing actions appropriate to Engagement Files
    - if `canEdit`, allow uploading a new version by dragging and dropping again onto the `EngagementFileCard` or selection "New Version" from the File Actions Menu
    - if `!canEdit`, disable new version uploading options
  - if `!canRead`, show redaction message and no `FileActionsMenu`

~~Our original `FileActionsContext` was designed without a complete understanding of how this part of the API would work, so we also modify it in this PR to require a `context` prop to be passed in to the Provider with a value of `'project' | 'engagement'`. We then use this to determine certain display options. The `FileActionsMenu` also now requires an `onVersionUpload` prop, because Engagements do not use the same `createFileVersion` mutation as Projects.~~

Our original `FileActionsContext` was designed around Project files and was therefore too opinionated. Specifically, it assumed certain subsets of `FileActions` (`Rename`, `Download`, `NewVersion`, `History`, `Delete`) would be available for Directories, Files, and File Versions. For defined files like PNP or Growth Plan, we may not want the same subsets of actions. Also, Project Files perform a different mutation when a new file version is uploaded than we will be using in some other situations. This updates the `FileActionsContext` and `FileActionsMenu` as follows:

- Update `FileActionsMenu` to require an `onVersionUpload` callback prop when it is passed an `item` that is not a `FileVersion`. This allows the `EngagementFileCard` component to control how the upload of the new version is handled
- Update `FileActionsMenu` to require an `actions` prop that is an array of `FileAction`s, thereby moving the logic for determining which actions will be shown up to the parent component
- Update the instance of `useDialog` in `FileActionsContext` that is used for opening the version history to take an object of type `{ item: FileActionItem, actions: FileAction[] }` instead of just an `item: FileActionItem`. This solves the problem of how to pass the permissible actions through from a file to its child `FileVersionItem`, which also displays a `FileActionsMenu` but should remove the actions for viewing file History and creating a new File Version. We also then tighten up the typings of the `handleFileActionClick` function returned by the `FileActionsContext` to pass through the `actions` prop when `FileAction.NewVersion` is called.

We also just generally make typings both a little more specific and a little less brittle throughout the `FileActionsContext` and its consumers and children.


CarsonF:
Haven't looked yet (1k added 😬 ), but just so you know this kind of thing will be done outside of engagements as well. We also have 2 files on Partnerships and 1 on Budget. Does that change things with naming or file/code placement?

glassblowerscat:
> We also have 2 files on Partnerships and 1 on Budget.

Will they work similarly to the files on Engagements? If so, I think we can probably just add `'partnership'` and `'budget'` to supported `context` values for passing to `FileActionsContextProvider`

CarsonF:
Yes they work similarly. Though the need to call that out explicitly the context is concerning, but I haven't seen the code yet. Maybe tomorrow...or next week.

glassblowerscat:
> Though the need to call that out explicitly the context is concerning

I'm not crazy about it, either. I was willing to live with it when I thought there were only two use cases, but I think I'm going to have to look at it again

glassblowerscat:
@CarsonF This is in much better shape now; should be ready for review again

glassblowerscat:
@CarsonF This is ready again, again

CarsonF:
- [ ] I'm getting an error trying to delete a file version. Something about Apollo cannot refetch a query defined with `cache-only` policy.
- [ ] Can we remove the title outside of the card here since it's defined with file name / placeholder text.
Screen Shot 2020-10-06 at 4 48 24 PM
- [ ] Can we rename the file name to `Planning and Progress` and placeholder text to `Add Planning and Progress File`

CarsonF:
New version as the `outline` css that different from the rest of the menu items.
Screen Shot 2020-10-06 at 4 52 08 PM

glassblowerscat:
> * [ ]  I'm getting an error trying to delete a file version. Something about Apollo cannot refetch a query defined with `cache-only` policy.

This bug is present in `v3` and will be fixed in #457. Do you want me to bring that change forward into this PR or live with it for now?

CarsonF:
If it's already fixed in a PR that depends on this one just leave it there 👍

glassblowerscat:
> New version as the `outline` css that different from the rest of the menu items.

Fixed

glassblowerscat:
@CarsonF This is ready to go, including fix for the styling regression on the Engagement Detail scenes

src/scenes/Engagement/InternshipEngagement/InternshipEngagementDetail.tsx
```diff
@@ -64,156 +75,228 @@ export const InternshipEngagementDetail: FC<EngagementQuery> = ({
...
+                        uploadFile({
+                          action: 'version',
+                          files,
+                          parentId: engagement.id,
+                        })
+                      }
+                      resourceType="engagement"
+                      securedFile={engagement.growthPlan}
+                    />
+                  )}
```
CarsonF, line 262: Maybe it would be better to split this out into a separate GrowthPlan component? There's a lot going on in this file. I think we could move the useEngagementFile hook there and then only wrap this section with the context instead of the whole component. Thoughts?

CarsonF:
Something got messed up here in the rebase. 
Screen Shot 2020-10-07 at 7 23 53 PM

src/scenes/Engagement/Files/useUploadEngagementFiles.ts
```diff
@@ -0,0 +1,54 @@
...
+    uploadId,
+    name,
+    parentId: id,
+    action,
+  }) => {
+    await uploadFile({
+      variables: {
+        id,
+        pnp: { uploadId, name },
+        growthPlan: { uploadId, name },
```
CarsonF, line 36: Does Apollo/GQL not complain about unused variables here?

src/scenes/Engagement/Files/useUploadEngagementFiles.ts
```diff
@@ -0,0 +1,54 @@
...
+      ],
+    });
+  };
+
+  const uploadEngagementFile = ({
+    action,
+    files,
+    parentId,
+  }: UploadFilesConsumerInput) =>
+    uploadFiles({ action, files, handleUploadCompleted, parentId });
```
CarsonF, line 51: I'm seeing a few places where we accept/pass around a list of files. I don't think it makes sense to upload more than one file at a time for these defined files, regardless of first or not. Do we limit this in some spot? If it's easier to pass around the list and then filter down to a single file in one spot then that makes sense too.

CarsonF (commented):
There's a lot going on here lol. I'm not going to nit pick this one as I know we'll have more changes coming anyways. But I do want to call out a couple things:

src/scenes/Engagement/InternshipEngagement/InternshipEngagementDetail.tsx
glassblowerscat: There's a lot going on, but not much of it is dedicated to handling the logic of whether it's being used for a Growth Plan vs. a PNP. I'm not sure we'd gain a lot by splitting it up. But I think we can wrap just the card with the context without doing that. I don't think `useUploadEngagementFile` hook needs to be inside the context; it just needs to be inside the `UploadContext`

src/scenes/Engagement/Files/useUploadEngagementFiles.ts
glassblowerscat: Not so far! I initially did something different that didn't involve passing these, but I switched to this because it was cleaner, and I didn't get any warnings or errors in the console.

src/scenes/Engagement/Files/useUploadEngagementFiles.ts
glassblowerscat: We can limit it at the Dropzone, which I'm pretty sure we do for Engagement files. But even if you limit to one it always passes that one around in an `Array`

glassblowerscat: Well, I'll be, we actually weren't limiting it. Fixed now

glassblowerscat:
> Something got messed up here in the rebase.
 
Really? It looks fine for me. Do you have my latest with the fixes for the other styling issues?

src/scenes/Engagement/InternshipEngagement/InternshipEngagementDetail.tsx
glassblowerscat: Okay, I made this change, and it worked just fine

glassblowerscat:
> Really? It looks fine for me. Do you have my latest with the fixes for the other styling issues?

Oh wait, nm. I was looking at the wrong part of the screen 🤦‍♀️

glassblowerscat:
> Oh wait, nm. I was looking at the wrong part of the screen 🤦‍♀️

Fixed now

src/scenes/Engagement/InternshipEngagement/InternshipEngagementDetail.tsx
glassblowerscat: > There's a lot going on in this file.

Sorry, I just realized you probably meant "the `InternshipEngagementDetail` file, not the `DefinedFileCard` file. In which case, yes, but no more so than in some other scenes that have conditional display of components. All the other "detail" scene files have all their UI elements right there in the file... it seems like kind of a shame to break that pattern when it's so predictable. It's kind of nice not to have to click around through multiple layers to work on some part of a detail scene

CarsonF (approved):
Awesome, lets do it! :tada:

PR #430, opened 2020-09-07
Prevent submitting form via enter key on LookupField while searching

### LINKED ISSUE #424: Swallow enter to submit form on lookup field while loading
I noticed people hitting enter to start the search for lookup fields. This cause the form to say field was required and missing (since the value is null until object selected). In order to make this UX nicer we should just do nothing when enter is hit while loading.

jmc637:
Fixes #424

### DIFF
--- a/src/components/form/Lookup/LookupField.tsx
+++ b/src/components/form/Lookup/LookupField.tsx
@@ -235,9 +235,10 @@ export function LookupField<
       inputValue={input}
       onBlur={field.onBlur}
       onFocus={field.onFocus}
-      onKeyDown={(_) => {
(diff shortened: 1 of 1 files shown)


jmc637:
Hi, this is to swallow enter while loading. Is this LookupField being used anywhere besides the storybook stories? I wanted to test this inside the app to make sure it is working as intended.

CarsonF (approved):
Thanks @jmc637!

> Is this LookupField being used anywhere besides the storybook stories? I wanted to test this inside the app to make sure it is working as intended.

Yes for create language/internship engagement (on the project overview page)

I just tried it myself and it works great!

PR #431, opened 2020-09-08
Bugfix:forms reverting to initial values

### LINKED ISSUE #421: Forms revert to show initial values while submitting
While submitting, form fields change current values to initial values. This doesn't affect submitted data only UI.
Example can be seen with project step update, just throttle network to make it obvious.

frankyma:
Memoize all initialValues in final form with `useMemo`

**Testing**
When updating a field, click submit and the field shouldn't flash back its initial value

Fixes #421

### DIFF
--- a/src/scenes/Engagement/CeremonyCard/CeremonyCard.tsx
+++ b/src/scenes/Engagement/CeremonyCard/CeremonyCard.tsx
@@ -10,8 +10,7 @@ import {
 } from '@material-ui/core';
 import { Skeleton } from '@material-ui/lab';
 import clsx from 'clsx';
-import * as React from 'react';
-import { FC } from 'react';
+import React, { FC, useMemo } from 'react';
 import { canEditAny, UpdateCeremonyInput } from '../../../api';
 import { useDialog } from '../../../components/Dialog';
 import { DialogForm } from '../../../components/Dialog/DialogForm';
@@ -78,6 +77,17 @@ export const CeremonyCard: FC<CeremonyCardProps> = ({
     >
       Edit dates
     </Button>
+  );
+
+  const initialValues = useMemo(
+    () => ({
+      ceremony: {
+        id: id || '',
+        estimatedDate: estimatedDate?.value,
+        actualDate: actualDate?.value,
+      },
+    }),
+    [actualDate?.value, estimatedDate?.value, id]
   );
 
   return (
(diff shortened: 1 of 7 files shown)


src/scenes/Engagement/CeremonyCard/CeremonyCard.tsx
```diff
@@ -80,6 +79,17 @@ export const CeremonyCard: FC<CeremonyCardProps> = ({
...
 
+  const initialValues = useMemo(
+    () => ({
+      ceremony: {
+        id: id || '',
+        estimatedDate: estimatedDate?.value,
+        actualDate: actualDate?.value,
+      },
+    }),
+    [actualDate?.value, estimatedDate?.value, id]
```
CarsonF, line 90: I wonder for performance reasons if we should do `[ceremony]` and ignore the lint error. It's pretty straight forward for us here that all of these props are coming from a single variable. Each of these items in the array has to be checked on every render to see if it's changed, so this would be a _small_ performance hit.

Same for really all of these cases in this PR. I think this is a simple case of we know better than the linter.
@glassblowerscat @frankyma what do yall think?

frankyma: It's a marginal decision... and complicated.

First, You won't get a hooks lint error if you just include the entire object.  I should've made a comment but I did that here because TS was giving me errors in the array: https://github.com/SeedCompany/cord-field/pull/431/files#diff-1af1b9db776b34e03ec0e39c00d815a4R217.

In this specific Ceremony example we'll get an error becasue `ceremony` is nullable so we'll need a slight rewrite (like https://github.com/SeedCompany/cord-field/pull/431/files#diff-6c47258f6b951919447de6b601fac666R18-R19) in the useMemo.  Not a big deal.

The tradeoff is a slight performance gain, versus thinking about whether or not we want the memoized value to change if the entire object changes.  In our cases I don't think it matters, but generally I just like to use my ide suggestion to auto-populate the dependencies and not have to worry about it

glassblowerscat: Premature optimization is the root of all evil. I'm a much bigger fan of doing what seems straightforward and un-clever, then going on an optimization hunt later if we have issues

CarsonF: Touché 👍

CarsonF (approved):

PR #433, opened 2020-09-09
Add `noData` prop on `PdfPreview` `Document` component

### LINKED ISSUE #418: Show pdf loading spinner while requesting download
Currently you get "no pdf file specified" and then a spinner (and then the preview).
Easy to see with network throttled.

glassblowerscat:
Closes #418 

The PDF Previewer is a little different than our other previewers; the third-
party library we use implements its own state tracking, and it has a `noData`
component unlike anything we use. This caused a "No PDF file specified" message
to be displayed when `PdfPreview` had been mounted but the `loading` state
hadn't yet been set to true by the main `FilePreview` component. We fix this
by telling `noData` to display the same `PreviewLoading` component we use for
the loading state.


CarsonF (approved):
Easy peasy

PR #435, opened 2020-09-10
Create partnership dialog

### LINKED ISSUE #405: Create Partnership Dialog

glassblowerscat:
Close #405 

- Create a `PartnershipForm` that can be used to either create or update a partnership
- Refactor `EditPartnership` component to consume `PartnershipForm`
- Create `CreatePartnership` component that also consumes `PartnershipForm
- Wire up `CreatePartnership` to "Add" button on `/projects/:projectId/partnerships`
- Add `onlyDirtySubmit={false}` prop to `PartnershipForm` in `EditPartnership` to fix bug preventing partnership from being deleted when no fields in the form have been changed

### DIFF
--- a/src/components/Dialog/DialogForm.tsx
+++ b/src/components/Dialog/DialogForm.tsx
@@ -41,7 +41,7 @@ export type DialogFormProps<T, R = void> = Omit<
   CloseProps?: ButtonProps;
 
   /** Only call onSubmit if form is dirty, else just close dialog. */
-  onlyDirtySubmit?: boolean | string;
+  sendIfClean?: boolean | string;
 
   /** A prefix for all form fields */
   fieldsPrefix?: string;
@@ -100,7 +100,7 @@ export function DialogForm<T, R = void>({
   SubmitProps,
   closeLabel,
   CloseProps,
-  onlyDirtySubmit = true,
+  sendIfClean,
   open,
   onSuccess,
   errorHandlers,
@@ -121,14 +121,13 @@ export function DialogForm<T, R = void>({
       onSubmit={async (data, form) => {
         const submitAction = (data as any).submitAction;
         if (
-          (onlyDirtySubmit === true && !form.getState().dirty) ||
-          (typeof onlyDirtySubmit === 'string' &&
-            onlyDirtySubmit !== submitAction)
+          (!sendIfClean && !form.getState().dirty) ||
+          (typeof sendIfClean === 'string' && sendIfClean !== submitAction)
         ) {
           onClose?.('cleanSubmit', form);
           return null;
         }
-        if (!onlyDirtySubmit || onlyDirtySubmit === submitAction) {
+        if (sendIfClean || sendIfClean === submitAction) {
           try {
             const res = await onSubmit(data, form);
             onSuccess?.(res);
--- a/src/components/Dialog/DialogForm.tsx
+++ b/src/components/Dialog/DialogForm.tsx
@@ -131,15 +131,13 @@ export function DialogForm<T, R = void>({
           onClose?.('cleanSubmit', form);
           return null;
         }
-        if (sendIfClean === true || sendIfClean === submitAction) {
-          try {
-            const res = await onSubmit(data, form);
-            onSuccess?.(res);
-            onClose?.('success', form);
-            return null;
-          } catch (e) {
-            return await handleFormError(e, errorHandlers);
-          }
+        try {
+          const res = await onSubmit(data, form);
+          onSuccess?.(res);
+          onClose?.('success', form);
+          return null;
+        } catch (e) {
+          return await handleFormError(e, errorHandlers);
         }
       }}
     >
--- a/src/components/Dialog/DialogForm.tsx
+++ b/src/components/Dialog/DialogForm.tsx
@@ -40,7 +40,11 @@ export type DialogFormProps<T, R = void> = Omit<
   closeLabel?: ReactNode;
   CloseProps?: ButtonProps;
 
-  /** Only call onSubmit if form is dirty, else just close dialog. */
+  /**
+   * Optionally call onSubmit even if form is clean,
+   * OR, when a string value, if `sendIfClean === data.submitAction`.
+   * Else just close dialog.
+   */
   sendIfClean?: boolean | string;
 
   /** A prefix for all form fields */
--- a/src/components/Dialog/DialogForm.tsx
+++ b/src/components/Dialog/DialogForm.tsx
@@ -131,7 +131,7 @@ export function DialogForm<T, R = void>({
           onClose?.('cleanSubmit', form);
           return null;
         }
-        if (sendIfClean || sendIfClean === submitAction) {
+        if (sendIfClean === true || sendIfClean === submitAction) {
           try {
             const res = await onSubmit(data, form);
             onSuccess?.(res);
--- a/src/components/Dialog/DialogForm.tsx
+++ b/src/components/Dialog/DialogForm.tsx
@@ -41,7 +41,7 @@ export type DialogFormProps<T, R = void> = Omit<
   CloseProps?: ButtonProps;
 
   /** Only call onSubmit if form is dirty, else just close dialog. */
-  onlyDirtySubmit?: boolean;
+  onlyDirtySubmit?: boolean | string;
 
   /** A prefix for all form fields */
   fieldsPrefix?: string;
@@ -119,17 +119,24 @@ export function DialogForm<T, R = void>({
       decorators={defaultDecorators}
       {...FormProps}
       onSubmit={async (data, form) => {
-        if (onlyDirtySubmit && !form.getState().dirty) {
+        const submitAction = (data as any).submitAction;
+        if (
+          (onlyDirtySubmit === true && !form.getState().dirty) ||
+          (typeof onlyDirtySubmit === 'string' &&
+            onlyDirtySubmit !== submitAction)
+        ) {
           onClose?.('cleanSubmit', form);
           return null;
         }
-        try {
-          const res = await onSubmit(data, form);
-          onSuccess?.(res);
-          onClose?.('success', form);
-          return null;
-        } catch (e) {
-          return await handleFormError(e, errorHandlers);
+        if (!onlyDirtySubmit || onlyDirtySubmit === submitAction) {
+          try {
+            const res = await onSubmit(data, form);
+            onSuccess?.(res);
+            onClose?.('success', form);
+            return null;
+          } catch (e) {
+            return await handleFormError(e, errorHandlers);
+          }
         }
       }}
     >
--- a/src/components/form/Lookup/LookupField.tsx
+++ b/src/components/form/Lookup/LookupField.tsx
@@ -313,7 +313,7 @@ export function LookupField<
         <CreateDialogForm
           {...createDialogState}
           initialValues={createInitialValues}
-          onlyDirtySubmit={false}
+          sendIfClean={true}
           onSuccess={(newItem: T) => {
             field.onChange(
               multiple ? [...(field.value as T[]), newItem] : newItem
--- a/src/scenes/Partnerships/Create/CreatePartnership.tsx
+++ b/src/scenes/Partnerships/Create/CreatePartnership.tsx
@@ -1,42 +1,50 @@
-import { useSnackbar } from 'notistack';
 import React from 'react';
 import { Except } from 'type-fest';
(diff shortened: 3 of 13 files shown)


src/scenes/Partnerships/Create/CreatePartnership.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { PartnershipForm, PartnershipFormProps } from '../PartnershipForm';
+import { useCreatePartnershipMutation } from './CreatePartnership.generated';
+
+export interface CreatePartnershipFormInput {
+  partnership: Pick<
+    CreatePartnershipType,
+    'projectId' | 'types' | 'fundingType'
+  > & {
+    organizationId: OrganizationLookupItem;
+  };
```
CarsonF, line 17: I'm assuming you are just trying to change the type of `organizationId`. If that's the case, `Merge` from `type-fest` would probably better indicate your intentions and doesn't require you to list the inverse set explicitly.
```tsx
Merge<CreatePartnershipType, { organizationId: OrganizationLookupItem }>
```

src/scenes/Partnerships/Edit/EditPartnership.tsx
```diff
@@ -102,19 +74,20 @@ export const EditPartnership: FC<EditPartnershipProps> = ({
   );
 
   return (
-    <DialogForm<UpdatePartnershipInput & SubmitAction<'delete'>>
+    <PartnershipForm<EditPartnershipFormInput>
       {...props}
-      onSubmit={async (input) => {
+      onlyDirtySubmit={false} // Lets us delete without changing any fields
```
CarsonF, line 79: Heck. Maybe we should also allow a string here for to limit to this functionality to the specified submit actions?

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { PartnershipFormFragment } from './PartnershipForm.generated';
+
+export type PartnershipFormProps<T> = DialogFormProps<T> & {
+  partnership?: PartnershipFormFragment;
+};
+
+export const hasManagingType = (types: Nullable<readonly PartnershipType[]>) =>
+  types?.includes('Managing') ?? false;
+
+export const PartnershipForm = <
```
CarsonF, line 35: Since you are here, can you wrap all of these fields in `SecuredField`'s (except org obv)?

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+        row
+        disabled={partnership ? !partnership.types.canEdit : false}
+      >
+        {([
+          'Managing',
+          'Funding',
+          'Impact',
+          'Technical',
+          'Resource',
+        ] as PartnershipType[]).map((type: PartnershipType) => (
```
CarsonF, line 71: This can be pulled from the generated list now.

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const radioOptions = PartnershipStatuses.map((status) => (
+    <RadioOption
+      key={status}
+      label={displayPartnershipStatus(status)}
+      value={status}
+    />
+  ));
+
+  return (
+    <DialogForm<T> {...rest}>
```
CarsonF, line 50:
```suggestion
    <DialogForm<T> {...rest} fieldsPrefix="partnership">
```
Now that it's available which is also necessary for it to be compatible with `SecuredField` usage

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+export const hasManagingType = (types: Nullable<readonly PartnershipType[]>) =>
+  types?.includes('Managing') ?? false;
+
+export const PartnershipForm = <
+  T extends CreatePartnershipFormInput | EditPartnershipFormInput
+>({
+  partnership,
+  ...rest
+}: PartnershipFormProps<T>) => {
+  const radioOptions = PartnershipStatuses.map((status) => (
```
CarsonF, line 41: `PartnershipStatuses` can be replaced with generated list

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+        disabled={partnership ? !partnership.types.canEdit : false}
+      >
+        {([
+          'Managing',
+          'Funding',
+          'Impact',
+          'Technical',
+          'Resource',
+        ] as PartnershipType[]).map((type: PartnershipType) => (
+          <Box width="50%" key={type}>
```
CarsonF, line 72: Can you replace `Box` with another styling solution?

src/scenes/Partnerships/Create/CreatePartnership.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+export const CreatePartnership = ({
+  projectId,
+  ...props
+}: CreatePartnershipProps) => {
+  const [createPartnership] = useCreatePartnershipMutation();
+
+  return (
+    <PartnershipForm<CreatePartnershipFormInput>
+      title="Create Language"
```
CarsonF, line 34:
```suggestion
      title="Create Partnership"
```

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+    />
+  ));
+
+  return (
+    <DialogForm<T> {...rest}>
+      <SubmitError />
+      {!partnership && (
+        <OrganizationField
+          name="partnership.organizationId"
+          label="Organization"
```
CarsonF, line 55:
```suggestion
```
It's defaulted now.

CarsonF (changes requested):
Thanks @glassblowerscat. Can I ask you make some more changes to bring this up to date with recent changes?

src/scenes/Partnerships/Create/CreatePartnership.tsx
glassblowerscat: I was also narrowing the field types to leave out several that we allow in `EditPartnership` but don't allow here

CarsonF: Got it, no worries then.

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
glassblowerscat: Haha, your nemesis!

src/scenes/Partnerships/Edit/EditPartnership.tsx
glassblowerscat: So maybe change the prop from `onlyDirtySubmit` to something like `allowCleanSubmit: 'never' | 'always' | 'delete'`, and default would be `never`?

CarsonF: I do like changing the prop name so the default is false. I like "dirty" -> "clean", but "allow" feels wrong. We "allow" clean submits, they just do nothing.

Submit action names are dynamic though; they can be whatever string you specify. So that's why I was thinking of using a boolean to distinguish between a submit action or applying it to all.

glassblowerscat: Okay, so you're just thinking the type of `onlyDirtySubmit` would be `boolean | string | undefined`, default to `false`, and we just check for a string value and compare to `input.SubmitAction`?

I'm not married to `allowCleanSubmit`, but to argue against your objection, `onlyDirtySubmit` implies a similar (opposite) thing: that we don't "allow" dirty submits

glassblowerscat: So the `onSubmit` prop of `DialogForm` would look like this:

```
onSubmit={async (data, form) => {
    if (onlyDirtySubmit === true && !form.getState().dirty) {
      onClose?.('cleanSubmit', form);
      return null;
    }
    if (
      !onlyDirtySubmit ||
      onlyDirtySubmit === (data as any).submitAction
    ) {
      try {
        const res = await onSubmit(data, form);
        onSuccess?.(res);
        onClose?.('success', form);
        return null;
      } catch (e) {
        return await handleFormError(e, errorHandlers);
      }
    }
  }}
```

CarsonF: > I'm not married to `allowCleanSubmit`, but to argue against your objection, `onlyDirtySubmit` implies a similar (opposite) thing: that we don't "allow" dirty submits

Yeah I agree. Maybe we just remove the option all together. I can't even think of a use case where we want to send clean submissions. If we know it's clean, there's no point. Actually scratch that - delete is a use case 🤦 . What about `sendCleanSubmissions`?
I still like changing it to "clean" because those are the exceptions. Maybe "send" is the verb we are looking for, or something similar.

glassblowerscat: > What about `sendCleanSubmissions`?

Sounds good to me!

CarsonF: So the default then would be false, and any strings given that match the submit action would call onSubmit even if clean (skip the dirty check).

Change your mind on `sendIfClean`? I could get behind that too.

glassblowerscat: I didn't as much change my mind as decide it wasn't worth haggling over it. Up to you

CarsonF: Haha I'm feeling the same way. I'm good with anything like that

glassblowerscat: `sendIfClean` it is! 'Cuz it's shorter

CarsonF: Could you update the prop description as well?

glassblowerscat: 🤪

glassblowerscat: @CarsonF Ready again

src/components/Dialog/DialogForm.tsx
```diff
@@ -119,17 +123,23 @@ export function DialogForm<T, R = void>({
...
           return null;
         }
-        try {
-          const res = await onSubmit(data, form);
-          onSuccess?.(res);
-          onClose?.('success', form);
-          return null;
-        } catch (e) {
-          return await handleFormError(e, errorHandlers);
+        if (sendIfClean || sendIfClean === submitAction) {
```
CarsonF, line 134: Is this if statement catching anything different than the else case of the if statement above?

src/components/form/Lookup/LookupField.tsx
```diff
@@ -313,7 +313,7 @@ export function LookupField<
         <CreateDialogForm
           {...createDialogState}
           initialValues={createInitialValues}
-          onlyDirtySubmit={false}
+          sendIfClean={true}
```
CarsonF, line 316:
```suggestion
          sendIfClean="delete"
```

src/components/Dialog/DialogForm.tsx
glassblowerscat: It should have been `if (sendIfClean === true || sendIfClean === submitAction)`

`sendIfClean` could have a value other than `true` but that doesn't match `submitAction`; that's what I was trying to catch

glassblowerscat: (It's fixed now)

glassblowerscat: Actually, this just needs to be reworked a bit....

glassblowerscat: Okay, derpderpderp, the real problem is **Ryan can't read** and didn't see the second line in the first if statement:

`(typeof sendIfClean === 'string' && sendIfClean !== submitAction)`

So you're right, @Carson, that second if statement was useless and has now been removed.

CarsonF: 😆  Cheers 🍻

CarsonF (approved):
Thanks @glassblowerscat! :shipit:

PR #436, opened 2020-09-10
Progressive download of previews for image, audio, and video files

### LINKED ISSUE #422: Possible to stream videos directly from S3
@glassblowerscat Is it possible to stream videos directly from S3 instead of loading the full file into memory first?
I noticed Seth had a 90MB video that took forever to load because of this.

glassblowerscat:
Closes #422 

We'll be letting the HTML tags for these types handle the downloading of the source file from our presigned URL instead of first downloading the file, passing it to the `NativePreview` component, then creating a local URL to feed to the tag. This will let the HTML tag start playback as soon as it's able to, based on download state.

We primarily wanted this for videos, because their large file size can make wait times pretty long if playback can't start until the file is completely downloaded. Unfortunately, results will probably still be mixed. This PR removes the obstacles to progressive playback, but if a video file doesn't put its metadata at the front of the file, the player still won't be able to play right away. This is outside of our control, but it's good to set expectations.

- Add logic to `FilePreview` to skip pre-downloading the file for one of the
      above types, instead letting the rendered HTML tag handle download
- Update props for `NativePreview` to take a `string` value for `file` instead
      of `File`
- Update `NativePreview` component to pass `file` prop directly to HTML tags
- Add stricter logic for returning image, audio, and video types that are previewable by user

### DIFF
--- a/src/components/files/FilePreview/FilePreview.tsx
+++ b/src/components/files/FilePreview/FilePreview.tsx
@@ -20,7 +20,7 @@ import { useGetFileDownloadUrl } from '../hooks';
 import { CsvPreview } from './CsvPreview';
 import { ExcelPreview } from './ExcelPreview';
 import { HtmlPreview } from './HtmlPreview';
-import { NativePreview, NativePreviewType } from './NativePreview';
+import { NativePreview } from './NativePreview';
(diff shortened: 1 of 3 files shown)


CarsonF (approved):
LGTM!

PR #438, opened 2020-09-10
Update Project Name Dialog

glassblowerscat:
- Update `UpdateProjectDialog` to include `name` as one of the editable fields
- Update `ProjectOverview` to add edit button next to project name and wire up to `UpdateProjectDialog`

### DIFF
--- a/src/scenes/Projects/Overview/ProjectOverview.tsx
+++ b/src/scenes/Projects/Overview/ProjectOverview.tsx
@@ -143,7 +143,7 @@ export const ProjectOverview: FC = () => {
             {data?.project.name.canEdit && (
               <Fab
                 color="primary"
-                aria-label="edit language"
+                aria-label="edit project name"
                 onClick={() => editField(['name'])}
               >
                 <Edit />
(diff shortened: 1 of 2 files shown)


src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -118,20 +125,31 @@ export const ProjectOverview: FC = () => {
...
-              )
-            ) : (
-              <Skeleton width="50%" />
+                <Skeleton width="50%" />
+              )}
+            </Typography>
+            {data?.project.name.canEdit && (
+              <Fab
+                color="primary"
+                aria-label="edit language"
```
CarsonF, line 146:
```suggestion
                aria-label="edit project name"
```

CarsonF (changes requested):

src/scenes/Projects/Overview/ProjectOverview.tsx
glassblowerscat: D'oh

glassblowerscat:
So here's a fun thing....

image

This was the project I edited (three times) to test this PR. This seems like an API issue, but what do you think, @CarsonF

CarsonF:
Nice! Definitely an API issue. I think @frankyma discovered something similar yesterday as well.

frankyma:
Yeah it looks like the bug I'm working on now.  Basically our update [update query match](https://github.com/SeedCompany/cord-api-v3/blob/master/src/core/database/database.service.ts#L484-L499) was resolving to multiple paths (since there's multiple security groups a baseNode is a part of), and each path runs the create statement once.  @michaelmarshall figured this out and we're putting a limit 1 to the match

CarsonF (approved):
Thanks @glassblowerscat!

PR #443, opened 2020-09-11
Connect create user form to user field

CarsonF:


src/components/form/Lookup/User/UserField.tsx
```diff
@@ -1,13 +1,27 @@
...
   useLookup: useUserLookupLazyQuery,
   getOptionLabel: (option) => option.fullName,
   label: 'Person',
   placeholder: 'Search for a person by name',
+  CreateDialogForm: CreateUser,
+  getInitialValues: (val): Partial<CreatePersonInput> => ({
+    // @ts-expect-error the partial type doesn't match and the generic is not
+    // being passed around everywhere yet.
+    person: {
+      realFirstName: val,
```
blakekrammes, line 23: I was thinking that val should be split by whitespace and populate first and last names. If nothing after whitespace, then just do first name.

blakekrammes:
```typescript
getInitialValues: (val): Partial<CreatePersonInput> => {
  const [first, last] = val.split(' ');
  return { 
    person: {
      realFirstName: first,
      realLastName: last,
      displayFirstName: first,
      displayLastName: last,
    }
  }
}



```

CarsonF:
It was previously not required. PR? ☺️

PR #444, opened 2020-09-11
Fix useDialog error `Cannot read property 'persist' of undefined`

glassblowerscat:


glassblowerscat:
Aw nerds, I accidentally based this on a different branch than `v3`

CarsonF:
It's easy to rebase on v3 and drop that commit FYI. No need to start over in future :)

PR #446, opened 2020-09-11
Resize Header

frankyma:
Give a slight margin to top header so the scroll doesn't touch items.  The original  had a faux header with an empty helper text item but if a helper text did show it would almost touch.
Screen Shot 2020-09-11 at 1 05 41 PM


CarsonF (approved):
Thanks @frankyma!

PR #456, opened 2020-09-16
Bugfix for refocusing infinitely register form

jmc637:


jmc637:
Hi, this is for the bug reported in the register form when the fields autocomplete but then it infinitely refocuses between the fields that autocompleted at the same time. The issue is inside useFocus when is being used inside the TextField component. There's is a timeout when refocusing the field. I think what is happening is that at some point they all get focused close to the same time when autocomplete happens. The field refocuses and the component rerenders but then launches another timeout to do the same thing. Before this timeout finishes one of the other timeouts from the other fields finishes and relaunches that components focus function causing an infinite loop. I removed the timeout but I'm nervous about this because I don't know why there was a timeout in the first place. Do you know why this was here? I don't want to remove it if it's going to break something.

CarsonF:
Thanks for looking into this @jmc637!

The timeout was there because the focus wouldn't be applied without it. Specific times to check are on initial render and after a failed submission that has field errors. 

If you validate that the timeout is needed, maybe a solution would be to listen for a blur event and cancel the timeout?

glassblowerscat:
> The timeout was there because the focus wouldn't be applied without it. Specific times to check are on initial render and after a failed submission that has field errors.

@CarsonF I did some playing around on the login form with Julius' current branch, and it seemed to be working as I would expect. The Email field focuses on page load. If I submit an incorrect password, the password field focuses again after the failed submit.

Is there some more particular case I should check to make sure this is working as intended without the timeout?

CarsonF (approved):
I cannot find any regression in functionality from removing this delay 👍 
It could be that since we changed the way we are focusing fields with #417 the delay was rendered unnecessary.

Thanks for this @jmc637! Better late than never 😬

PR #457, opened 2020-09-16
Update file version list after adding new version

glassblowerscat:
When adding a new version of a Project File or Defined File, the list of file versions was not refreshing in the UI until after a page reload because we were relying on `refetchQueries`. But `refetchQueries` only refreshes "live" queries, meaning those currently being used in mounted components. Since the `FileVersions` component is not mounted when a new version is added, `refetchQueries` did not serve our purpose.

Instead, we now call the `update` method of all mutations that add new file versions. This creates an `updateCachedVersions` function that can be passed the Apollo `cache` instance provided by the `update` method, the new array of `FileVersion`s, and the `id` of the file whose versions need to be updated. We also update all relevant mutations to use this function.

### DIFF
--- a/src/components/files/FileActions/FileActions.graphql
+++ b/src/components/files/FileActions/FileActions.graphql
@@ -13,7 +13,11 @@ query FileVersions($id: ID!) {
 mutation CreateFileVersion($input: CreateFileVersionInput!) {
   createFileVersion(input: $input) {
     id
-    name
+    children {
+      items {
+        ...FileVersionItem
+      }
+    }
   }
 }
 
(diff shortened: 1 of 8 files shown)


CarsonF (commented):
`file.children.total` isn't updated in cache which renders the dividers weird
Screen Shot 2020-10-13 at 11 14 37 AM

glassblowerscat:
> `file.children.total` isn't updated in cache which renders the dividers weird

🤦‍♀️ Fixed

CarsonF (approved):
Sweet thanks @glassblowerscat!

PR #461, opened 2020-09-16
Update Budget Record loading state

frankyma:
Include the update loading in the `ProjectBudget.tsx` loading state

### DIFF
--- a/src/scenes/Projects/Budget/ProjectBudget.tsx
+++ b/src/scenes/Projects/Budget/ProjectBudget.tsx
@@ -2,7 +2,7 @@ import { Breadcrumbs, makeStyles, Typography } from '@material-ui/core';
 import { Skeleton } from '@material-ui/lab';
 import { sumBy } from 'lodash';
 import { Column, Components } from 'material-table';
-import React, { useMemo } from 'react';
+import React, { useMemo, useRef } from 'react';
 import { useParams } from 'react-router-dom';
 import { Breadcrumb } from '../../../components/Breadcrumb';
 import { useCurrencyFormatter } from '../../../components/Formatters/useCurrencyFormatter';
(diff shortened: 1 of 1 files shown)


CarsonF (commented):
This just shows the loading skeleton right? I would rather not hide all of the present data just because we are changing one piece.

frankyma:
It's two pieces, the budget line and the total.  When I was testing the update was really quick, do you expect his to be a big enough issue to warrant a specific loading state?

CarsonF:
> It's two pieces, the budget line and the total.

`loading` also goes to the table which I think makes the whole thing loading. Maybe I'm wrong though I haven't checked out the code.

> When I was testing the update was really quick, do you expect his to be a big enough issue to warrant a specific loading state?

Yeah just because the API is fast on your machine doesn't mean that's always going to be the case in every situation.

frankyma:
Changed to only load the row and the total.  I don't know how to add more rows to the table so haven't been able to test that yet.
ezgif com-gif-maker

frankyma:
There's a react warning that was already there 

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in MTableEditCell (created by WithTheme(MTableEditCell))
    in WithTheme(MTableEditCell) (created by MTableBodyRow)
```

I tried to do a little debug bug couldn't solve it.  Not sure if either anyone else have ideas or it's just part of the MUI table

glassblowerscat:
> Not sure if either anyone else have ideas or it's just part of the MUI table

This is a bug in `material-table`; I have an issue open for it

src/scenes/Projects/Budget/ProjectBudget.tsx
```diff
@@ -57,7 +58,12 @@ export const ProjectBudget = () => {
   const { data, loading, error } = useProjectBudgetQuery({
     variables: { id: projectId },
   });
-  const [updateBudgetRecord] = useUpdateProjectBudgetRecordMutation();
+  const [
+    updateBudgetRecord,
+    { loading: isUpdateLoading },
```
glassblowerscat, line 63: @frankyma Maybe we should play it safe and use the `onCompleted` method of the mutation to set `budgetRecordIdInUpdate.current` back to `null`; what do you think?

frankyma: Technically it shouldn't be necessary but if you insist on it I don't mind.  Lmk

glassblowerscat: > Technically it shouldn't be necessary but if you insist on it I don't mind. Lmk

Eh, you're right. Just being overly cautious

glassblowerscat (approved):

PR #490, opened 2020-09-18
Refactor session handling to use plain Apollo cache

### LINKED ISSUE #349: Return to from login doesn't work
The query parameter is ignored. async dragons.
image
https://recordit.co/uencovXBTS

glassblowerscat:
We were creating a local-state context in which to keep the `user` values returned on login or registration, which meant that we were using a combination of `useEffect` and `useLazyQuery` to fire the `session` query and load its values into state. This turned out to have some undiagnosed synchronicity issues that affected our login redirects.

Using the `update` method of the `login` or `register` mutations, we can instead just load the returned `user` values directly into the cache and then pull those values from cache henceforward. Here we simplify the `useSession` hook to do that, as well as adding a function to perform the cache update with the `user` values.

**Update:** This does **not** address the bug that causes the call to `client.resetStore()` to throw an error when the user logs out from the Project Files view. We **could** solve that by calling `client.clearStore()`, which doesn't refetch all live queries, but that seems like a hack. Instead, we should figure out why one query in that one scene is staying live when others are not.

Closes #349 #356

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,10 +1,9 @@
 import { addDecorator, addParameters } from '@storybook/react';
 import { withInfo } from '@storybook/addon-info';
 import { HashRouter } from 'react-router-dom';
-import React, { createElement, Fragment } from 'react';
+import { createElement, Fragment } from 'react';
 import { appProviders } from '../src/App';
 import { Nest } from '../src/components/Nest';
-import { SessionProvider } from '../src/components/Session';
 
 // Do hacking to show dates easier
 import '../src/util/CalenderDate';
@@ -12,13 +11,9 @@ import '../src/util/hacky-inspect-dates';
 
 addDecorator(withInfo);
 
-const storybookProviders = [
-  createElement(HashRouter),
-  ...appProviders,
-  createElement(SessionProvider, { user: null }),
-];
+const storybookProviders = [createElement(HashRouter), ...appProviders];
 
-addDecorator(story => {
+addDecorator((story) => {
   // render story with contexts provided here.
   // if story fn is called directly it won't be a child the providers
   // defined here and thus contexts will not be provided.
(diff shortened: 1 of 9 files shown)


CarsonF (commented):
Looks good @glassblowerscat

glassblowerscat:
>  Except that storybook was previously configured to not try to grab a session every time it does a preview.

Okay, so my experience with Storybook is limited, but it seems like we can probably just delete this line. There's no separate `SessionProvider` anymore; it's all just the `ApolloProvider`, and so Storybook will just automatically get an empty session if there's no authentication mutation sent. Am I missing anything here?

CarsonF:
Is that because the `Session` query is not called by the app providers now, but only called when the hook use executed? And the hook is never executed in storybook?

glassblowerscat:
> Is that because the `Session` query is not called by the app providers now, but only called when the hook use executed? And the hook is never executed in storybook?

Yes, the `Session` query is only called inside `useSession`. `useSession` **is** called inside the stories for some components, but it will just be returning `undefined` for the `session.user` values those components are looking for. So they will still behave as if there is no session, which is what (I believe) we're looking for. I re-ran `yarn storybook` with my current branch, and everything seemed to work the same.

CarsonF:
> `useSession` **is** called inside the stories for some components, but it will just be returning `undefined` for the `session.user` values those components are looking for.

Why would it return undefined without making a network request if there is no value in the apollo cache?

glassblowerscat:
> Why would it return undefined without making a network request if there is no value in the apollo cache?

Ah, I see what you're after. No, it would still make a request. Is that important to prevent? That might take some refactoring of this, and it seems a little odd to write the code around its impact on `storybook` performance

CarsonF:
I thought it was at the time haha. Now I think it's more important to have this functionality fixed in the app.

CarsonF (approved):
Thanks for diving in and simplifying this for us @glassblowerscat!!

PR #492, opened 2020-09-20
Fix Project Detail Card Size

### LINKED ISSUE #466: Project detail card size
The top row height does not need to expand to the size of the second row.

Screen Shot 2020-09-17 at 7 38 57 PM

glassblowerscat:
Closes #466 

We were trying to fix a problem where the Budget card's height did not match the Files card's height during loading. We had originally fixed this by wrapping all the cards in a `material-ui` `<Grid>` component, but that had the
side effect of making the top row of cards the same height as the bottom row, which was undesirable.

The real problem was that the Budget card lacked a `loading` prop. Supplying this fixed the original issue, and removing the `<Grid>` fixed the new issue.

While we were in here messing around, we updated the props for `FilesOverviewCard` and `BudgetOvervieCard` to extend those of `FieldOverviewCard`, which they are wrapping.

### DIFF
--- a/src/components/BudgetOverviewCard/BudgetOverviewCard.tsx
+++ b/src/components/BudgetOverviewCard/BudgetOverviewCard.tsx
@@ -1,18 +1,21 @@
 import { AccountBalance } from '@material-ui/icons';
 import * as React from 'react';
 import { FC } from 'react';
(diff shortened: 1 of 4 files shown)


CarsonF (approved):
Thanks @glassblowerscat!

PR #495, opened 2020-09-20
Filter Projects by "My Projects"

### LINKED ISSUE #468: Filter Projects by 'My Projects'
Add an option to the Project List filters for 'My Projects' that displays only the projects a user is a Team Member of.

glassblowerscat:
Closes #468 

Add `mine` field to `ProjectFilterOptions`

### DIFF
--- a/src/scenes/Projects/List/ProjectFilterOptions.tsx
+++ b/src/scenes/Projects/List/ProjectFilterOptions.tsx
@@ -42,6 +42,7 @@ export const useProjectFilters = makeQueryHandler({
       decoder(value)?.map((v) => upperFirst(v) as Sensitivity),
   }),
   onlyMultipleEngagements: withKey(withDefault(BooleanParam(), false), 'multi'),
+  mine: withDefault(BooleanParam(), true),
 });
 
 export const ProjectFilterOptions = () => {
@@ -83,6 +84,7 @@ export const ProjectFilterOptions = () => {
           </Tooltip>
         }
       />
+      <SwitchField name="mine" label="Only Show My Projects" />
     </>
   );
 };


CarsonF (changes requested):
I think we want to default to `mine`. Maybe we could change this to `all` / "Show All Projects"?

glassblowerscat:
> I think we want to default to `mine`. Maybe we could change this to `all` / "Show All Projects"?

From a UX perspective, I kind of think the better thing would be to just change the default. That way the text ("Only Show My Projects") explains why some projects aren't shown. I'm always a little annoyed when some things are hidden by default and the option to change that is labeled "Show all", because it doesn't tell me what is hidden

CarsonF:
I agree that "show all" isn't very indicative of the other option.

I wanted to replace the toggle switch here for that reason with [two toggle buttons](https://material-ui.com/components/toggle-button/#exclusive-selection). That way labels for either option could be represented. Kinda like this:


Screen Shot 2020-09-21 at 4 40 30 PM

I'm working on an `EnumField` that could be used for here for that.

glassblowerscat:
> I think we want to default to `mine`.

So, as far as I can tell, our `useProjectFilters` hook will always return an empty object from the default filters, which means there's no obvious way to use it to set a default filter that's different than the API default. We could do something like this:

```
const [filters, setFilters] = useProjectFilters();

const defaultMine = {
  ...filters,
...(filters.mine ? null : { mine: true }),
}

  const { data } = useProjectListQuery({
    variables: {
      input: {
        ...sort.value,
        filter: filters,
      },
    },
  });
```

But that feels a little gross. It seems more like if we want this to be the default we should change it on the API side. Thoughts, @CarsonF?

CarsonF:
Yes there is a mismatch here between the UI & the API, but the API mutation is called `projects` not `myProjects`. I think the fact that this is defaulted to my projects is a UI concern.

And yes I believe the default values should be left off of the query parameters. Ideally this change doesn't leave the `ProjectFilterOptions.tsx` file.

It's possible that this is enough:
```tsx
  mine: BooleanParam(true),
```
SwitchField might also need a default as well:
```tsx
<SwitchField name="mine" defaultValue={true} />
```

glassblowerscat:
> It's possible that this is enough:
> 
> ```
> export const useProjectFilters = makeQueryHandler({
>   mine: BooleanParam(true),
>   ...
> ```
> 
> `<SwitchField name="mine" defaultValue={true} />`

I made both of these changes in a recent commit, but the `filters` object is still empty. The `SwitchField` does mount in `true` position, but it has no effect on what gets queried

glassblowerscat:
@CarsonF this is ready to review again, unless you want to wait for the `EnumField` to be ready first

CarsonF (approved):
> @CarsonF this is ready to review again, unless you want to wait for the EnumField to be ready first

Nope, looks great! I just rebased and tweaked the commit message.

PR #506, opened 2020-09-22
Rename clusters to onlyMultipleEngagements

frankyma:


src/scenes/Projects/List/ProjectFilterOptions.tsx
```diff
@@ -48,7 +48,10 @@ export const ProjectFilterOptions = () => {
           ))}
         </Grid>
       </CheckboxesField>
-      <SwitchField name="clusters" label="Only Show Cluster Projects" />
+      <SwitchField
+        name="clusterOrCohort"
+        label="Only Show Cluster/Cohort Projects"
+      />
```
CarsonF, line 54: Can we add a tooltip here that says "Clusters/Cohorts are projects with multiple engagements"

CarsonF (approved):

PR #507, opened 2020-09-22
Feature/fe 374 add language details and edit fields

frankyma:
Show All the fields on detail scene that exists on edit.  Added the estimated end fiscal year, sign language, etc.  Skipped Location.

### DIFF
--- a/src/scenes/Languages/Detail/LanguageDetail.tsx
+++ b/src/scenes/Languages/Detail/LanguageDetail.tsx
@@ -153,15 +153,15 @@ export const LanguageDetail = () => {
           ) : (
             <DisplayProperty
               label="Ethnologue Code"
-              value={
-                ethnologue?.code.value ??
-                (ethnologue?.provisionalCode.value
-                  ? `${ethnologue.provisionalCode.value} (provisional)`
-                  : null)
-              }
+              value={ethnologue?.code.value}
               loading={!language}
             />
           )}
+          <DisplayProperty
+            label="Provisional Code"
+            value={ethnologue?.provisionalCode.value}
+            loading={!language}
+          />
           <DisplayProperty
             label="Registry of Dialects Code"
             value={registryOfDialectsCode?.value}
--- a/src/scenes/Languages/Detail/LanguageDetail.tsx
+++ b/src/scenes/Languages/Detail/LanguageDetail.tsx
@@ -154,9 +154,10 @@ export const LanguageDetail = () => {
             <DisplayProperty
               label="Ethnologue Code"
               value={
-                ethnologue?.code.value ?? ethnologue?.provisionalCode.value
+                ethnologue?.code.value ??
+                (ethnologue?.provisionalCode.value
                   ? `${ethnologue.provisionalCode.value} (provisional)`
-                  : null
+                  : null)
               }
               loading={!language}
             />
--- a/src/scenes/Languages/Detail/LanguageDetail.tsx
+++ b/src/scenes/Languages/Detail/LanguageDetail.tsx
@@ -115,6 +115,12 @@ export const LanguageDetail = () => {
               wrap={(node) => <Grid item>{node}</Grid>}
             />
             <LeastOfThese language={language} />
+            <BooleanProperty
+              label="Sign Language"
+              redacted="You do not have permission to view whether the language is a sign language"
+              data={language?.isSignLanguage}
+              wrap={(node) => <Grid item>{node}</Grid>}
+            />
           </Grid>
           <DisplayProperty
             label="Pronunciation Guide"
--- a/src/scenes/Languages/Detail/LanguageDetail.tsx
+++ b/src/scenes/Languages/Detail/LanguageDetail.tsx
@@ -1,11 +1,12 @@
 import { Grid, makeStyles, Tooltip, Typography } from '@material-ui/core';
-import { Add, Edit } from '@material-ui/icons';
+import { Add, DateRange, Edit } from '@material-ui/icons';
 import { Skeleton } from '@material-ui/lab';
 import clsx from 'clsx';
 import React from 'react';
 import { useParams } from 'react-router';
-import { canEditAny } from '../../../api';
+import { canEditAny, securedDateRange } from '../../../api';
 import { BooleanProperty } from '../../../components/BooleanProperty';
+import { DataButton } from '../../../components/DataButton';
 import { useDialog } from '../../../components/Dialog';
 import {
   DisplaySimpleProperty,
@@ -13,6 +14,7 @@ import {
 } from '../../../components/DisplaySimpleProperty';
 import { Fab } from '../../../components/Fab';
 import {
+  useDateFormatter,
   useDateTimeFormatter,
   useNumberFormatter,
 } from '../../../components/Formatters';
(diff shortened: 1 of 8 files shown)


glassblowerscat:
@CarsonF @frankyma This code is working, but the "Edit Language" form is **very** crowded now:

image

I think we're doing it this way because we're basically just sharing the same `LanguageForm` component between `CreateLanguage` and `EditLanguage`, which is good for developers, but maybe not so good for users who might just want to edit one or two fields in a `Language`. Should we (either now or with a subsequent ticket/PR) refactor this to be more like the `EditEngagement` form, and use more `DataButton`s in the Language Detail scene instead of `DisplayProperty`s?

CarsonF:
@prutherford Asked that we don't use multiple columns here either. I've asked @AmyRickwartz to look at this page and make it better, so we could probably wait until then to talk about this UX as well.

glassblowerscat:
> I've asked @AmyRickwartz to look at this page and make it better, so we could probably wait until then to talk about this UX as well.

Okay, that sounds like a new ticket to me; I'll go ahead and approve this so we can have the functionality, unless I hear any objection

CarsonF:
Can we add signLanguageCode here too? ([src](https://github.com/SeedCompany/cord-api-v3/blob/d3e277fd9ec8d56114b62206754c7189b7162aed/src/components/language/dto/language.dto.ts#L92))

src/scenes/Languages/Edit/EditLanguage.tsx
```diff
@@ -33,6 +33,9 @@ export const EditLanguage = (props: EditLanguageProps) => {
               registryOfDialectsCode: language.registryOfDialectsCode.value,
               leastOfThese: language.leastOfThese.value,
               leastOfTheseReason: language.leastOfTheseReason.value,
+              isSignLanguage: language.isSignLanguage.value,
```
glassblowerscat, line 36: Carson would like us to add `signLanguageCode` as well, if we can

glassblowerscat (changes requested):

frankyma:
@CarsonF should we show sign language code as a tooltip on the sign language button like on the Least Of These Partnership Button?  Or just on a displaySimpleProperty
Screen Shot 2020-09-25 at 11 12 26 AM

CarsonF:
> @CarsonF should we show sign language code as a tooltip on the sign language button like on the Least Of These Partnership Button? Or just on a displaySimpleProperty

I think it actually should replace ethnologue code. I believe they are mutually exclusive, but I would need to confirm with @sethmcknight.
While we are talking about that, the language could have both eth code and provisional code. So we need to change that logic to display both if given.

frankyma:
> While we are talking about that, the language could have both eth code and provisional code. So we need to change that logic to display both if given.

So not this if else logic? 
```<DisplayProperty
            label="Ethnologue Code"
            value={
              ethnologue?.code.value ?? ethnologue?.provisionalCode.value
                ? `${ethnologue.provisionalCode.value} (provisional)`
                : null
            }```
It seems like all three are tied together (Sign language code, ethnologue code, and provisional code)?  Let's make sure we're clear on exactly what to show here

**Comment by CarsonF** [comment]
Btw these sign language codes are restricted to two letters followed by two numbers ([src](https://github.com/SeedCompany/cord-api-v3/blob/7baa5519628c670d814f28f4a531c5194c5b854e/src/components/language/dto/update-language.dto.ts#L84)).
Could follow suit of `EthnologueCodeField` to enforce that here.

**Comment by CarsonF** [comment]
Yeah unfortunately Seth is off today so you'll have to wait until Monday if you want confirmation.

File: src/scenes/Languages/Edit/EditLanguage.tsx (hunk 1 above)
Comment (frankyma) [reply src/scenes/Languages/Edit/EditLanguage.tsx:41]: Added

**Comment by frankyma** [comment]
Added Sign language code on the language form, disabled this field unless `isSignLanguage` is selected.  On Language Detail, show sign language code if language is sign language and the code value exists, otherwise show ethnologue code

File: src/scenes/Languages/Detail/LanguageDetail.tsx [hunk 2]
```diff
@@ -104,45 +116,72 @@ export const LanguageDetail = () => {
...
+            <DisplayProperty
+              label="Sign Language Code"
+              value={signLanguageCode.value}
+              loading={!language}
+            />
+          ) : (
+            <DisplayProperty
+              label="Ethnologue Code"
+              value={
+                ethnologue?.code.value ?? ethnologue?.provisionalCode.value
```
glassblowerscat, line 157: Can we fix this while we're in here? If the `code` has a value but the `provisionalCode` doesn't, this will display "null":

image

src/scenes/Languages/LanguageForm/LanguageForm.tsx
```diff
@@ -48,212 +68,267 @@ export const LanguageForm = <T extends any>({
...
-                    </Grid>
-                  )}
-                </SecuredField>
-              </Grid>
+      {({
+        values,
+      }: {
+        values: Partial<LanguageFormValues<CreateLanguage | UpdateLanguage>>;
+      }) => {
+        console.log('values: ', values);
```
glassblowerscat, line 78:
```suggestion
```

glassblowerscat (changes requested):

src/scenes/Languages/Detail/LanguageDetail.tsx
CarsonF: While we are here we also want to display both the eth code **and** provisional code. I was mistaken previously to try to pick one or the other.

frankyma: good catch, fixed

src/scenes/Languages/LanguageForm/LanguageForm.tsx
frankyma: oops fixed.  About time we put in that eslint console log warning?

frankyma:
Just noticed we're missing delete language on the ticket, I'll add that to this PR

src/scenes/Languages/Detail/LanguageDetail.tsx
frankyma: @CarsonF didnt see your comment earlier.  To confirm Ethnologue Code is conditional based on 

> While we are here we also want to display both the eth code **and** provisional code. I was mistaken previously to try to pick one or the other.

@CarsonF didn't see this.  Just to confirm, Ethnologue Code has that conditional relationship with Sign Language Code as discussed, but Provisional Code is independent of the other two?

frankyma:
Pushed to always show provisional code and added delete langauge (although the refetchquery isn't working after the delete so navigating back to language list still shows the deleted language).

src/scenes/Languages/LanguageForm/LanguageForm.tsx
glassblowerscat: Maybe! It can't annoy me too much more than the one for `debugger`

glassblowerscat:
> the refetchquery isn't working after the delete so navigating back to language list still shows the deleted language

@frankyma Refetches will only work on live queries, meaning queries that are currently being used to render. This is also why you're getting yelled at about not providing an input; the client will only auto-provide variables for a refetch if the query is currently live with those variables.

You're going to have to use the `update` method of the mutation to manually seed the updated data into the cache. You can check out `Login` and `Register` components for examples of this.

CarsonF:
@glassblowerscat "live queries" means components that are mounted? The language details is still in view under the edit dialog...

glassblowerscat:
> @glassblowerscat "live queries" means components that are mounted? The language details is still in view under the edit dialog...

But the languages is list is not, and that's where Frank needs the UI refresh to occur

frankyma:
Moved out Delete Language for later.  Everything else should be working now.

glassblowerscat (approved):
🙌

PR #508, opened 2020-09-22
More query param handling

CarsonF:
This adds better handling of query parameters, including default values & mapping key/values to different strings for url.

## Default Values

Default values can now be specified with the `withDefault` wrapper function
```tsx
export const useProjectFilters = makeQueryHandler({
  onlyMultipleEngagements: withDefault(BooleanParam(), false),
});
```
The underlying library has this function as well but we are using a custom one now to:
- Remove default values from the url when using setter
- Supplies default values to the hooks current value even when omitted from the url

The types also now correctly account for the value being `undefined` when `withDefault` is not used.

## Different keys for url

Our query param keys are usually pretty specific due to whatever is best for code, and for syncing with API keys.
However, these are not always the best keys to use in urls.

The best example is `onlyMultipleEngagements` flag which just got renamed (formally "cluster").
This is really long for urls, so now we can map that to something shorter like `multi`.
https://github.com/SeedCompany/cord-field/blob/92dcead3797b4201a05d8163595c7338291ba85f/src/scenes/Projects/List/ProjectFilterOptions.tsx#L44

## Transforms

Along the same lines as keys, it would also be nice to be able to map values in the url.
I've added a `withTransform` wrapper function.

Previously `useSort` transformed values by referencing the underlying functions directly:
https://github.com/SeedCompany/cord-field/blob/a57b6e10ea80b54ea0ee9da90ef8de85b249c041/src/components/Sort/useSort.ts#L16-L19

Now with the decorator we have those functions given to us:
https://github.com/SeedCompany/cord-field/blob/3c2d8f4548c39c97ed9f3940b91840eac7619e40/src/components/Sort/useSort.ts#L15-L18

I'm not exactly thrilled with the API - I would've liked to make it simpler. However, we don't know if we should be calling the inner decoder/encoder before or after the transform function, so I just opted to give it to the transform function. This way it can decided whether to call it before or after the transformation. 

I also updated project filter [`status`](https://github.com/SeedCompany/cord-field/blob/92dcead3797b4201a05d8163595c7338291ba85f/src/scenes/Projects/List/ProjectFilterOptions.tsx#L23-L37) & [`sensitivity`](https://github.com/SeedCompany/cord-field/blob/92dcead3797b4201a05d8163595c7338291ba85f/src/scenes/Projects/List/ProjectFilterOptions.tsx#L38-L43) definitions to lowercase their values.

---

I also know there's that render error on status change due to an infinite loop. I opted to not solve that here.

### DIFF
--- a/src/components/Sort/useSort.ts
+++ b/src/components/Sort/useSort.ts
@@ -1,5 +1,4 @@
-import { decodeString, encodeString } from 'serialize-query-params';
-import { makeQueryHandler, StringParam } from '../../hooks';
+import { makeQueryHandler, StringParam, withTransform } from '../../hooks';
 import { SortValue } from './SortControl';
 
 export const useSort = <T>() => {
@@ -13,8 +12,8 @@ export const useSort = <T>() => {
 
 const useSortHandler = makeQueryHandler({
   sort: StringParam,
-  order: {
-    decode: (val) => decodeString(val)?.toUpperCase(),
-    encode: (val) => encodeString(val)?.toLowerCase(),
-  },
+  order: withTransform(StringParam, {
+    encode: (val, encoder) => encoder(val?.toLowerCase()),
+    decode: (val, decoder) => decoder(val)?.toUpperCase(),
+  }),
 });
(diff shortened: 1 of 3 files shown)


glassblowerscat:
> I also know there's that render error on status change due to an infinite loop. I opted to not solve that here.

This also happens when I change `sensitivity`

CarsonF:
Err I believe that's what I meant. I think it's a problem with the SwitchField. If I comment out one of the fields the other works.

glassblowerscat:
> Err I believe that's what I meant. I think it's a problem with the SwitchField. If I comment out one of the fields the other works.

Coolcool, just checking

glassblowerscat (approved):
💪

PR #514, opened 2020-09-24
Fix non-boolean `checked` value passed to Edit Engagement checkbox fields

glassblowerscat:
When creating the initial values for this form, we add `engagement.lukePartnership.value` `engagement.firstScripture.value`, even though those values might be `null`, and the checkbox field is expecting a boolean.

- Add double-! to value in `initialValues` object for `lukePartnership` and `firstScripture` fields to ensure that `null` values get passed to the form field as booleans

### DIFF
--- a/src/components/form/CheckboxField.tsx
+++ b/src/components/form/CheckboxField.tsx
@@ -62,7 +62,7 @@ export const CheckboxField: FC<CheckboxFieldProps> = ({
           <Checkbox
             {...rest}
             inputRef={ref}
-            checked={input.value}
+            checked={input.value || defaultValue}
             value={name}
             onChange={(e) => input.onChange(e.target.checked)}
             required={props.required}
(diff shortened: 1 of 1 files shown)


CarsonF (commented):
I believe this should be handled by the CheckboxField. That way it can't be bugged in other places in the future.

CarsonF:
I think changing https://github.com/SeedCompany/cord-field/blob/4325f2c401a88751f022c7818f9cfbba9851f757/src/components/form/CheckboxField.tsx#L65
to 
```tsx
checked={input.value ?? defaultValue}
```
should do the trick

glassblowerscat:
> ```tsx
> checked={input.value ?? defaultValue}
> ```

had to use `||` because `input.value` defaults to `""`, but yup!

CarsonF:
What if defaultValue is true and the current value is false?

glassblowerscat:
> What if defaultValue is true and the current value is false?

Aw, nerds, wasn't thinking

glassblowerscat:
@CarsonF Fixed again

CarsonF (approved):
LGTM

PR #517, opened 2020-09-24
Fix cropping of long language names in `LanguageListItemCard`

### LINKED ISSUE #501: Languages > Population field cropped on language listing.
**Describe the bug**
Languages > Population field cropped on language listing.

**Steps to Reproduce**
1. Login
2. Click on the languages menu and view the population field

**Expected Result**
Population field should be displayed properly

**Actual Result**
Population field cropped on language listing.
See screenshot -  
image

cord-field# b1ebad03

glassblowerscat:
Closes #501 

Add `overflow-wrap: break-word` and `word-break: break-all` CSS properties to h4 containing potentially(?) long language name

### DIFF
--- a/src/components/LanguageListItemCard/LanguageListItemCard.tsx
+++ b/src/components/LanguageListItemCard/LanguageListItemCard.tsx
@@ -14,9 +14,13 @@ const useStyles = makeStyles(({ spacing }) => {
       width: '100%',
       maxWidth: 400,
     },
-    cardContent: {
+    name: {
+      marginBottom: spacing(2),
+    },
+    bottomSection: {
       display: 'flex',
       justifyContent: 'space-between',
+      alignItems: 'flex-end',
     },
     leftContent: {
       flex: 1,
@@ -27,11 +31,6 @@ const useStyles = makeStyles(({ spacing }) => {
       flexDirection: 'column',
       justifyContent: 'flex-end',
       textAlign: 'right',
-    },
-    name: {
-      marginBottom: spacing(2),
-      overflowWrap: 'break-word',
-      wordBreak: 'break-all',
     },
   };
 });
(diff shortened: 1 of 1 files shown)


CarsonF (commented):
I'm hesitant to apply this and even to try to fix this "bug".

Here's the longest language name we have in v2 applied to v3 - "Quechua, Huamalíes-Northern Dos de Mayo Huánuco"

Screen Shot 2020-09-28 at 5 29 50 PM
(I re-ordered words in screenshot just to play around with wrapping)

It looks like it is already wrapping down on overflow breaking on "words".
So `overflowWrap` changes nothing.

I really doubt we want to break on any letters (`wordBreak`) - this is almost never a good UX.
Screen Shot 2020-09-28 at 5 36 15 PM
"Dos de Ma yo Huánuco"?

----

I didn't get a chance to add population to see what that looked like. My local env is indisposed with API auth stuff. 
If population gets pushed off the card with the language name I mentioned here then that needs to be fixed, but probably a different way.

glassblowerscat:
> I'm hesitant to apply this and even to try to fix this "bug".

I'd totally be with you if you wanted to just close this ticket at a "won't fix". I think you're right that it doesn't seem like a real problem we'll actually encounter.

CarsonF:
Yeah I think I will. @glassblowerscat would you mind real quick confirming that that language name above "works" (doesn't cut off any text) in our current card with a population value.

CarsonF (approved):
I took a different approach and moved the name above and out of the two columns. Now it spans the full width and looks better in both the bugged case and a real case:

Before: 
Before

After:
After

@glassblowerscat this good with you?

glassblowerscat:
> @glassblowerscat this good with you?

Yup!

PR #519, opened 2020-09-24
Default budget records sorting

### LINKED ISSUE #472: Default Budget Records Sorting
Default the Budget Records sorting by Funding Partner (Ascending) then Fiscal Year (Ascending). Also, change the column header from 'Organization' to 'Funding Partner'.

For example, if Wycliffe and Seed Company are funding a Project with Fiscal Years 2020-2022 the records should default sort as:

Seed Company | 2020 | $$$
Wycliffe             | 2020 | $$$
Seed Company | 2021 | $$$
Wycliffe             | 2021 | $$$
Seed Company | 2022 | $$$
Wycliffe             | 2022 | $$$

glassblowerscat:
Closes #472 

`material-table` doesn't offer much in terms of default sorting of data; it expects the consumer to pass in the data pre-sorted into the desired default. So we do that here, and also add `options.thirdSortClick: false` to the props we pass to our `Table` component. This is because the normal `material-table` sort button click is as follows:

First click: Sort column ascending
Second click: Sort column descending
Third click: Return column to original unsorted state

We're disabling that third click because the original "unsorted" state for two of our three columns will actually be sorted, so leaving the third click enabled would make it appear to the user as if the behavior were:

First click: Sort column ascending
Second click: Sort column descending
Third click: Do nothing

### DIFF
--- a/src/scenes/Projects/Budget/ProjectBudgetRecords.tsx
+++ b/src/scenes/Projects/Budget/ProjectBudgetRecords.tsx
@@ -59,6 +59,7 @@ export const ProjectBudgetRecords: FC<ProjectBudgetRecordsProps> = (props) => {
       {
         field: 'fiscalYear',
         editable: 'never',
+        defaultSort: 'asc',
       },
       {
         field: 'amount',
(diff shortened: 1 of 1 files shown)


CarsonF (approved):
💯

PR #520, opened 2020-09-28
95 infinite scroll virtualized lists

### LINKED ISSUE #95: Infinite Scroll & Virtualized Lists

glassblowerscat:
Closes #95 

## Library

We had originally started work on this using the third-party library `react-virtualized`, but ran into trouble. The author of this library actually recommends users to consider avoiding it in favor of his newer library `react-window` (plus additional libraries that provide auto-sizing and infinite scroll). This worked well, except for the fact that it required all children of the virtualized list to be of uniform height. Our items are rarely if ever of uniform height, so I looked around for alternatives.

What we are using currently, then, is `react-infinite-scroll-component`. This is not as full-featured as either of the above options, and it lacks any windowing capability. But it was so simple to implement and performed so well that I thought it was worth using.

Windowing is therefore implemented using `refs` to measure the height of the top-level element returned by `react-infinite-scroll-component`, then return that value up to the consumer of the `VirtualList` component to use in calculating the number of items per page that should be queried, based on the rough or average height of the children. We use the height of the window as a fallback in case we have trouble getting the height of the container. I went conservative on the number of items to do per page: rounded up + 1. Might not need that +1; thoughts?

I also discovered that we don't need any "auto-sizer" component due to the way our elements/CSS are already laid out. What we **do** need to do is set the height of the `ListContainer` that encloses the `VirtualList` to `100%`. So I've added a `className` prop to `ListContainer ` to allow consumers of it to do this. It's possible that we ought to just bake this into `ListContainer `; ~~I'd be interested in opinions on this.~~ **Update:** I did that thing

## Data

I started out using the Apollo query `fetchMore.update()` method to push new data into the cache, but it turns out this is now deprecated and will be removed in the next version of the client. The recommendation is to instead set field policies on the individual fields, which has the added benefit of being a single-time write. So our `typePolicies.base` file has been updated with policies for `Query.projects`.

## ARIA

Carson identified that we ought to use `role='feed'` and the associated ARIA settings with the Virtual List to make it friendly to keyboard navigation/screen readers. This turned out to include making sure that the children of the virtual list are either `<article>` elements or elements with `role='article'`. In this PR I've only updated the `ProjectList` to use `VirtualList`, so I've also only updated the `ProjectListItemCard` to comply with this spec. I opted to just change the element returned by the `material-ui` `<Card>` component to `<article>` instead of using `role='article'`.


src/api/typePolicies/typePolicies.base.ts
```diff
@@ -1,3 +1,35 @@
...
+            ...existingProjects.slice(endNew),
+          ];
+          return {
+            ...incoming,
+            items: updatedProjects,
+          };
+        },
+        read(existing: ProjectListOutput | undefined) {
+          return existing;
+        },
```
CarsonF, line 31: What's this read do?

src/api/typePolicies/typePolicies.base.ts
```diff
@@ -1,3 +1,35 @@
...
+          const updatedProjects = [
+            ...existingProjects.slice(0, startNew),
+            ...incoming.items,
+            ...existingProjects.slice(endNew),
+          ];
+          return {
+            ...incoming,
+            items: updatedProjects,
+          };
+        },
```
CarsonF, line 28: Can we abstract this? We'll need it for the other lists as well.

Also if any of the input changes besides count or page we'll need to reset the cache.

src/components/Layout/ListContainer.tsx
```diff
@@ -9,7 +9,18 @@ const useStyles = makeStyles(({ spacing }) => ({
...
+}
+
+export const ListContainer: FC<ListContainerProps> = ({
+  className,
+  children,
+}) => {
   const classes = useStyles();
-  return <div className={classes.root}>{children}</div>;
+  return (
+    <div className={`${classes.root}${className ? ` ${className}` : ''}`}>
```
CarsonF, line 22:
```suggestion
    <div className={clsx(classes.root, className)}>
```

src/components/VirtualList/useItemsPerPage.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react';
+
+export function calculateItemsPerPage(
+  containerHeight: number,
+  itemHeight: number
+) {
+  // Throwing on an extra `1` just to play it safe
+  return Math.ceil(containerHeight / itemHeight) + 1;
```
CarsonF, line 8: react-virtualized has a row overscan prop that allows extra rows off screen, and a prop to start fetching next page when X rows close to the bottom. This allowed the buffer to be configured based on usage to help make the loading more seamless.

src/scenes/Projects/List/ProjectList.tsx
```diff
@@ -64,14 +86,24 @@ export const ProjectList: FC = () => {
...
+        <VirtualList
+          dataLength={currentItemsCount}
+          next={loadMoreItems}
+          hasMore={hasMore}
+          itemType="project"
+          setContainerHeight={setContainerHeight}
+        >
+          {listOrPlaceholders(data?.projects.items, itemsPerPage).map(
+            (item, index) => (
+              <ProjectListItemCard
```
CarsonF, line 99: Could `VirtualList` use a render prop (or children function) to render individual rows? That would allow rendering to be decoupled from list pagination logic.

src/scenes/Projects/List/ProjectList.tsx
```diff
@@ -20,25 +21,46 @@ const useStyles = makeStyles(({ spacing }) => ({
...
+      variables: {
+        input: {
+          ...sort.value,
+          filter: filters,
+          count: itemsPerPage,
+          page: Math.floor(currentItemsCount / itemsPerPage + 1),
+        },
+      },
+    });
+  };
```
CarsonF, line 63: All of this could be put into a `useInfiniteList` hook that returns props specifically matching what `VirtualList` is expecting.

src/components/VirtualList/VirtualList.tsx
```diff
@@ -0,0 +1,122 @@
...
+      ref={containerRef}
+    >
+      <InfiniteScroll
+        dataLength={dataLength}
+        next={next}
+        hasMore={hasMore}
+        loader={
+          <div className={classes.loading} style={statusStyle}>
+            <CircularProgress size={20} />
+          </div>
```
CarsonF, line 107: What do you think about letting children specify loading. We already have a loading skeletons for all of these cards. Isn't it a little weird for the initial load to use skeletons and subsequent loads to use a spinner ?

src/components/VirtualList/VirtualList.tsx
```diff
@@ -0,0 +1,122 @@
...
+        }
+        endMessage={
+          <div className={classes.endMessageContainer} style={statusStyle}>
+            <Typography variant="body1" className={classes.endMessage}>
+              All {itemType ? `${itemType}s` : 'items'} retrieved
+            </Typography>
```
CarsonF, line 113: While this generally works, there are exceptions. educations, unavailabilitys.

I wonder if it's sufficient to just show nothing? "Here's the bottom of the list. There's no loading so there's not more."

CarsonF (commented):
Some initial notes/thoughts/concerns. Haven't pulled the code down yet.

I can also already hear, "I click on a project and then I go back and the list resets" because we don't maintain the page/scroll state.

src/api/typePolicies/typePolicies.base.ts
glassblowerscat: I would have thought nothing, so at first I left it out. It then returned no data. Seems almost like a bug

src/scenes/Projects/List/ProjectList.tsx
glassblowerscat: I mean, I'm sure it could. I'm struggling to understand the benefit, though... maybe because it's getting late and I'm tired. The consuming component still has to pass in basically all the same information, except with a layer of abstraction now inserted. To me, render props are for when you need more than one "children" slot, or when you have to do some kind of extra logic to the `children` before you render them. Neither seems to really be the case here.

src/components/VirtualList/VirtualList.tsx
glassblowerscat: Updated to a dot instead of text

src/components/VirtualList/useItemsPerPage.ts
glassblowerscat: Derp, it turns out `react-infinite-scroll-component` has this as well, and it defaults to 80% of the height of the container. So we probably don't need the extra element.

src/api/typePolicies/typePolicies.base.ts
glassblowerscat: Hm, I just tried it again, and it worked fine without it. Maybe a fluke before?

CarsonF:
Closing for now. Will revisit when infinite scroll becomes a priority.

PR #533, opened 2020-09-30
Product full testament and full book

frankyma:
Add ability to set full testament and full book ranges with a toggle field.

Moving alot of the helpers to src/util/biblejs/reference.ts.

The product form is even more algorithmically heavy.  I thought I saw some UI stutters one day but that might be because I had alot running on my machine. Please if you see any lags on your browser.


glassblowerscat:
I'm not familiar with this aspect of the product, so I don't really understand how this feature is supposed to work... can you tell me what I should be expecting when I use this form, @frankyma?

glassblowerscat (approved):

PR #534, opened 2020-10-01
Separate project engagements query and loading

frankyma:
Separate the loading states for project detail and engagement list with the same query fragment as before.  

ezgif com-gif-maker (3)

There's an imperfect behavior with the population line.  Just to confirm this is supposed to evaluate to undefined when the populations add up to 0? https://github.com/SeedCompany/cord-field/pull/534/files#diff-b6f8a4fabe48e786a616409ff388e241L116

If so then then on non zero population the page below will shift down when this number gets loaded.  We can put a loading skeleton there, but that means on 0 population (on all internship projects) the page will shift up.

What we can do is if it's an internship project never show this line.  If Translation project then always show a loader or the population total, even if it adds to 0.

### DIFF
--- a/src/scenes/Projects/Overview/ProjectOverview.tsx
+++ b/src/scenes/Projects/Overview/ProjectOverview.tsx
@@ -115,14 +115,13 @@ export const ProjectOverview: FC = () => {
       : 'Intern'
     : null;
 
-  const populationTotal =
-    engagementListData?.project.engagements.items.reduce(
-      (total, item) =>
-        item.__typename === 'LanguageEngagement'
-          ? total + (item.language.value?.population.value ?? 0)
-          : total,
-      0
-    ) || undefined;
+  const populationTotal = engagementListData?.project.engagements.items.reduce(
+    (total, item) =>
+      item.__typename === 'LanguageEngagement'
+        ? total + (item.language.value?.population.value ?? 0)
+        : total,
+    0
+  );
 
   const date = projectOverviewData
     ? securedDateRange(
@@ -205,27 +204,20 @@ export const ProjectOverview: FC = () => {
               />
             </Grid>
           </Grid>
-          <DisplaySimpleProperty
-            label="Population Total"
-            value={formatNumber(populationTotal)}
-            loadingWidth={100}
-            LabelProps={{ color: 'textSecondary' }}
-            ValueProps={{ color: 'textPrimary' }}
-            wrap={(node) => (
-              <Grid item>
-                <Tooltip
-                  title={
-                    projectOverviewData
-                      ? 'Total population of all languages engaged'
-                      : ''
-                  }
-                >
-                  {node}
-                </Tooltip>
+          {projectOverviewData?.project.__typename === 'TranslationProject' && (
+            <Tooltip title={'Total population of all languages engaged'}>
+              <Grid item xs={3}>
+                <DisplaySimpleProperty
+                  loading={!engagementListData}
+                  label="Population Total"
+                  value={formatNumber(populationTotal)} // formats to string
+                  loadingWidth={100}
+                  LabelProps={{ color: 'textSecondary' }}
+                  ValueProps={{ color: 'textPrimary' }}
+                />
               </Grid>
-            )}
-          />
-
+            </Tooltip>
+          )}
           <Grid container spacing={1} alignItems="center">
             <Grid item>
               <DataButton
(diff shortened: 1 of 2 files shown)


src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -100,20 +103,28 @@ export const ProjectOverview: FC = () => {
...
     : null;
 
-  const populationTotal =
-    data?.project.engagements.items.reduce(
-      (total, item) =>
-        item.__typename === 'LanguageEngagement'
-          ? total + (item.language.value?.population.value ?? 0)
-          : total,
-      0
-    ) || undefined;
```
frankyma, line 116: The old reduce would have the 0 value evaluated to undefined, which will not show that on the UI at all, was that on purpose?  I thought it was a mistake so I removed `|| undefined`

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -100,14 +103,20 @@ export const ProjectOverview: FC = () => {
     },
   });
 
+  const { data: engagementListData } = useProjectEngagementListOverviewQuery({
+    variables: {
+      input: projectId,
+    },
+  });
```
glassblowerscat, line 110: What's to stop this from starting to run while the main query is still fetching?

glassblowerscat:
> What we can do is if it's an internship project never show this line. If Translation project then always show a loader or the population total, even if it adds to 0.

I'd definitely be in favor of this. We're doing something similar elsewhere with a property I can't recall at the moment. But if something doesn't exist on Internship Engagements, we should definitely logic it out of here. As for whether we should default to `0` on Language Engagements to avoid the UI jump, maybe @CarsonF or @sethmcknight would like to weigh in?

src/scenes/Projects/Overview/ProjectOverview.tsx
frankyma: There's nothing stopping it, it's an independent query that runs automatically on component mount.  Is the concern here that this query finishes before the project details?  If so we can render the engagement list if and only if the project details main query finishes also

glassblowerscat: Well, I guess you would know more since you work on the API side a lot more than I do: Is having the db/API work on two queries at once going to make the first query more likely to lag? If not, then it doesn't matter

frankyma: I would guess it's treated like separate requests? Like a rest api?? @CarsonF @michaelmarshall let me know if I'm off base with this assumption

But I did need to only load the engagement cards after the above data has loaded so I added that check.

sethmcknight:
> > What we can do is if it's an internship project never show this line. If Translation project then always show a loader or the population total, even if it adds to 0.
> 
> I'd definitely be in favor of this. We're doing something similar elsewhere with a property I can't recall at the moment. But if something doesn't exist on Internship Engagements, we should definitely logic it out of here. As for whether we should default to `0` on Language Engagements to avoid the UI jump, maybe @CarsonF or @sethmcknight would like to weigh in?

We should always show the Population Total for Language Engagements. Though I can't imagine it happening with real data, if the Population Total is 0 then we should show that. It would be important information to know.

A though on Internship Projects... I would leave the space in for now. I'm going to create an issue that replaces the population total with a Total Interns count.

frankyma:
Updated to show population total with loader always on Language Engagement
ezgif com-gif-maker (4)

frankyma:
Added total interns on intern engagement

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -218,6 +218,20 @@ export const ProjectOverview: FC = () => {
               </Grid>
             </Tooltip>
           )}
+          {projectOverviewData?.project.__typename === 'InternshipProject' && (
+            <Grid item>
+              <DisplaySimpleProperty
+                loading={!engagementListData}
+                label="Total Interns"
+                value={formatNumber(
+                  engagementListData?.project.engagements.total
```
glassblowerscat, line 227: Is it possible for a project to have both Language and Internship Engagements? If so, this total might be incorrect

frankyma: Pretty sure a translation project only has language engagements and internship project only has internship engagements.  We've also enforced that creation on the client

glassblowerscat (approved):

PR #540, opened 2020-10-02
Enum Field

CarsonF:
This replaces `CheckboxesField`, `RadioField`, and `ToggleButtonField` with a single `EnumField`.
This field should be useful for all places we need one or more enum options chosen.

## Variants
Currently there are 4 _variants_ to pick from: `checkbox`, `radio`, `toggle-split`, `toggle-grouped`

checkbox
radio
toggle-split
toggle-grouped

## Multiple / Single Selection 
A `multiple` boolean prop controls whether multiple options can be selected at once.

## Layout

A `layout` prop can switch the options between being laid out in a `row`, a `column`, or `two-column`s.

row
column
two-column

## Option List

The recommended way for options to be specified is with the `options` prop.

```tsx
<EnumField options={['red', 'blue', 'green']} />
```

A label function can be applied to convert the values to what's shown in the UI
```tsx
<EnumField options={['red', 'blue', 'green']} getLabel={startCase} />
```

This works well with our generated enum lists and saves us the array-map-render boilerplate in a lot of cases.

## Custom Options

In the cases where a custom layout is needed, options can be defined as children allowing different layouts or sections/groups.

```tsx
<EnumField>
  <EnumOption default label="None" />
  <EnumOption value="red" label="Red" />
  <EnumOption value="blue" label="Blue" />
  <EnumOption value="green" label="Green" />
</EnumField>
```
**Question**: Would you rather label be specified via `children`?
```tsx
<EnumOption value="red">Red</EnumOption>
```

## Default Value

The `defaultValue` can be specified in order to give a different starting value (assuming there's not an _initial value_ from the form).
If a `defaultValue` is not specified it falls back to `null` for single and `[]` for multiple.

```tsx
<EnumField options={['red', 'blue']} defaultValue="red" />
```
Here _red_ will be selected first.

### Note on `required` prop
Also, if `required` prop is `false` then unselecting _blue_ will re-select _red_.

If `required` prop is `true` then neither option can be un-selected.

## Default Option

A default option can also be specified.
The value of this option is the `defaultValue` so this is typically used to show the "unselected" state.

```tsx
<EnumField options={[...]} defaultOption="None" />
<EnumField>
  <EnumOption default label="None" />
</EnumField>
```
Usage of this is shown in most of the screenshots here.

This should probably only be used if the default value is not an actual option. Otherwise the interaction would be silly. For example:
```tsx
<EnumField options={['red', 'blue']} defaultValue="blue" defaultOption="None" />
```

## Usage in App

I updated the codebase to use this new component, but I could pull that out into a separate PR if desired.

- [`ProjectFilterOptions`](https://github.com/SeedCompany/cord-field/commit/db99cd0)
- [`PartnershipForm`](https://github.com/SeedCompany/cord-field/commit/cae069d)
- [`EditEngagementDialog`](https://github.com/SeedCompany/cord-field/commit/139eae9)
- [`ProductForm`](https://github.com/SeedCompany/cord-field/commit/020e77a)
- [`CreateProjectForm`](https://github.com/SeedCompany/cord-field/commit/5464282)

## Questions / Notes / Etc
- I tried adding a `select` variant but it was proving difficult so I left it out (for now at least)
- I thought about renaming `EnumOption.label` to `EnumOption.children`.
  The select variant would require this since it uses reflection / element cloning.
- I think `EnumField.getLabel` should be renamed to `getOptionLabel`


src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -38,97 +36,60 @@ export const PartnershipForm = <
...
+    {({ values }) => (
+      <>
+        <SubmitError />
+        {!partnership && <OrganizationField name="organizationId" required />}
         <SecuredField obj={partnership} name="types">
           {(props) => (
-            <CheckboxesField label="Types" row {...props}>
-              {typesCheckboxes}
-            </CheckboxesField>
+            <EnumField
```
glassblowerscat, line 47: When I visit `/projects/:projectId/partnerships` for a project with partners, I get a crash and this error in the browser console:

```
PartnershipForm.tsx:47 Uncaught TypeError: Cannot read property 'types' of undefined
    at PartnershipForm.tsx:47
    at DialogForm.tsx:167
    at renderComponent (react-final-form.es.js:32)
    at ReactFinalForm (react-final-form.es.js:281)
```

CarsonF: Herp derp. Thanks for catching that!

glassblowerscat (dismissed):
Look pretty good to me!

glassblowerscat (approved):

PR #541, opened 2020-10-02
Update location api

glassblowerscat:
A recent refactor of the API made substantial changes to "locations" (using the term loosely to describe several different classes revolving around location). This updates our frontend GraphQL and resulting data to match (hopefully.... assuming the refactor was understood correctly 😬).

P.S. Since I was in messing around with our fragments, I centralized and de-duplicated a couple things with new .graphql files in `src/api/fragments`


src/components/form/Lookup/Location/LocationFields.tsx
```diff
@@ -1,32 +1,10 @@
...
-export const ZoneField = LookupField.createFor<ZoneLookupItem>({
-  resource: 'Zone',
-  useLookup: useZoneLookupLazyQuery,
-  label: 'Zone',
-  placeholder: 'Search for a zone by name',
+export const LocationField = LookupField.createFor<LocationLookupItem>({
+  resource: 'Location',
+  useLookup: useLocationLookupLazyQuery,
+  label: 'Location',
+  placeholder: 'Search for a location by name',
```
CarsonF, line 9: We'll still need the Region & Zone Fields. Probably just prefixed with "Field" now. Feel free to further split them out into other files since they are more separated now.

src/scenes/SearchResults/Search.graphql
```diff
@@ -11,9 +11,47 @@ fragment SearchResultItem on SearchResult {
...
+  ...LocationSearchResultItem
   ...FilmLookupItem
   ...StoryLookupItem
   ...LiteracyMaterialLookupItem
   ...SongLookupItem
 }
+
+# We have to inline this because the fragment we would usually
+# insert here has a `type` field that conflicts with a field of the
+# same name on the `Project` type, so we need to use an alias.
```
CarsonF, line 23: Is this a TS error or gql gen error?

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -236,10 +235,12 @@ export const ProjectOverview: FC = () => {
...
               <DataButton
                 loading={!projectOverviewData}
-                secured={projectOverviewData?.project.location}
+                secured={projectOverviewData?.project.primaryLocation}
                 empty="Enter Location"
                 redacted="You do not have permission to view location"
-                children={displayLocation}
+                children={
+                  projectOverviewData?.project.primaryLocation.value?.name.value
+                }
```
CarsonF, line 243: We'll want to display the field region here as well, but this can wait until we do the update functionality.

CarsonF (changes requested):
Thanks for diving in and updating this with little info from us!

src/scenes/SearchResults/Search.graphql
glassblowerscat: gql-gen error

src/api/fragments/secured.graphql
```diff
@@ -0,0 +1,17 @@
+fragment securedString on SecuredString {
```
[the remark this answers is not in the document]
glassblowerscat: So, the GraphQL convention is actually camelCase for fragments. Do we care about that?

src/components/form/Lookup/Location/LocationFields.tsx
glassblowerscat: I was going to leave these in, but the queries wouldn't pass codegen because `Query.search` won't accept `FieldZone` or `FieldRegion` as a valid `input.type`. Seems like that was maybe missed in the API update?

src/api/fragments/secured.graphql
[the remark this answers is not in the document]
CarsonF: Bahhhh. Yes we do lol. That does leave us inconsistent, but I'll live with it.

src/components/form/Lookup/Location/LocationFields.tsx
CarsonF: Ok we can leave them out for now. https://github.com/SeedCompany/cord-api-v3/issues/1295

src/scenes/SearchResults/Search.graphql
CarsonF: Wait I don't think we actually need type in the search results do we? If we are showing them on the card I guess we do.
I'm ok leaving this for now then as long as gql gen gets that rename into the TS types. It does right?

glassblowerscat: Yeppers

glassblowerscat:
@CarsonF All your requested changes are in

CarsonF (approved):
Thanks @glassblowerscat!!

PR #557, opened 2020-10-05
Newly created engagement doesn't show

frankyma:
Recently we separated out the engagement list query on the project overview so need to reference that on the refetch after create engagement


CarsonF (commented):
I think we need to keep the other one as well. Adding an engagement can change other project properties like sensitivity.

frankyma:
good point, pushed

CarsonF (approved):
🎉

PR #558, opened 2020-10-06
Prevent stale search results

### LINKED ISSUE #546: [BUG] Searching by a term that has already been searched for only returns the cached search results.
- Search for a term (ex. Oct)
- Load Search Results
- Create a new item that includes that term (ex. Oct 5th Project)
- Search for term again (ex. Oct)

Expected Results: Newly created item will be included in search results

Actual Results: Only the original results are returned.

glassblowerscat:
Closes #546 

- Add `fetchPolicy: 'cache-and-network'` to search query to make sure that results added since last fetch are pulled in
- Add logic to reload the search page if the user searches for the exact same thing again

I don't love this page reload, but `makeQueryParams` is memoized so as not to re-render unless the query changes. It seems like performing a search, then performing the exact same search again without navigating away from the search page, will probably be kind of an edge case, so it's probably okay.

### DIFF
--- a/src/scenes/Root/Header/HeaderSearch/HeaderSearch.tsx
+++ b/src/scenes/Root/Header/HeaderSearch/HeaderSearch.tsx
@@ -3,7 +3,7 @@ import { Search } from '@material-ui/icons';
 import { FC } from 'react';
 import * as React from 'react';
 import { Form } from 'react-final-form';
-import { useNavigate } from 'react-router-dom';
+import { useLocation, useNavigate } from 'react-router-dom';
 import { TextField } from '../../../../components/form';
 import { makeQueryHandler, StringParam } from '../../../../hooks';
 
@@ -25,6 +25,8 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
 export const HeaderSearch: FC = () => {
   const classes = useStyles();
   const [{ q: search = '' }] = useSearch();
+  const { pathname, search: routeSearch } = useLocation();
+  const currentSearch = routeSearch.split('?q=')[1];
   const navigate = useNavigate();
 
   return (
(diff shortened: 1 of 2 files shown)


CarsonF:
What if the change is made by someone else or just another tab, same problem. Cache invalidation is hard.

glassblowerscat:
> What if the change is made by someone else or just another tab, same problem. Cache invalidation is hard.

`cache-and-network` covers both of these scenarios

src/scenes/Root/Header/HeaderSearch/HeaderSearch.tsx
```diff
@@ -25,14 +25,21 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
...
   const navigate = useNavigate();
 
   return (
     <Form
       initialValues={{ search }}
       onSubmit={({ search }) => {
         if (search) {
-          navigate(`/search?q=${search}`);
+          // cover cases where the user searches for the exact same thing again
+          if (pathname === '/search' && search === currentSearch) {
```
CarsonF, line 38: I think `search` (from `useSearch`) will always equal `currentSearch`. The variable is just shadowed here in the submit handler. 

Would this work?
```tsx
onSubmit={({ search: newSearch }) => {
  if (newSearch === search) {
})
```

CarsonF:
> > What if the change is made by someone else or just another tab, same problem. Cache invalidation is hard.
> 
> `cache-and-network` covers both of these scenarios

You are absolutely right.


Since we just talked about this in slack. What do you want to do here?

glassblowerscat:
> Since we just talked about this in slack, what do you want to do here?

Kinda makes this ticket feel like just another symptom of that problem we discussed. If we implement `update` after "creation" mutations across the board, this will solve itself, unless I'm missing something.

CarsonF:
> If we implement `update` after "creation" mutations across the board, this will solve itself, unless I'm missing something.

I like solving the root problem 👍

glassblowerscat:
We agree that given our newly determined approach to invalidating the cache after "creation" actions, the issue this PR was intended to address is symptomatic. Closing this in favor of a different change entirely

PR #561, opened 2020-10-06
Redacted engagement list width

frankyma:


CarsonF (approved):
Hmm I don't particularly like the hardcoded width for this but I do see we are using that for the skeleton as well. I guess it's fine for now.

PR #568, opened 2020-10-08
Fixup language engagement card

frankyma:


CarsonF:
Sorry forgot to comment on this. Can you add number formatting to poluation?

frankyma:
> Sorry forgot to comment on this. Can you add number formatting to population?

PRed: https://github.com/SeedCompany/cord-field/pull/571

@CarsonF can you explain real quick what the number formatter does?

CarsonF:
> @CarsonF can you explain real quick what the number formatter does?

It...formats numbers lol. Think Excel. It has thousands separator (really the only thing we are looking for here), decimal places (min and max), prefix or type for currencies, etc. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

frankyma:
Ah ok.  Also I thought I saw somewhere that we depended on the returned string value for a truthy eval?  So since 0 becomes "0" it still shows

CarsonF:
I think that's ok. 0 is a number input by a user - it's not the lack of a number. The hook fn does return empty string for null values.

PR #574, opened 2020-10-12
Project workflow dialog

CarsonF:
Updated project status/step usage to use new workflow API in https://github.com/SeedCompany/cord-api-v3/pull/1340

Screen Shot 2020-10-12 at 4 34 29 PM


glassblowerscat:
@CarsonF Tried to "Submit for Concept Approval":
image

I also got toast message I foolishly dismissed. The Dialog then stayed open, but after a page refresh the step had changed to "Pending Concept Approval"

Then I tried to "Approve Concept" and got what I think was probably the same toast error:
image

This pattern persisted; I could choose a step, and I'd get an error, but it would still take and be visible after a page reload.

glassblowerscat:
Actually, here's a different error:
image

glassblowerscat:
It seems like "Resubmit for Concept Approval" ought to be the other button style here....

glassblowerscat:
image

glassblowerscat:
Seems like maybe more "Will Not Terminate"s than we need:
image

CarsonF:
Ah I should've clarified. The root user had an incorrect timezone that we are now using in email templates. We fixed it but the root user needs to be updated.

> Could not find step for given status

Yeah this seems legit. I'll check the mapping on the API.

> Resubmit for Concept Approval

You think this should be _Neutral_ instead of _Approve_?. It seemed like a approve action...

> Seems like maybe more "Will Not Terminate"s than we need

The duplicates are because this next step is actually like a "back" action. Currently we coded it as all the steps it could go back to. We need to do it a better way, this was just a first step.
I said I would put a tooltip on these buttons displaying the step it would go to. I forgot - I'll do that now.

glassblowerscat:
> You think this should be _Neutral_ instead of _Approve_?. It seemed like a approve action...

I'll take your word for it; I don't really know the business logic here. It was just the only one of these that had more than one _Approve_

CarsonF:
> > You think this should be _Neutral_ instead of _Approve_?. It seemed like a approve action...
> 
> I'll take your word for it; I don't really know the business logic here. It was just the only one of these that had more than one _Approve_

That's a good catch. I checked the wiki again it does look like a "back" action so I changed it to _neutral_ 👍 

I fixed the missing status error too.

glassblowerscat (approved):
🎉

PR #583, opened 2020-10-13
Additional partner fields

frankyma:
Loading: Screen Shot 2020-10-13 at 1 12 48 PM
Partner Scene: Screen Shot 2020-10-13 at 1 13 37 PM

### DIFF
--- a/src/scenes/Partners/Detail/AddressCard.tsx
+++ b/src/scenes/Partners/Detail/AddressCard.tsx
@@ -12,6 +12,9 @@ import React, { FC } from 'react';
 import { PartnerDetailsFragment } from './PartnerDetail.generated';
 
 const useStyles = makeStyles(({ spacing }) => ({
+  cardContent: {
+    flexGrow: 1,
+  },
   cardActions: {
     display: 'flex',
     justifyContent: 'space-between',
@@ -36,7 +39,7 @@ export const AddressCard: FC<AddressCardProps> = ({
 
   return (
     <Card className={className}>
-      <CardContent>
+      <CardContent className={classes.cardContent}>
         <Grid container direction="column" spacing={1}>
           <Grid item>
             <Typography variant="h4">
--- a/src/scenes/Partners/Detail/AddressCard.tsx
+++ b/src/scenes/Partners/Detail/AddressCard.tsx
@@ -1,6 +1,7 @@
 import {
   Button,
   Card,
+  CardActionArea,
   CardActions,
   CardContent,
   Grid,
@@ -39,15 +40,21 @@ export const AddressCard: FC<AddressCardProps> = ({
 
   return (
     <Card className={className}>
-      <CardContent className={classes.cardContent}>
-        <Grid container direction="column" spacing={1}>
-          <Grid item>
-            <Typography variant="h4">
-              {partner ? partner.address.value : <Skeleton width="75%" />}
-            </Typography>
+      <CardActionArea
+        onClick={onEdit}
+        className={classes.cardContent}
+        disabled={!partner}
+      >
+        <CardContent className={classes.cardContent}>
+          <Grid container direction="column" spacing={1}>
+            <Grid item>
+              <Typography variant="h4">
+                {partner ? partner.address.value : <Skeleton width="75%" />}
+              </Typography>
+            </Grid>
           </Grid>
-        </Grid>
-      </CardContent>
+        </CardContent>
+      </CardActionArea>
       <CardActions className={classes.cardActions}>
         <Button color="primary" disabled={!partner} onClick={onEdit}>
           Edit
--- a/src/scenes/Partners/Detail/AddressCard.tsx
+++ b/src/scenes/Partners/Detail/AddressCard.tsx
@@ -1,0 +1,55 @@
+import {
+  Button,
+  Card,
+  CardActions,
+  CardContent,
+  Grid,
+  makeStyles,
+  Typography,
+} from '@material-ui/core';
+import { Skeleton } from '@material-ui/lab';
+import React, { FC } from 'react';
+import { PartnerDetailsFragment } from './PartnerDetail.generated';
+
+const useStyles = makeStyles(({ spacing }) => ({
+  cardActions: {
+    display: 'flex',
+    justifyContent: 'space-between',
+    paddingRight: spacing(2),
+  },
+}));
+
+interface AddressCardProps {
+  partner?: PartnerDetailsFragment;
+  onEdit: () => void;
+  className?: string;
+}
+
+export const AddressCard: FC<AddressCardProps> = ({
+  partner,
+  className,
+  onEdit,
+}) => {
+  const classes = useStyles();
+
+  // TODO: Implement full address saving and parsing (st, city, state, etc)
+
+  return (
+    <Card className={className}>
+      <CardContent>
+        <Grid container direction="column" spacing={1}>
+          <Grid item>
+            <Typography variant="h4">
+              {partner ? partner.address.value : <Skeleton width="75%" />}
+            </Typography>
+          </Grid>
+        </Grid>
+      </CardContent>
+      <CardActions className={classes.cardActions}>
+        <Button color="primary" disabled={!partner} onClick={onEdit}>
+          Edit
+        </Button>
+      </CardActions>
+    </Card>
+  );
+};
--- a/src/scenes/Partners/Detail/AddressCard.tsx
+++ b/src/scenes/Partners/Detail/AddressCard.tsx
@@ -49,7 +49,11 @@ export const AddressCard: FC<AddressCardProps> = ({
           <Grid container direction="column" spacing={1}>
             <Grid item>
               <Typography variant="h4">
-                {partner ? partner.address.value : <Skeleton width="75%" />}
+                {partner ? (
+                  partner.address.value || 'Add Address'
+                ) : (
+                  <Skeleton width="75%" />
+                )}
               </Typography>
             </Grid>
           </Grid>
--- a/src/scenes/Partners/Detail/PartnerDetail.tsx
+++ b/src/scenes/Partners/Detail/PartnerDetail.tsx
@@ -4,12 +4,11 @@ import {
   CardActionArea,
   CardContent,
   Grid,
-  IconButton,
   makeStyles,
   Tooltip,
   Typography,
 } from '@material-ui/core';
-import { Add } from '@material-ui/icons';
+import { Add, Edit } from '@material-ui/icons';
 import { Skeleton } from '@material-ui/lab';
 import clsx from 'clsx';
 import { Many } from 'lodash';
@@ -19,14 +18,14 @@ import { Avatar } from '../../../components/Avatar';
 import { BooleanProperty } from '../../../components/BooleanProperty';
 import { DataButton } from '../../../components/DataButton';
 import { useDialog } from '../../../components/Dialog';
-import { useDateFormatter } from '../../../components/Formatters';
-import { PencilCircledIcon } from '../../../components/Icons';
+import { Fab } from '../../../components/Fab';
+import { useDateTimeFormatter } from '../../../components/Formatters';
 import { UserListItemCardPortrait } from '../../../components/UserListItemCard';
 import { square } from '../../../util';
 import { EditablePartnerField, EditPartner } from '../Edit';
 import { AddressCard } from './AddressCard';
 import { PartnerDocument } from './PartnerDetail.generated';
-import { PartnerTypeCard } from './PartnerTypesCard';
+import { PartnerTypesCard } from './PartnerTypesCard';
 
 const useStyles = makeStyles(({ spacing, breakpoints, palette }) => ({
   root: {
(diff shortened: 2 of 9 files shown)


src/scenes/Partners/Edit/EditPartner.tsx
```diff
@@ -0,0 +1,161 @@
...
+  types: ({ props }) => (
+    <EnumField
+      multiple
+      label="Types"
+      options={PartnerTypeList}
+      layout="two-column"
+      {...props}
+    />
+  ),
+  //TODO: fix hard crash and missing initialValues
```
frankyma, line 84: Max rerender crashing here, and also for some reason the initialValues aren't populating.

src/scenes/Partners/Detail/AddressCard.tsx
```diff
@@ -0,0 +1,58 @@
+import {
```
frankyma, line 1: Let me know if you think I should move this and `PartnerTypesCard.tsx` to the components directory, wasn't sure if this may be reusued

src/scenes/Partners/Edit/EditPartner.tsx
frankyma: Fixed by Carson's fix!

src/scenes/Partners/Detail/AddressCard.tsx
glassblowerscat: I don't see anything else in the existing designs that resembles this, so I say we just leave it here unless it changes to more closely resemble another existing component. I suppose there's a chance we'll need it or something very similar to it again elsewhere in the app, won't see it in here, and will end up duplicating some code. But ideally any time we need to create some UI we should be looking in Figma to see if it already exist in the designs for a scene we've already implemented, and if so, we ought to find it then.

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -1,16 +1,38 @@
...
 import { Skeleton } from '@material-ui/lab';
 import clsx from 'clsx';
+import { Many } from 'lodash';
 import React from 'react';
 import { useParams } from 'react-router-dom';
+import { BooleanProperty } from '../../../components/BooleanProperty';
+import { DataButton } from '../../../components/DataButton';
+import { useDialog } from '../../../components/Dialog';
+import { useDateFormatter } from '../../../components/Formatters';
+import { PencilCircledIcon } from '../../../components/Icons';
```
glassblowerscat, line 18: This is different than the component we use in other similar scenes. From Language Engagement Detail:

image

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -60,21 +109,97 @@ export const PartnerDetail = () => {
...
+          <div className={classes.sectionTitle}>
+            <Typography variant="h3">Point of Contact</Typography>
+            <IconButton
+              color="primary"
+              aria-label="edit partner"
+              onClick={() => editPartner('pointOfContactId')}
+            >
+              <AddCircle />
+            </IconButton>
+            <Typography>Edit</Typography>
```
glassblowerscat, line 189: This seems like another place where we should maybe use the same FAB as above. I realize this reflects the design, so maybe we need to ask the design team for clarification on this.

If they want to stick with the existing design, we should make sure the word "Edit" is clickable

src/scenes/Partners/Edit/EditPartner.tsx
glassblowerscat: This isn't populating because we're giving a radio input, which takes a single string value, an array, which is what our `UpdatePartner` input wants us to use. This seems like a mismatch between the UI and the schema to me. Are users supposed to be able to submit multiple `financialReportingTypes` for a partner? Then this should be a checkbox field. Otherwise, `UpdatePartner` should take an Enum, not an array of Enums

src/scenes/Partners/Detail/PartnerTypesCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export const PartnerTypeCard: FC<PartnerTypeCardProps> = ({
+  partner,
+  className,
+  onEdit,
+}) => {
+  const classes = useStyles();
+
+  return (
+    <Card className={className}>
```
glassblowerscat, line 39: Our other similar cards are clickable using a `CardActionArea`. The "Edit" button is just there for a UI signal to the user

src/scenes/Partners/Detail/AddressCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+  className,
+  onEdit,
+}) => {
+  const classes = useStyles();
+
+  // TODO: Implement full address saving and parsing (st, city, state, etc)
+
+  return (
+    <Card className={className}>
+      <CardContent className={classes.cardContent}>
```
glassblowerscat, line 42: This should take a `CardActionArea` as well

glassblowerscat (changes requested):

src/scenes/Partners/Detail/PartnerDetail.tsx
CarsonF: Yeah we've translated the word to a tooltip so we don't have to make the word clickable with the FAB

src/scenes/Partners/Edit/EditPartner.tsx
CarsonF: Yes multiple here. This only defines what's allowed to be selected in the partnership object

frankyma: Got it, changed to multiple and fixed.  I noticed that in UpdatePartnership the `financialReportingType` is singular, is that intended?

CarsonF: No it should be multiple. I had to point that out on the API PR too, so it must've been missed in the update.

src/scenes/Partners/Detail/PartnerDetail.tsx
frankyma: I actually copied this from Project Details:
Screen Shot 2020-10-18 at 11 26 08 AM

I believe in language engagement details we have a breadcrumb to project and a link out to language because we have details scenes for those, but here we don't have a detail scene for Organization.

src/scenes/Partners/Detail/PartnerDetail.tsx
frankyma: Sorry guys not sure what the decision here is.  Keep as is (edit word non-clickable) or change it to a tooltip, something else?

CarsonF: Let's not match the designs and instead make it a Fab with a tooltip. This makes it consistent with the rest of the app.

If design doesn't like this then we can talk about the reasons for us doing this (text is clickable that isn't styled as it) and possibly update across the app.

src/scenes/Partners/Detail/PartnerTypesCard.tsx
frankyma: Added `CardActionArea`

CarsonF (changes requested):
Empty state for pmc should start with "Enter ..." to match everything else.

---

Screen Shot 2020-10-20 at 11 57 14 AM

These empty states look pretty poor. Can some more UI be added here?

---

Address should be a textarea. See https://github.com/SeedCompany/cord-field/blob/d2042d4f57da660f6d92d8dbbd3fc68efd2c084b/src/scenes/Users/UserForm/UserForm.tsx#L128-L134

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -60,21 +113,98 @@ export const PartnerDetail = () => {
...
-            {/* {partner ? (
+            <Tooltip title="Update Global Innovations Client">
               <IconButton
                 color="primary"
                 aria-label="edit partner"
-                onClick={editPartner}
+                onClick={() => editPartner('globalInnovationsClient')}
               >
                 <PencilCircledIcon />
               </IconButton>
```
CarsonF, line 123: Use FAB. Tooltip should match aria-label.

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -60,21 +113,98 @@ export const PartnerDetail = () => {
...
-            ) : null} */}
+            </Tooltip>
+          </header>
+          <div className={classes.subheader}>
+            <Typography variant="h4">
+              {partner ? 'Partner Information' : <Skeleton width={200} />}
+            </Typography>
+            {partner && (
+              <Typography variant="body2" color="textSecondary">
+                Created {formatDate(partner.createdAt)}
```
CarsonF, line 132: Use date time formatting for DateTime fields.

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -60,21 +113,98 @@ export const PartnerDetail = () => {
...
+                label="Global Innovations Client"
+                redacted="You do not have permission to view whether this is a Global Innovations Client"
+                data={partner?.globalInnovationsClient}
+                wrap={(node) => <Grid item>{node}</Grid>}
+              />
+            </Grid>
+          </Grid>
+          <Grid container spacing={3}>
+            <Grid item xs={6} className={classes.cardSection}>
+              <Typography variant="h3">Partner Types</Typography>
```
CarsonF, line 168: It's weird that these don't have loading states. The scene type even does ("Partner")

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -60,21 +113,98 @@ export const PartnerDetail = () => {
...
+                redacted="You do not have permission to view Status"
+                children={partner?.active.value ? 'Active' : 'Inactive'}
+              />
+            </Grid>
+            <Grid item>
+              <BooleanProperty
+                label="Global Innovations Client"
+                redacted="You do not have permission to view whether this is a Global Innovations Client"
+                data={partner?.globalInnovationsClient}
+                wrap={(node) => <Grid item>{node}</Grid>}
```
CarsonF, line 162: You've got two grid items here. One is unnecessary.

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -60,21 +113,98 @@ export const PartnerDetail = () => {
...
+                onClick={() => editPartner('pmcEntityCode')}
+                secured={partner?.pmcEntityCode}
+                redacted="You do not have permission to view PMC Entity Code"
+                children={`PMC Entity Code${
+                  partner?.pmcEntityCode.value
+                    ? `: ${partner.pmcEntityCode.value}`
+                    : ''
+                }`}
+              />
+            </Grid>
```
CarsonF, line 148: Let's put status first, like the designs have it. Set the `empty` prop here too.

CarsonF (dismissed):

src/scenes/Partners/Detail/PartnerTypesCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+                ) : (
+                  <Skeleton width="75%" />
+                )}
+              </Typography>
+            </Grid>
+
+            {partner?.types.value.includes('Managing') && (
+              <Grid item>
+                <Typography variant="body2" color="textSecondary">
+                  Financial Reporting Type
```
CarsonF, line 61:
```suggestion
                  Financial Reporting Types
```

frankyma:
> Empty state for pmc should start with "Enter ..." to match everything else.

Changed.  I got that from LanguageEngagementDetail DataButton so changed it there also

Added Empty State text to cards 
Screen Shot 2020-10-20 at 12 26 42 PM

> Address should be a textarea. See
> 
> https://github.com/SeedCompany/cord-field/blob/d2042d4f57da660f6d92d8dbbd3fc68efd2c084b/src/scenes/Users/UserForm/UserForm.tsx#L128-L134

Changed

src/scenes/Partners/Detail/PartnerDetail.tsx
frankyma: Changed.  I haven't used many FABs before this, what's the difference between this and a regular icon button?  From what I've read looks like they're mostly for those fixed position buttons on mobile screens?

src/scenes/Partners/Detail/PartnerDetail.tsx
frankyma: Changed.  I did hear Amy say that we don't need to show the time on alot of these, if that becomes important we can either just use the date formatter for dateTime or we'll have to change the type

src/scenes/Partners/Detail/PartnerDetail.tsx
frankyma: Added loading states, this is what the loading screen looks like now:
Screen Shot 2020-10-20 at 1 32 37 PM

src/scenes/Partners/Detail/PartnerDetail.tsx
frankyma: Changed

glassblowerscat (approved):
🕺

PR #590, opened 2020-10-16
Location CRUD

frankyma:
Create on Location Field (that's not being used anywhere because adding locations on resources not ready).
Screen Shot 2020-10-16 at 10 49 52 AM

Location Form, which is being used by Create and Edit location.  Create location tied to Location Field.
Screen Shot 2020-10-16 at 10 37 11 AM

Funding Account Field tied to a create funding account form.
Screen Shot 2020-10-16 at 10 37 25 AM

Location Detail View that ties to edit location.
Screen Shot 2020-10-16 at 10 38 32 AM

Location and funding Account Cards
Screen Shot 2020-10-16 at 11 03 42 AM
Screen Shot 2020-10-16 at 11 03 35 AM

Not on this PR: funding account detail scene or updating a specific funding account (different from updating the funding account on a location)

### DIFF
--- a/src/components/form/Lookup/FundingAccount/FundingAccountField.tsx
+++ b/src/components/form/Lookup/FundingAccount/FundingAccountField.tsx
@@ -1,5 +1,6 @@
 import { LookupField } from '../..';
 import { CreateFundingAccountInput } from '../../../../api';
+import { CreateFundingAccount } from '../../../../scenes/Locations/LocationForm/FundingAccount/CreateFundingAccount';
 import {
   FundingAccountLookupItemFragment as FundingAccount,
   FundingAccountLookupDocument,
@@ -13,4 +14,11 @@ export const FundingAccountField = LookupField.createFor<
   lookupDocument: FundingAccountLookupDocument,
   label: 'Funding Account',
   placeholder: 'Search for a funding account by name',
+  CreateDialogForm: CreateFundingAccount,
+  // @ts-expect-error don't need to pass through entire initialValues
+  getInitialValues: (val) => ({
+    fundingAccount: {
+      name: val,
+    },
+  }),
 });
--- a/src/components/form/Lookup/FundingAccount/FundingAccountField.tsx
+++ b/src/components/form/Lookup/FundingAccount/FundingAccountField.tsx
@@ -1,0 +1,16 @@
+import { LookupField } from '../..';
+import { CreateFundingAccountInput } from '../../../../api';
+import {
+  FundingAccountLookupItemFragment as FundingAccount,
+  FundingAccountLookupDocument,
+} from './FundingAccountField.generated';
+
+export const FundingAccountField = LookupField.createFor<
+  FundingAccount,
+  CreateFundingAccountInput
+>({
+  resource: 'FundingAccount',
+  lookupDocument: FundingAccountLookupDocument,
+  label: 'Funding Account',
+  placeholder: 'Search for a funding account by name',
+});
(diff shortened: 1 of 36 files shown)


src/scenes/Locations/LocationForm/LocationForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+              selectOptions={locationTypeSelectOptions}
+              required
+              defaultValue="City"
+              {...props}
+            />
+          )}
+        </SecuredField>
+      </Grid>
+    </Grid>
+    {title === 'Create Location' && (
```
frankyma, line 117: Once update funding account field on location is in, can remove the conditional https://github.com/SeedCompany/cord-api-v3/pull/1374 is

src/components/form/Lookup/FundingAccount/FundingAccountField.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const FundingAccountField = LookupField.createFor<
+  FundingAccount,
+  CreateFundingAccountInput
+>({
+  resource: 'FundingAccount',
+  useLookup: useFundingAccountLookupLazyQuery,
+  label: 'Funding Account',
+  placeholder: 'Search for a funding account by name',
+  CreateDialogForm: CreateFundingAccount,
+  // @ts-expect-error don't need to pass through entire initialValues
```
frankyma, line 18: Saw the same expect error tag on userfield also, but let me know if there's a better way around the error

src/scenes/Locations/LocationForm/FundingAccount/CreateFundingAccount.tsx
```diff
@@ -0,0 +1,52 @@
...
+        name="name"
+        label="Account Name"
+        placeholder="Enter account name"
+        required
+      />
+      <NumberField
+        name="accountNumber"
+        label="Account Number"
+        placeholder="Enter account number"
+        formatInput={(numString) => numString.replace(',', '')}
```
frankyma, line 47: Still has the bug where if you click away on the field the commas come back, will look into it separately

CarsonF: Let's not even worry about it. Production data has 7 funding accounts. All have "account numbers" of 0-7. There's no need to try and handle thousands separator.

CarsonF:
~Let's make the type default to country.~ Nvm I'd rather change this on the API so we don't have to "hardcode" an enum value here.

Can we change "ISO Alpha-3 Code" to "ISO Alpha-3 Country Code"?

src/scenes/Locations/LocationForm/FundingAccount/CreateFundingAccount.tsx
frankyma: fixed with reppaceAll

frankyma:
> ~Let's make the type default to country.~ Nvm I'd rather change this on the API so we don't have to "hardcode" an enum value here.

Are you talking about on FundingAccountField.tsx?

> Can we change "ISO Alpha-3 Code" to "ISO Alpha-3 Country Code"?

Done

src/scenes/Locations/LocationForm/LocationForm.tsx
CarsonF: Done

glassblowerscat:
This looks good, but I'm unable to check it in UI because I'm having some weird issue where, even if I create things as root, I don't have read permissions. So I can run the `createLocation` mutation in devtools, but I can't then view the Location Detail scene in the UI for it. @CarsonF, can you take another look?

PR #591, opened 2020-10-16
Debug API delays

CarsonF:
A lot of times are trying to dev loading states and we want to simulate more delay from the API. This is usually done with the browser's network throttling. However, this is frustrating because it also slows down retrieving our UI code with every change. Which is typically very large for non-production - a few MBs at least.

So this change allows you to only "slow down" (delay) requests to the API. This increases efficiency while deving because it shortens the cycle between making code change, seeing loading state in UI, seeing UI transition to real data, repeat.

The delay can be set in dev console
```
API_DEBUG.delay = 5000; // milliseconds
```
It's persisted in history so it doesn't reset on page refreshes (either from your CMD+R or code changes).

### DIFF
--- a/src/api/ApolloProvider.tsx
+++ b/src/api/ApolloProvider.tsx
@@ -25,11 +25,27 @@ import { typePolicies } from './typePolicies';
 
 const serverHost = process.env.REACT_APP_API_BASE_URL || '';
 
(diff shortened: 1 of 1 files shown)


glassblowerscat (approved):
🙌

PR #592, opened 2020-10-20
Use powers to check create

frankyma:
Ask for powers in session, login and register, and merge all into session cache with `updateSessionCache`.  have `useSession` hook also return powers to be used at the component level.

Have `LookupField` take in a `canCreate` prop and tie it to `freeSolo`.  Now every time a field is used must pass in `canCreate=true` to enable the create functionality.

### DIFF
--- a/src/components/Session/Session.tsx
+++ b/src/components/Session/Session.tsx
@@ -11,8 +11,9 @@ import {
 export const useSession = () => {
   const { data, loading: sessionLoading } = useSessionQuery();
   const session = data?.session.user;
+  const powers = data?.session.powers;
 
-  return { session, sessionLoading };
+  return { session, sessionLoading, powers };
 };
 
 export const updateSessionCache = <T extends LoginMutation | RegisterMutation>(
--- a/src/components/Session/Session.tsx
+++ b/src/components/Session/Session.tsx
@@ -1,4 +1,5 @@
 import { ApolloCache } from '@apollo/client';
+import { SessionOutput } from '../../api';
 import { LoginMutation } from '../../scenes/Authentication/Login/Login.generated';
 import { RegisterMutation } from '../../scenes/Authentication/Register/register.generated';
 import {
@@ -18,8 +19,9 @@ export const useSession = () => {
 
 export const updateSessionCache = <T extends LoginMutation | RegisterMutation>(
   cache: ApolloCache<T>,
-  user: LoggedInUserFragment
+  sessionData: { user?: LoggedInUserFragment; powers?: SessionOutput['powers'] }
 ) => {
+  const { user, powers } = sessionData;
   const currentSession = cache.readQuery<SessionQuery>({
     query: SessionDocument,
   });
@@ -32,6 +34,7 @@ export const updateSessionCache = <T extends LoginMutation | RegisterMutation>(
           ...currentSession.session.user,
           ...user,
         },
+        powers: [...(powers ? powers : currentSession.session.powers || [])],
       },
     };
     cache.writeQuery({ query: SessionDocument, data: updatedSession });
--- a/src/components/form/Lookup/LookupField.tsx
+++ b/src/components/form/Lookup/LookupField.tsx
@@ -22,9 +22,10 @@ import React, {
   useState,
 } from 'react';
 import { Except, SetOptional } from 'type-fest';
-import { isNetworkRequestInFlight } from '../../../api';
+import { isNetworkRequestInFlight, Power } from '../../../api';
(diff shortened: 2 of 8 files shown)


src/components/Session/Session.tsx
```diff
@@ -11,13 +12,15 @@ import {
 export const useSession = () => {
   const { data, loading: sessionLoading } = useSessionQuery();
   const session = data?.session.user;
+  const powers = data?.session.powers;
 
-  return { session, sessionLoading };
+  return { session, sessionLoading, powers };
```
CarsonF, line 17: Powers is already in the session object. I think the moving it up here is redundant.

src/scenes/Authentication/Login/Login.tsx
```diff
@@ -27,9 +27,10 @@ export const Login = (props: Except<Props, 'onSubmit'>) => {
         variables: { input },
         update: (cache, { data }) => {
           const user = data?.login.user;
-          if (user) {
+          const powers = data?.login.powers;
+          if (user || powers) {
             setSuccess(true);
-            updateSessionCache(cache, user);
+            updateSessionCache(cache, user, powers);
           }
```
CarsonF, line 34: Just a small thing but maybe `updateSessionCache` could take a session-like object instead `{ user, powers }`.

```tsx
          const session = data?.login;
          if (session) {
            setSuccess(true);
            updateSessionCache(cache, session);
          }
```

src/components/form/Lookup/LookupField.tsx
```diff
@@ -105,9 +106,10 @@ export function LookupField<
   getCompareBy,
   getOptionLabel: getOptionLabelProp,
   variant,
+  canCreate = false,
```
CarsonF, line 109: Let's move the power check inside of this component. Pulling this boolean shouldn't be a responsibility of the callers of this component. It's already opinionated and aware of our API anyways.

The `createFor` could default it as well.
```tsx
<LookupField<...>
  ...
  createPower={`Create${resource}`}
  {...config}
  {...props}
/>
```

CarsonF (changes requested):

src/components/Session/Session.tsx
frankyma: The `session` is actually just the user as you can see from above 

`const session = data?.session.user;`

I thought about renaming it but it's used in alot of files, and wasn't sure if it's named as `session` instead of `user` in the first place for a reason?  Let me know what you think

src/scenes/Authentication/Login/Login.tsx
frankyma: It can...althought technically `login` doesn't have to match session, so inside the function we'd have to do the destructuring

src/components/Session/Session.tsx
CarsonF: Ah I see. It's only used in a few spots anyways so let's just leave it.

src/scenes/Authentication/Login/Login.tsx
CarsonF: How does it not match?
```tsx
const updateSessionCache = (cache, newSession: { user: User, powers: Power[] }) => {};
```

frankyma: They match now because we made the fragments match (we had to manually add the powers field on the login and register queries), but they are still separate fragments.  But currently yes they are the same data.

I'm not even sure we're concerned about the same thing, but yes I can pass in this session like object, it's cleaner

src/components/form/Lookup/LookupField.tsx
frankyma: It has to be set in `createFor` right?  How else will the component know what power to check?

CarsonF: Right. The hook will still need to be called in the LookupField component.

src/scenes/Authentication/Login/Login.tsx
CarsonF: Sure I didn't mean change the gql fragments.

src/components/form/Lookup/LookupField.tsx
frankyma: Updated.  Is there a way to make this resource type more strict?  Even if we manually create a list of Resource enums...  A misspelling will cause the powers check to not work

CarsonF: Yeah I wish there was. I think it might be possible with TS 4.1. For now it's ok.

CarsonF (approved):

PR #601, opened 2020-10-20
Add status on lang engagement detail

frankyma:
Screen Shot 2020-10-20 at 4 59 19 PM


CarsonF:
Would you mind putting it before paratext id? I think it's more important than that.

frankyma:
@CarsonF should I make status first on internship engagement also for consistency?  For reference here's the current order
Screen Shot 2020-10-21 at 8 53 54 AM

CarsonF:
Yeah let's put the status first on both. Then the date range, which looks to be missing from language engs. Then specifics for each type.

frankyma:
Put status and date range first on both

CarsonF (approved):
Thanks @frankyma

PR #602, opened 2020-10-21
Add tooltips to fab icons

frankyma:
Only one left out is engagement loading scene but I don't think we need a tooltip there.
https://github.com/SeedCompany/cord-field/blob/41b471fba6ec34c367746f56468ccb573e515d95/src/scenes/Engagement/EngagementDetailLoading.tsx#L50

I used the same text as the fab aria labels, but that means some are upper case and some are lower case.  Let me know if that needs to be normalized.

### DIFF
--- a/src/scenes/Languages/Detail/LanguageDetail.tsx
+++ b/src/scenes/Languages/Detail/LanguageDetail.tsx
@@ -111,9 +111,11 @@ export const LanguageDetail = () => {
               )}
             </Typography>
             {canEditAnyFields ? (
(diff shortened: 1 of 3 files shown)


CarsonF:
> I used the same text as the fab aria labels, but that means some are upper case and some are lower case. Let me know if that needs to be normalized.

Yeah screen readers don't care about capitalization - all the text it gets read the same way. For tooltips in the UI though, we should make them proper sentence case.

frankyma:
Capitalized the tooltip text

CarsonF (approved):

PR #614, opened 2020-10-21
Refresh Project Files List on new file upload

### LINKED ISSUE #612: [BUG] Major delay in showing uploaded file
Uploaded files show successful but do not appear in the file browser for a very long time, leading users to assume that the upload was unsuccessful.

Can we get the file to display when successfully uploaded or show a loading spinner until the file is available to be viewed?

glassblowerscat:
Closes #612 

- Add new `ProjectDirectoryContents` fragment and use in `ProjectDirectory` query instead of inline fields
- Update `useUploadProjectFiles` hook to write newly uploaded file into cache for immediate UI refresh


CarsonF (approved):
Sweet!

PR #621, opened 2020-10-23
Typed graphql documents instead of generated hooks

CarsonF:
Now instead of custom hooks being generated to provide strict types, documents are generated with their types declared on them. The gql library types are patched on handle these typed documents.
```diff
-const {} = useSessionQuery();
+const {} = useQuery(SessionDocument);
```

This allows everything to be more straight forward with types "just working".


glassblowerscat (approved):
💪

PR #623, opened 2020-10-25
547 update cache after user actions

### LINKED ISSUE #547: [BUG] Cache is not updating after user actions.
- View list of projects filter to 'My Projects' and have 5.
- I create a new Project and add information to it, I'm listed on the Project as a Project Manager.
- Click on the Projects list to see Projects.

Expected Results: I will see 6 Projects, including my newly created Project

Actual Results: I only see the 5 Projects I saw at first. I have to force-refresh the page to see all 6.

glassblowerscat:
Closes #547 

For all top-level "create" mutations—Project, Language, User, Partner—we call the `update` method and use `cache.modify` to write the new item to the `items` field of the relevant List query. This required refactoring all relevant mutations to use a fragment so that fragment can be used in the write operation.

### DIFF
--- a/src/scenes/Languages/Create/CreateLanguage.graphql
+++ b/src/scenes/Languages/Create/CreateLanguage.graphql
@@ -1,17 +1,9 @@
 mutation createLanguage($input: CreateLanguageInput!) {
   createLanguage(input: $input) {
     language {
-      ...NewLanguage
+      id
+      name { value }
+      ...LanguageListItem
     }
   }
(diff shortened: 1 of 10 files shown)


glassblowerscat:
@CarsonF So, even though I'm up and running with the app again I still can't really tests this because neither the root user nor another user I create can actually read the items they create. But the code looks good!

CarsonF (approved):

PR #627, opened 2020-10-26
Partnerships dialog refactor

frankyma:
On create and update partnership, the type and financialReportingTypes (FRT) fields need to reflect the values on the Partner.  So the enumField has the available types passed in: https://github.com/SeedCompany/cord-field/pull/627/files#diff-04c13b2404d5a2cdbcdf6d57823dc8297575a613af9d032c52beda3a7d73709cR75

Also, if the partner is changed on create, the type and FRT values need to be wiped (or they will be part of the submit input).  Same goes with if the Managing Type is checked or not for FRT.

### DIFF
--- a/src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
+++ b/src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
@@ -6,7 +6,6 @@ import {
   PartnershipAgreementStatus,
   PartnershipAgreementStatusList,
   PartnerType,
-  PartnerTypeList,
 } from '../../../api';
 import {
   DialogForm,
@@ -18,14 +17,25 @@ import {
   SecuredField,
   SubmitError,
 } from '../../../components/form';
-import { PartnerField } from '../../../components/form/Lookup';
+import {
+  PartnerField,
+  PartnerLookupItem,
+} from '../../../components/form/Lookup';
 import { Nullable } from '../../../util';
 import { CreatePartnershipFormInput } from '../Create';
 import { EditPartnershipFormInput } from '../Edit';
 import { PartnershipFormFragment } from './PartnershipForm.generated';
 
+type PartnershipformValues = Partial<
+  CreatePartnershipFormInput | EditPartnershipFormInput
+> & {
+  partnership?: {
+    partnerLookupItem?: PartnerLookupItem;
+  };
+};
+
 export type PartnershipFormProps<
-  T extends Partial<CreatePartnershipFormInput | EditPartnershipFormInput>
+  T extends PartnershipformValues
 > = DialogFormProps<T> & {
   partnership?: PartnershipFormFragment;
 };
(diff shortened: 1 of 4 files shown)


src/scenes/Partnerships/PartnershipForm/renderFields.tsx
```diff
@@ -0,0 +1,118 @@
...
+import {
+  EnumField,
+  EnumFieldProps,
+  SecuredField,
+  SubmitError,
+} from '../../../components/form';
+import { PartnerField } from '../../../components/form/Lookup';
+import { Nullable } from '../../../util';
+
+export const renderFields = (partnership?: PartnershipFormFragment) => ({
```
glassblowerscat, line 24: This is a React component, so it should be named `RenderFields`, not `renderFields`

src/scenes/Partnerships/PartnershipForm/renderFields.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const currentPartnerFinTypes =
+    partnership?.partner.value?.financialReportingTypes.value;
+
+  const prevPartnerId = useRef<string>();
+  const prevTypes = useRef<Nullable<readonly PartnerType[]>>();
+
+  useEffect(() => {
+    // if user selects a different partner (on create partnership), wipe the types and fin type values
+    if (lookupPartnerObj?.id !== prevPartnerId.current) {
+      form.change('partnership.types', undefined);
```
glassblowerscat, line 45: I'm seeing this behavior here, which seems like not how this is intended to work?

partnership

glassblowerscat (changes requested):

src/scenes/Partnerships/PartnershipForm/renderFields.tsx
frankyma: Well technically it's a function that takes in a partnership object and returns a react component, I named it this way to make the distinction and make sure the next developer knows it's different.  I'm not sure if this thing has a formal name or what the best naming convention is.  Thoughts?

glassblowerscat: Yeah, I guess you're right... my eyes kind of glossed over the fact that you weren't passing props but a single argument. But you're calling `useEffect`, so my thinking is this *should* be a React component and should be named like one and called like one.

glassblowerscat: (Because otherwise what you're technically doing is calling `useEffect` inside the `return` of a function component, which is a violation of the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html))

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -18,73 +18,129 @@ import {
...
-                options={FinancialReportingTypeList}
-                getLabel={displayFinancialReportingType}
-                {...props}
-              />
-            )}
-          </SecuredField>
-        ) : null}
-        {partnership && (
+}: PartnershipFormProps<T>) => {
+  const decorators = useMemo<Array<Decorator<T>>>(
```
frankyma, line 51: Defining the decorators here because it needs the generic `T` to match when passed in to `DialogForm` (which receives the same generic).  Let me know if there's a better way

CarsonF: The specific generic type really isn't that important for these decorators (esp since keys are just strings due to dot notation).
You could use `PartnershipFormValues` instead and defined the array outside of the component saving the memo step.
Since we know/enforce that the generic conforms to `PartnershipFormValues` as long as we only need types from `PartnershipFormValues` the generic doesn't add anything. Does that make sense?

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -18,73 +18,129 @@ import {
...
+  CreatePartnershipFormInput | EditPartnershipFormInput
+> & {
+  partnership?: {
+    partnerLookupItem?: PartnerLookupItem;
+  };
+};
+
 export type PartnershipFormProps<
-  T extends Partial<CreatePartnershipFormInput | EditPartnershipFormInput>
+  T extends PartnershipformValues
```
frankyma, line 39: capitalize Form

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
frankyma: Done, moved outside of component

glassblowerscat (approved):

PR #632, opened 2020-10-27
Clarify Project date sort labels

frankyma:
Match asc desc to cypher

### DIFF
--- a/src/scenes/Projects/List/ProjectSortOptions.tsx
+++ b/src/scenes/Projects/List/ProjectSortOptions.tsx
@@ -19,8 +19,8 @@ export const ProjectSortOptions = () => (
     <Option
       value="estimatedSubmission"
       label="Estimated Submission Date"
-      asc="Furthest Date to Closest"
-      desc="Closest Date to Furthest"
+      asc="Earliest Date to Latest"
+      desc="Latest Date to Earliest"
       defaultOrder="DESC"
     />
   </>


CarsonF:
This seems backwards...

frankyma:
It's not, that's how Cypher orders dates.  You can check on master it's backwards right now

CarsonF:
I mean logically. Sorting dates _ascending_ should be _oldest to newest_.
Wikipedia agrees:
> A standard order is often called ascending (corresponding to the fact that the standard order of numbers is ascending, i.e. A to Z, 0 to 9), the reverse order descending (Z to A, 9 to 0). For dates and times, ascending means that earlier values precede later ones e.g. 1/1/2000 will sort ahead of 1/1/2001.
https://en.wikipedia.org/wiki/Sorting#:~:text=A%20standard%20order%20is%20often,of%201%2F1%2F2001.

If cypher does it backwards we should handle that on the API.

frankyma:
you know what, our **Naming** for the date sort description is unclear.  We use words like "closest' and "furthest", but we really should use words like "earliest" and "latest".  You and I had different definitions of what closest and furthest meant.

CarsonF:
That's a good point. closest/furthest is could be interrupted different based on whether the date is in the past or future. I agree that earliest/latest removes that ambiguity.

frankyma:
Changed

CarsonF (commented):
Can you flip the the props back?

frankyma:
Sure switched back

CarsonF (approved):
👍

PR #647, opened 2020-10-29
Only assign field errors if the form has the field registered

CarsonF:
This prevents the form from showing "nothing" because the API error had a field that didn't match a form field that we've defined.


glassblowerscat (approved):
🤞

PR #654, opened 2020-11-02
First Scripture Property

### LINKED ISSUE #39: First Scripture Flag Validation
We need to validate that the 'First Scripture Flag' has not been selected for a prior Language Engagement before displaying the input and allowing a user to select it for a new Language Engagement.

It would probably be good to add a 'First Scripture' or 'Has Scripture' field to a Language that is set based on the Language Engagements if not manually set.

- Language has a boolean that identifies if it has scripture that isn't recorded in CORD Field.

glassblowerscat:
Closes #39 

- Update Language Detail to display whether the language is marked as having an external First Scripture OR has an engagement that is marked as `firstScripture`
- Update Language Form to include a `hasExternalFirstScripture` checkbox (validation will be performed by the API)

### DIFF
--- a/src/scenes/Languages/Detail/LanguageDetail.tsx
+++ b/src/scenes/Languages/Detail/LanguageDetail.tsx
@@ -28,6 +28,7 @@ import { EditLanguage } from '../Edit';
 import { AddLocationToLanguageForm } from '../Edit/AddLocationToLanguageForm';
 import {
   LanguageDocument,
+  LanguageProjectEngagement_LanguageEngagement_Fragment as LanguageFragment,
   RemoveLocationFromLanguageDocument,
 } from './LanguageDetail.generated';
 import { LeastOfThese } from './LeastOfThese';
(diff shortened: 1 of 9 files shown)


src/scenes/Languages/LanguageForm/LanguageForm.tsx
```diff
@@ -191,6 +191,17 @@ export const LanguageForm = <T extends any>({
                       </Grid>
                     )}
                   </SecuredField>
+                  <SecuredField obj={language} name="hasExternalFirstScripture">
+                    {(props) => (
+                      <Grid item xs={12} sm={6}>
+                        <CheckboxField
+                          label="Was this language's first scripture produced outside of CORD?"
+                          margin="none"
```
CarsonF, line 199: Can you add the `keepHelperTextSpacing` flag here? This is an unusual case where we are expecting the server to error on this field if it's not valid based on the rules.

src/scenes/Languages/Detail/LanguageDetail.tsx
```diff
@@ -188,6 +226,13 @@ export const LanguageDetail = () => {
             }
             loading={!language}
           />
+
+          <BooleanProperty
+            label="First Scripture"
+            redacted="You do not have permission to view whether the language has a first Scripture"
+            data={hasFirstScripture}
+            wrap={(node) => <Grid item>{node}</Grid>}
+          />
```
CarsonF, line 221: I think we need to show more than a boolean here. The options are an engagement, external, or false. It would be useful to show that first scripture is fulfilled by X engagement (name/link) or fulfilled externally, or not fulfilled.

This complexity is why I voted not to have it as a calculated field in the API.

CarsonF (changes requested):
I couldn't help myself and replaced the `.reduce` usage 😬 

I think we need to do more than show a simple boolean, see note below.

CarsonF (approved):

PR #672, opened 2020-11-03
Remove add proj button language detail

frankyma:


CarsonF (commented):
Grid container probably isn't needed either

frankyma:
Removed the grid and added in margins

src/scenes/Languages/Detail/LanguageDetail.tsx
```diff
@@ -52,6 +52,11 @@ const useStyles = makeStyles(({ spacing }) => ({
   hidden: {
     visibility: 'hidden',
   },
+  projectSection: {
+    '& > *': {
+      marginBottom: spacing(2),
+    },
```
CarsonF, line 58: the list items already have this applied. Maybe we can just use paragraph or gutterBottom on the title instead of ths?

frankyma: yeah good point.  The gutterBottom isn't big enough so pointed to `h3`

CarsonF: Did you try paragraph? If so, then let's just move the css class to typography instead unnecessarily going up to parent section and then down in css.

frankyma: Didn't know about `paragraph` but that actually works perfect
Screen Shot 2020-11-05 at 8 49 33 AM
Screen Shot 2020-11-05 at 8 49 39 AM

CarsonF (approved):

PR #673, opened 2020-11-03
Expose mou start and end overrides on partnership Edit

frankyma:
@CarsonF should I also show the dates on the partnership cards?

Also this seem different than the other override values?  There's no direct `mouDate` field on the Partnership object

### DIFF
--- a/src/components/PartnershipCard/PartnershipCard.graphql
+++ b/src/components/PartnershipCard/PartnershipCard.graphql
@@ -10,6 +10,16 @@ fragment PartnershipCard on Partnership {
   mouStatus {
     value
   }
+  mouStart {
+    canRead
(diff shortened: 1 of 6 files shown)


CarsonF:
> @CarsonF should I also show the dates on the partnership cards?
> 
> Also this seem different than the other override values? There's no direct `mouDate` field on the Partnership object

Yes we probably should. mouStart/mouEnd do exist on the Partnership.

frankyma:
> Yes we probably should. mouStart/mouEnd do exist on the Partnership.

Added (card still pending final design).  Not sure how I didn't see the dates, brain must've skipped.

Screen Shot 2020-11-04 at 12 03 02 PM

PR #742, opened 2021-03-03
Restrict project completion if non terminal engagements

blakekrammes:
Depends on API PR https://github.com/SeedCompany/cord-api-v3/pull/1686

### DIFF
--- a/src/scenes/Projects/Update/ProjectWorkflowDialog.tsx
+++ b/src/scenes/Projects/Update/ProjectWorkflowDialog.tsx
@@ -88,9 +88,12 @@ export const ProjectWorkflowDialog = ({
       <Grid container direction="column" spacing={1}>
         {transitions.map((transition, i) => (
           <Tooltip
-            title={`This will change the project step to ${displayProjectStep(
-              transition.to
-            )}`}
+            title={
+              transition.disabledReason ??
+              `This will change the project step to ${displayProjectStep(
+                transition.to
+              )}`
+            }
             key={i}
           >
             <Grid item>
@@ -100,6 +103,7 @@ export const ProjectWorkflowDialog = ({
                 action={`${transition.to}:${i}`}
                 color={transitionTypeToColor[transition.type]}
                 variant={transition.type === 'Approve' ? 'contained' : 'text'}
+                disabled={transition.disabled}
               >
                 {transition.label}
               </SubmitButton>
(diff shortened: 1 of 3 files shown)


src/scenes/Projects/Update/ProjectWorkflowDialog.tsx
```diff
@@ -87,9 +87,13 @@ export const ProjectWorkflowDialog = ({
       <Grid container direction="column" spacing={1}>
         {transitions.map((transition, i) => (
           <Tooltip
-            title={`This will change the project step to ${displayProjectStep(
-              transition.to
-            )}`}
+            title={
+              !transition.enabled
+                ? 'The project step cannot be changed to Completed until all engagements have a terminal status'
```
CarsonF, line 92: We should not assume this is the reason the transition is not enabled. The API should provide the reason why it's not enabled.

CarsonF (changes requested):

src/scenes/Projects/Update/ProjectWorkflowDialog.tsx
blakekrammes: What if this is the only one?

CarsonF: If it feels unnecessary, trust me it's not. There's lots of validation we'd like to add in the future, that's dependent upon the current state. i.e. "the project cannot become active if it doesn't have start or end dates defined."

blakekrammes: What if we check `enabled` and `to`? Is your main concern keeping the FE clean?

blakekrammes: Though I guess this does seem like more of an API responsibility.

blakekrammes:
@CarsonF, @sethmcknight  How should we handle engagement statuses if an admin overrides a project to completed? Should that be allowed or prohibited if there are non terminal engagement statuses?

CarsonF:
Overrides are treated like normally workflow, so I would say yes the engagements should be updated.

blakekrammes:
But the normal workflow in this case is to prohibit engagement statuses being updated unless they are individually updated to a terminal status.

CarsonF (approved):

Issue #718, opened 2021-02-02
Project Favorites/Subscriptions Design


sethmcknight:
- [x] Define 
- [x] Design 
- [x] Review
- [x] Prototype
- [ ] Test
- [ ] Analyze & Report
- [ ] Review 
- [ ] Design Adjustments (if necessary)

- Allow users who are not part of the Project Team to Favorite/Subscribe to a Project
- This Project would then show up in their ‘My Projects’ list
- Especially useful for EDR’s, Communications, etc
- Would be good to denote on the Project card if it is a Favorite/Subscription

sethmcknight:
I need to get a list of users to test with. We should go ahead and set up those testing sessions this sprint.

We should plan to have designs done by the end of this sprint. We'll then test and refine next sprint.

sethmcknight:
@AmyRickwartz here's some users you could reach out to for testing:

Advancement
- Becca Lewis
- Annie Jones
- Jessica Lambaise
- Amy Reynolds
- Dana Fay
- Tony Pitaniello
- Scottye Perry
- Sweena Varghese

Communications
- Paula Kordic
- Kim Driggers
- Lauren Kidd

ParkeBrown:
Blocker on getting prototype to work; Amy going to reach out to Patrick

AmyRickwartz:
> Blocker on getting prototype to work; Amy going to reach out to Patrick

Got it figured out. Working on the prototypes again.

AmyRickwartz:
Defining | Designing | Review | Testing | Anylyize Results | Housekeeping | Complete
-- | -- | -- | -- | -- | -- | --


 

Define

 Design

 Review

 Prototype

 Test

 Analyze & Report

 Review

 Design Adjustments (if necessary)

Allow users who are not part of the Project Team to Favorite/Subscribe to a Project

This Project would then show up in their ‘My Projects’ list

Especially useful for EDR’s, Communications, etc

Would be good to denote on the Project card if it is a Favorite/Subscription

AmyRickwartz:
`
<a href="https://www.figma.com/proto/IfEmcJeuNY6ucgjYctlKB7/Cord-Field?node-id=4312%3A31014&amp;scaling=min-zoom">

<span style="font-size:13pt;background-color:#F8F8F8;font-family:Arial,sans-serif;">
Prototype 2</span></a></span>
<br>
<b><span style="color:#1D1C1D;font-size:11.5pt;font-family:Arial,sans-serif;">
[...]

PR #761, opened 2021-03-17
Show pin button in project list and detail page

zkhin:
Fixes: https://github.com/SeedCompany/cord-field/issues/718
Need to check what contents should be shown under the `Recent` tab.


src/components/ProjectPinButton/ProjectPinButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+  };
+});
+
+export interface ProjectPinButtonProps {
+  projectId?: string;
+  pinned?: boolean;
+  className?: string;
+}
+
+export const ProjectPinButton: FC<ProjectPinButtonProps> = ({
```
CarsonF, line 23: Ok several things here. We can decouple this button from projects, and use Apollo to store the state.
```gql
fragment TogglePin on Pinnable {
  id
  pinned
}
```
I missed `id` here in the API, but it should be added.

This fragment allows us two things.
1) we can add it to our other fragments
```gql
fragment ProjectListItem on Project {
  ...TogglePin
}
fragment ProjectOverview on Project {
  ...TogglePin
}
```
2) we can reference this fragment in our props

```tsx
export interface TogglePinButtonProps {
  object: TogglePinFragment;
  ...
}
```
```tsx
props.object.pinned ? <Filled /> : <Outline />
```

3) With this fragment we can update the cache directly instead of having to do refetch queries.
```tsx
const [togglePinned] = useMutation(TogglePinnedDocument, {
  variables: {
    id: props.object.id,
  },
  update: (cache, result) => {
    if (!result.data) {
      return;
    }
    cache.modify({
      id: cache.identify(props.object),
      fields: {
        pinned: () => result.data.togglePinned,
      },
    });
  },
});
```
(I haven't tested exactly so it may need some tweaks)

Then we can just pass the function right to the button handler
```tsx
<IconButton onClick={togglePinned}
```

We don't need to manage state ourselves because Apollo cache will handle that.

src/components/ProjectPinButton/ProjectPinButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+      } catch (err) {
+        setIsPinned(!isPinned);
+      }
+    },
+    [isPinned, projectId, togglePinned]
+  );
+
+  return (
+    <div className={className}>
+      <IconButton onClick={handlePin} className={classes.pinIcon}>
```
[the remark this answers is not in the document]
CarsonF: I wonder if we need a progress indicator. Or maybe we just use an `optimisticResponse` to set it immediately.

src/components/ProjectPinButton/ProjectPinButton.tsx
CarsonF: Oh we probably need to add the project to the pinned list if we can find it. I can help with that later. I'm working on the list cache right now.

src/scenes/Projects/List/ProjectList.tsx
```diff
@@ -25,12 +25,16 @@ const useStyles = makeStyles(({ spacing }) => ({
...
+  projectTotal: {
+    margin: spacing(2, 2),
+  },
 }));
 
 export const ProjectList: FC = () => {
   const formatNumber = useNumberFormatter();
   const sort = useSort<Project>();
   const [filters, setFilters] = useProjectFilters();
+  const [activeTab, setActiveTab] = useState('mine');
```
CarsonF, line 37: This should be stored in the `useProjectFilters` state. This way page reloads & navigation bring users back to the same tab. We also define the filters so they can be passed straight to API filters.
So we need to add a `pinned` property. Maybe it and `mine` can share a `tab` enum query param?

The mine toggle should also be removed from `<ProjectFilterOptions />`

src/scenes/Projects/List/ProjectList.tsx
```diff
@@ -60,22 +97,53 @@ export const ProjectList: FC = () => {
...
-      </ListContainer>
+      <TabContext value={activeTab}>
+        <Tabs
+          value={activeTab}
+          onChange={(_e, newActiveTab) => setActiveTab(newActiveTab)}
+          aria-label="project navigation tabs"
+        >
+          <Tab label="My Projects" key="mine" value="mine" />
+          <Tab label="All" key="all" value="all" />
+          <Tab label="Recent" key="recent" value="recent" />
```
CarsonF, line 108: Recent was only in designs. Don't add it here as we don't have an implementation.

src/scenes/Projects/List/ProjectList.tsx
```diff
@@ -40,13 +44,46 @@ export const ProjectList: FC = () => {
...
+    switch (activeTab) {
+      case 'mine':
+      case 'all':
+        return formatNumber(data?.projects.total);
+      case 'pinned':
+        return formatNumber(pinnedProjects?.length);
+      default:
+        return 0;
+    }
+  };
```
CarsonF, line 78: `pinned` needs to be sent in the API filters, so they API can give us the pinned ones and allow us to pagination through them. With that filter the total will be correct so it can come straight from the API result.

src/scenes/Projects/List/ProjectList.tsx
```diff
@@ -60,22 +97,53 @@ export const ProjectList: FC = () => {
...
+          <Tab label="All" key="all" value="all" />
+          <Tab label="Recent" key="recent" value="recent" />
+          <Tab label="Pinned" key="pinned" value="pinned" />
+        </Tabs>
+        <Typography variant="h3" className={classes.projectTotal}>
+          {data ? `${getProjectTotal()} Projects` : <Skeleton width="12ch" />}
+        </Typography>
+        <ListContainer>
+          <TabPanel value="mine">
+            {listOrPlaceholders(data?.projects.items, 5).map((item, index) => (
```
CarsonF, line 116: Not sure about this one off the top of my head. But all tabs will come from the same list query, so I don't think there's a need to specify all of them. Or maybe any of them?

CarsonF (changes requested):

src/components/ProjectPinButton/ProjectPinButton.tsx
zkhin:
```
cache.identify(props.object)
```
It returns `undefined` value because we don't cache `TogglePinFragment ` yet.
FYI, I passed `project` in this component, and `cache.identify(props.project)` will return correct id and it updates the pinned field correctly.

zkhin: Aha, I can see what you meant. NVM

src/scenes/Projects/List/ProjectList.tsx
zkhin: If we pin a project in the `All` tab, it should be added immediately in the `Pinned` tab. Since we don't re-fetch API after toggling the project,  we cannot get the newly added pinned project. So I filtered pinned projects from all projects list like below.
```
const pinnedProjects = useMemo(
    () => data?.projects.items.filter((project) => project.pinned),
    [data?.projects.items]
);
```
FYI, we can get the correct pinned projects and the total number if we add the `pinned` filter in the API filters.

CarsonF: Filtering client side won't work because the client doesn't have all the information. A user could have a pinned project that's on page 5. Sourcing the pinned list from the all list wouldn't grab that project if only the first 4 pages are loaded.

I've got a branch to help add items to lists in cache. Let's just ignore this problem until that branch lands.

zkhin: Thanks

sethmcknight:
FE for seedcompany/cord-api-v3#1721

CarsonF:
# List Before
Screen Shot 2021-03-30 at 3 41 20 PM

# List After

Screen Shot 2021-03-30 at 3 40 42 PM
Screen Shot 2021-03-30 at 3 40 49 PM

# Details Before 

Screen Shot 2021-03-30 at 3 41 48 PM
Screen Shot 2021-03-30 at 3 41 31 PM

# Details After 
Screen Shot 2021-03-30 at 3 40 26 PM
Screen Shot 2021-03-30 at 3 40 30 PM

PR #785, opened 2021-04-05
First rev of project comments

richardgirges:
Two outstanding things that need to be figured out:

1. The `List` component currently doesn't take into account sub-resolver lists (e.g. lists being loaded from`Project.posts.items` as opposed to a root-level query like `projects`). I'm using the `List` component in `ProjectOverview.tsx` L552. A TS error is being thrown, because the props passed into the `List` component don't contain the typical root-level list query props (e.g. `loading`, `loadMore`, etc)

2. I'm not sure how to use `addItemToList` to update the `Project.posts.items` array or if this is a use-case that has been considered in the existing implementation of `addItemToList`

### DIFF
--- a/src/scenes/Projects/Overview/ProjectOverview.graphql
+++ b/src/scenes/Projects/Overview/ProjectOverview.graphql
@@ -78,6 +78,17 @@ fragment ProjectOverview on Project {
   }
 }
 
+query ProjectPostListOverview($project: ID!, $input: PostListInput) {
+  project(id: $project) {
+    id
+    ... on Postable {
+      posts(input: $input) {
+        ...PostList
+      }
+    }
+  }
+}
+
 query ProjectEngagementListOverview(
   $project: ID!
   $input: EngagementListInput
--- a/src/scenes/Projects/Overview/ProjectOverview.tsx
+++ b/src/scenes/Projects/Overview/ProjectOverview.tsx
@@ -43,6 +43,7 @@ import {
   ProjectOverviewDocument,
   ProjectOverviewFragment,
 } from './ProjectOverview.generated';
+import { ProjectPostList } from './ProjectPostList';
 
 const useStyles = makeStyles(({ spacing, breakpoints, palette }) => ({
   root: {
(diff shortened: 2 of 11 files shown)


src/scenes/Projects/Overview/ProjectOverview.graphql
```diff
@@ -76,6 +76,12 @@ fragment ProjectOverview on Project {
   partnerships {
     ...PartnershipSummary
   }
+
+  ... on Postable {
```
CarsonF, line 80: You shouldn't need to specify this. Only maybe I just fixed that in develop schema last night

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -514,6 +521,44 @@ export const ProjectOverview: FC = () => {
...
+            </Grid>
+          </Grid>
+          {showCommentField && (
+            <CreatePostForm
+              parentId={projectId}
+              onSubmit={() => {
+                toggleCommentField();
+
+                // FIXME: need to figure out how to do list cache-updates for nested lists
+                refetch().catch((err) => console.log('refetch error', err));
```
CarsonF, line 548: Like this https://github.com/SeedCompany/cord-field/blob/7a789fba4da5b209e25e838fc5f8bbb8f83a2deb/src/scenes/Projects/Members/Create/CreateProjectMember.tsx#L53-L58

src/scenes/Projects/Overview/ProjectOverview.graphql
CarsonF: Let's actually move the comments list to another component & query. This will
- Not further add to the clutter of ProjectOverview
- Allows us to use the `useListQuery` hook which List wants and allows for pagination
- Data can load in parallel

CarsonF (changes requested):

src/scenes/Projects/Overview/ProjectOverview.graphql
richardgirges: I'm still not able to get it to work without `... on Postable` - I just merged latest updates from develop today.

richardgirges: Moved the post list into a separate component and query 👍

PR #746, opened 2021-03-03
Fix undefined step on project for workflow project update

blakekrammes:
Co-authored-by: Carson Full <carsonfull@gmail.com>

PR #799, opened 2021-04-08
Check if engagement exists on engagement workflow submit

blakekrammes:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/1793


CarsonF (changes requested):
Change the form values type to reflect that this engagement property is nullable instead of suppressing the linter.
Just like we did in #746

CarsonF (approved):

Issue #753, opened 2021-03-08
Partnership Card - Primary Partner


sethmcknight:
Add a field to define which Partnership is the 'Primary Partner' (there can only be 1 per Project)

sethmcknight:
@zkhin you're good to move forward with what you have on this issue, we won't have designs for this now.

zkhin:
@CarsonF Can we close this ticket or do we need to wait for the design?

CarsonF:
Up to @sethmcknight

sethmcknight:
@zkhin Just checked this out in Dev, looks great! We can close this out. Thanks so much!

PR #810, opened 2021-04-15
Primary partnership card

zkhin:
https://github.com/SeedCompany/cord-field/issues/753
image
image

### DIFF
--- a/src/components/PartnershipCard/PartnershipPrimaryIcon.tsx
+++ b/src/components/PartnershipCard/PartnershipPrimaryIcon.tsx
@@ -1,5 +1,5 @@
 import { makeStyles, SvgIconProps, Tooltip } from '@material-ui/core';
-import { Check } from '@material-ui/icons';
+import { Star } from '@material-ui/icons';
 import clsx from 'clsx';
 import { FC } from 'react';
 import * as React from 'react';
@@ -8,18 +8,13 @@ const useStyles = makeStyles(({ palette }) => ({
   primary: {
     color: palette.primary.main,
   },
-  grey: {
-    color: palette.grey[400],
-  },
 }));
 
 export interface PartnershipPrimaryIconProps extends SvgIconProps {
-  value?: boolean;
   loading?: boolean;
 }
 
 export const PartnershipPrimaryIcon: FC<PartnershipPrimaryIconProps> = ({
-  value,
   loading,
   className,
   ...rest
@@ -27,14 +22,8 @@ export const PartnershipPrimaryIcon: FC<PartnershipPrimaryIconProps> = ({
   const classes = useStyles();
 
   return (
-    <Tooltip title={!loading && value ? 'Primary' : 'Not Primary'}>
-      <Check
-        className={clsx(
-          !loading && value ? classes.primary : classes.grey,
-          className
-        )}
-        {...rest}
-      />
+    <Tooltip title="Primary">
+      <Star className={clsx(classes.primary, className)} {...rest} />
     </Tooltip>
   );
 };
--- a/src/components/PartnershipCard/PartnershipPrimaryIcon.tsx
+++ b/src/components/PartnershipCard/PartnershipPrimaryIcon.tsx
@@ -4,18 +4,13 @@ import clsx from 'clsx';
 import { FC } from 'react';
 import * as React from 'react';
 
-const useStyles = makeStyles(({ palette }) => ({
+const useStyles = makeStyles(() => ({
   primary: {
-    color: palette.primary.main,
+    color: '#ffb400',
   },
 }));
 
-export interface PartnershipPrimaryIconProps extends SvgIconProps {
-  loading?: boolean;
-}
-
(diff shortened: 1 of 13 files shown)


src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
```diff
@@ -153,6 +154,11 @@ export const PartnershipForm = <T extends PartnershipFormValues>({
                 </SecuredField>
               </>
             )}
+            {values.partnership?.partnerLookupItem || partnership ? (
```
CarsonF, line 157: Shouldn't need this condition

CarsonF: Though maybe we should hide the switch if the partnership is already primary. We don't allow that on the backend so it would be better just to hide it here.

src/scenes/Partnerships/Create/CreatePartnership.tsx
```diff
@@ -43,6 +44,7 @@ export const CreatePartnership = ({
         (res) => res.createPartnership.partnership
       )
     ),
+    refetchQueries: [GQLOperations.Query.ProjectPartnerships],
```
CarsonF, line 47: I'm trying to avoid refetch queries. It's much faster to update the cache directly. We have a handful of mutation `update` functions across the app now. Take a look at those and take a stab at just finding an old primary partnership and setting it to false.

src/components/PartnershipPrimaryIcon/PartnershipPrimaryIcon.tsx
```diff
@@ -0,0 +1,40 @@
...
+    color: palette.grey[400],
+  },
+}));
+
+export interface PartnershipPrimaryIconProps extends SvgIconProps {
+  value?: boolean;
+  loading?: boolean;
+}
+
+export const PartnershipPrimaryIcon: FC<PartnershipPrimaryIconProps> = ({
```
CarsonF, line 21: The checkmark icon suggests there is something "to do" or "is done". How about a `Star` instead?
Maybe it could go to the right of the partner name as well.
Also I don't think showing the false state adds much value; I'd just omit it.
If you do end up keeping this component, I'd just move it inside of components/PartnershipCard as it's only a helper for that component.

CarsonF (changes requested):

src/components/PartnershipPrimaryIcon/PartnershipPrimaryIcon.tsx
zkhin: Good idea!

src/scenes/Partnerships/PartnershipForm/PartnershipForm.tsx
zkhin: By the way, we need this switch while creating partnership, right?

CarsonF: Yeah, so show if no partnership or partnership.primary is false.

CarsonF:
Need to add the primary property to that storybook type failure

zkhin:
> Need to add the primary property to that storybook type failure

Done!

src/scenes/Partnerships/Edit/InvalidateOldPrimaryPartnership.ts
```diff
@@ -0,0 +1,26 @@
+import { ApolloCache, MutationUpdaterFn } from '@apollo/client';
+import { ProjectPartnershipsQuery } from '../List/PartnershipList.generated';
+
+export const invalidateOldPrimaryPartnership = <R>(
+  project: ProjectPartnershipsQuery['project']
+): MutationUpdaterFn<R> => (cache: ApolloCache<unknown>) => {
+  const filtered = project.partnerships.items.filter(
```
CarsonF, line 7: Can you use `find` instead of `filter` here to simplify logic.

src/components/PartnershipCard/PartnershipPrimaryIcon.tsx
```diff
@@ -0,0 +1,29 @@
+import { makeStyles, SvgIconProps, Tooltip } from '@material-ui/core';
+import { Star } from '@material-ui/icons';
+import clsx from 'clsx';
+import { FC } from 'react';
+import * as React from 'react';
+
+const useStyles = makeStyles(({ palette }) => ({
+  primary: {
+    color: palette.primary.main,
```
CarsonF, line 9:
```suggestion
    color: '#ffb400',
```

src/components/PartnershipCard/PartnershipPrimaryIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
+  },
+}));
+
+export interface PartnershipPrimaryIconProps extends SvgIconProps {
+  loading?: boolean;
```
CarsonF, line 14:
```suggestion
```

CarsonF (changes requested):

CarsonF (approved):
Looks great! Thanks @zkhin!

PR #820, opened 2021-04-20
#812 Defined File Cleanup

### LINKED ISSUE #812: Defined File Cleanup
* Move `AddItemCard` into `DefinedFileCard`
* Updated `DefinedFileCard` to handle drag & drop & upload internally. It can be passed a mutation document to facility this.
  It might need a couple other things too.
* Remove refetch queries from file mutations. We should be able to update the cache directly. We have helpers like `addItemToList` and `removeItemFromList` to help with that.
* `updateCachedVersions` is a current helper to update cache. It uses `cache.read/writeQuery()` which takes a query document. We can instead use `read/writeFragment()` which doesn't need to specifically know where the cached file comes from. [Here's an example of that](https://github.com/SeedCompany/cord-field/blob/2090c35e0f7748e1fc1809226938cac0a87fcb42/src/scenes/Projects/Budget/ProjectBudgetRecords.tsx#L36-L53), but this should only be used if `addItemToList` is not sufficient.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466731616) by [Unito](https://www.unito.io)

tmpacifitech:
Closes #812


src/scenes/Projects/Files/useUploadProjectFiles.ts
```diff
@@ -28,18 +26,8 @@ export const useUploadProjectFiles = (): UploadFilesConsumerFunction => {
     };
     await createFileVersion({
       variables: { input },
-      refetchQueries:
-        action === 'file' ? [GQLOperations.Query.ProjectDirectory] : undefined,
       update: (cache, { data }) => {
```
CarsonF, line 29: This should be able to use `addItemToList` as well

src/scenes/Projects/Files/useUploadProjectFiles.ts
```diff
@@ -92,29 +80,12 @@ export const useUploadBudgetFile = (): UploadFilesConsumerFunction => {
     uploadId,
     name,
     parentId: id,
-    action,
   }) => {
     await uploadFile({
       variables: {
         id,
         universalTemplateFile: { uploadId, name },
       },
```
CarsonF, line 88: This hook should be able to go away. Can't we just pass the mutation document to the DefinedFileCard?

src/scenes/Engagement/Files/useUploadEngagementFiles.ts
```diff
@@ -15,13 +14,6 @@ import {
...
 type EngagementType = 'language' | 'internship';
 
-const isLanguageEngagementMutation = (
-  data:
-    | UploadInternshipEngagementGrowthPlanMutation
-    | UploadLanguageEngagementPnpMutation
-): data is UploadLanguageEngagementPnpMutation =>
-  'updateLanguageEngagement' in data;
-
 export const useUploadEngagementFile = (
```
CarsonF, line 17: This hook should go away. Can't we just pass the mutation to the defined file card? If not we should have separate hooks dedicated to pnp & growthPlan.

src/components/DefinedFileCard/DefinedFileCard.tsx
```diff
@@ -100,29 +124,42 @@ const FileCardMeta: FC<FileCardMetaProps> = ({
 export const DefinedFileCard = (props: DefinedFileCardProps) => {
   const classes = useStyles();
   const { title, onVersionUpload, resourceType, securedFile } = props;
+  const { value: file, canRead, canEdit } = securedFile;
+
+  const { openFilePreview } = useFileActions();
 
-  const { canRead, canEdit } = securedFile;
-  const standardFileActions = getPermittedFileActions(canRead, canEdit);
+  const standardFileActions = useMemo(
```
CarsonF, line 131: My understanding of useMemo is it's not worth for general operations / jsx rendering. Like in this case react as to store and determine if `canEdit` and `canRead` on every render. It would be faster just to create the list again.

The only place I've really found a use for them is ensuring lists & objects identities don't change. So I see here you've probably got all of these memo-ed because they funnel into main card memo. I think my first point still standards for that as well.
Thoughts?

CarsonF (changes requested):

src/scenes/Projects/Files/CreateProjectFile.graphql
```diff
@@ -2,5 +2,11 @@ mutation CreateProjectDirectory($input: CreateDirectoryInput!) {
   createDirectory(input: $input) {
     id
     name
+    type
+    createdAt
+    createdBy {
+      id
+      fullName
+    }
```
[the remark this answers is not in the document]
tmpacifitech: image
It's used for Uploaded By column.

tmpacifitech:
@CarsonF  Updated the PR as per your comments!

PR #824, opened 2021-04-23
Fix product crash

elisabeth-sorrell:
add a sanity check in secured field for just in case the field is undefined.

### DIFF
--- a/src/components/form/SecuredField.tsx
+++ b/src/components/form/SecuredField.tsx
@@ -49,15 +49,18 @@ export const SecuredField = <
   if (obj) {
     // @ts-expect-error Grab key from object following naming convention. We check below that we grabbed it correctly.
     const field: SecuredProp<any> = obj[name] ?? obj[name.replace(/Id$/, '')];
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- sanity check since we ignored types above
-    if (field.canRead == null || field.canEdit == null) {
-      console.error(
-        `Cannot determine if field should be readable/editable: ${obj.__typename}.${name}`
-      );
-      return null;
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- sanity check
+    if (field) {
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- sanity check since we ignored types above
+      if (field.canRead == null || field.canEdit == null) {
+        console.error(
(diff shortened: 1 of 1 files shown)


src/components/form/SecuredField.tsx
```diff
@@ -49,15 +49,18 @@ export const SecuredField = <
   if (obj) {
     // @ts-expect-error Grab key from object following naming convention. We check below that we grabbed it correctly.
     const field: SecuredProp<any> = obj[name] ?? obj[name.replace(/Id$/, '')];
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- sanity check since we ignored types above
-    if (field.canRead == null || field.canEdit == null) {
-      console.error(
-        `Cannot determine if field should be readable/editable: ${obj.__typename}.${name}`
-      );
-      return null;
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- sanity check
```
CarsonF, line 52: Add `| undefined` to the `field` declaration so the linter knows it's a necessary condition.

src/components/form/SecuredField.tsx
```diff
@@ -49,15 +49,18 @@ export const SecuredField = <
...
+    if (field) {
```
CarsonF, line 53: This could be merged into the if statement below.
```tsx
if (!field || field.canRead == null...) {
```

CarsonF (changes requested):
This is a good start as it fixes the crash. But really the need the field to be matched. So that needs to be solved as well. i.e. Determine what field we are trying to show (and was previously crashing) and fix that. The error message logged should be a good starting point. This is fixed when that console error doesn't happen.

src/components/form/SecuredField.tsx
elisabeth-sorrell: I don't think so, because it's setting canRead and canEdit a few linkes outside of the if statement

CarsonF: Yes because of the return null statement

elisabeth-sorrell: oops. I get it now. because it can return null when I check for undefined. never mind! My bad.

CarsonF (approved):
I agree that we should get this crash fix out. I believe that this field not showing is actually the intended functionality, so out showing it is ok. We can ignore the error for now.

PR #853, opened 2021-05-26
Fix Monthly List Order & Change timing of report card

tmpacifitech:
https://github.com/SeedCompany/cord-api-v3/issues/1896
https://github.com/SeedCompany/cord-api-v3/issues/1897
https://github.com/SeedCompany/cord-field/issues/860


CarsonF:
Thanks @tmpacifitech!

PR #857, opened 2021-05-28
change default sort order for posts

richardgirges:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/1893


CarsonF:
Doing in API instead https://github.com/SeedCompany/cord-api-v3/pull/1918

PR #889, opened 2021-07-13
Edit and delete functionality for posts

richardgirges:
All project posts that belong to currently-logged-in user will have a context menu that will pop open edit and delete options.


src/components/posts/PostListItemCard/PostListItemCard.tsx
```diff
@@ -57,45 +69,83 @@ const useStyles = makeStyles(({ spacing, typography }) => {
...
+        }}
+      />
+      <EditPost post={props} open={edit} onClose={() => setEdit(false)} />
+      <DeletePost
+        open={deletePost}
+        project={props.project}
+        post={{
+          __typename: 'Post',
+          id: props.id,
+        }}
```
CarsonF, line 146: I'm trying to avoid doing this. Makes it harder when the type changes (though this is a concrete) and harder when more info is needed, like changeset ID. 



```suggestion
        post={props}
```

Or even better
```suggestion
        post={props.post}
```
I don't think it really helps to have the post props at the root level. It makes it harder to isolate them and makes JS object identity checks harder because the object is always changing with the spread.

src/components/posts/PostListItemCard/PostListItemCard.tsx
```diff
@@ -57,45 +69,83 @@ const useStyles = makeStyles(({ spacing, typography }) => {
...
+                  {`${props.creator.value?.displayFirstName.value} ${props.creator.value?.displayLastName.value}`}
+                </Typography>
+                <Typography variant="body2" color="textSecondary">
+                  {formatDate(props.createdAt)}
+                </Typography>
+              </div>
+              {props.canEdit && (
+                <IconButton onClick={(e) => setActionsAnchor(e.currentTarget)}>
+                  <MoreVert />
+                </IconButton>
```
CarsonF, line 105: I think it would look better if this was made square to the card
Screen Shot 2021-07-19 at 10 49 39 AM
Screen Shot 2021-07-19 at 11 02 44 AM

src/components/posts/PostForm/PostForm.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { minLength, required } from '../../form/validators';
+import { PostFormFragment } from './PostForm.generated';
+
+export type PostFormProps<T, R = void> = DialogFormProps<T, R> & {
+  /** The pre-existing user to edit */
+  post?: PostFormFragment;
+  prefix: string;
+};
+
+export const PostForm = <T, R = void>({
```
CarsonF, line 28: Isn't it weird that edit post is in a dialog but create is inline in the page?

src/components/posts/PostForm/PostForm.tsx
```diff
@@ -0,0 +1,74 @@
...
+  SecuredField,
+  SelectField,
+  SubmitError,
+  TextField,
+} from '../../../components/form';
+import { minLength, required } from '../../form/validators';
+import { PostFormFragment } from './PostForm.generated';
+
+export type PostFormProps<T, R = void> = DialogFormProps<T, R> & {
+  /** The pre-existing user to edit */
```
CarsonF, line 23: "post" you mean?

src/components/posts/PostForm/PostForm.tsx
```diff
@@ -0,0 +1,74 @@
...
+  post?: PostFormFragment;
+  prefix: string;
```
CarsonF, line 25: Shouldn't prefix be static to the specific form?

Also DialogForm has a fieldsPrefix prop that could be used instead of the <FieldGroup> component.

CarsonF (changes requested):
Thanks for your work @richardgirges, it is shaping up nicely!

CarsonF (approved):

PR #892, opened 2021-07-15
Rich text posts

richardgirges:
This is the first draft of rich text editing with CK Editor.

Things to note:
* This PR contains PR #889 and should be rebased after 889 is merged

* There are no typings for the React CKEditor wrapper. The type check is failing right now because of this. [Lack of typings for CKEditor is a long-standing issue with no solid resolution as of yet](https://github.com/ckeditor/ckeditor5/issues/504). Please advise on how to approach

### DIFF
--- a/src/components/posts/PostListItemCard/PostListItemCard.tsx
+++ b/src/components/posts/PostListItemCard/PostListItemCard.tsx
@@ -73,6 +73,9 @@ const useStyles = makeStyles(({ spacing, typography }) => {
       right: 0,
       top: 0,
     },
+    postBody: {
+      marginRight: spacing(2),
+    },
   };
 });
 
@@ -127,7 +130,7 @@ export const PostListItemCard: FC<PostListItemCardProps> = ({
                 </Typography>
               </div>
             </div>
-            <Typography variant="body2">{postBody}</Typography>
+            <div className={classes.postBody}>{postBody}</div>
           </div>
         </CardContent>
         {editable && (
--- a/src/components/posts/PostListItemCard/PostListItemCard.tsx
+++ b/src/components/posts/PostListItemCard/PostListItemCard.tsx
@@ -9,7 +9,8 @@ import {
 } from '@material-ui/core';
 import { MoreVert } from '@material-ui/icons';
 import clsx from 'clsx';
-import { FC, useState } from 'react';
+import parse from 'html-react-parser';
+import { FC, useMemo, useState } from 'react';
 import * as React from 'react';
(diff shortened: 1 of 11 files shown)


CarsonF:
> * There are no typings for the React CKEditor wrapper. The type check is failing right now because of this. [Lack of typings for CKEditor is a long-standing issue with no solid resolution as of yet](https://github.com/ckeditor/ckeditor5/issues/504). Please advise on how to approach


We've got a few libraries we provide our own types for in the `typings` folder. It looks like there are types in DefinitelyTyped for the core of the library, so really we just need to provide connection points to this library. Should be trivial since both referenced are thin wrappers.

src/components/posts/PostListItemCard/PostListItemCard.tsx
```diff
@@ -118,7 +119,9 @@ export const PostListItemCard: FC<PostListItemCardProps> = (props) => {
                 </Typography>
               </div>
             </div>
-            <Typography variant="body2">{props.body.value}</Typography>
+            <Typography variant="body2">
+              {parse(props.body.value || '')}
```
CarsonF, line 123: We should probably memoize this, eh?

CarsonF (approved):
Looks good @richardgirges!

src/components/posts/PostListItemCard/PostListItemCard.tsx
```diff
@@ -124,7 +127,7 @@ export const PostListItemCard: FC<PostListItemCardProps> = ({
                 </Typography>
               </div>
             </div>
-            <Typography variant="body2">{post.body.value}</Typography>
+            <Typography variant="body2">{postBody}</Typography>
```
CarsonF, line 130: Currently this puts any html tags under a `<p>` tag which is invalid HTML.

CarsonF (changes requested):
The CKE features should be limited to only those we can render. I'd remove quotes, tables, pics, and video.

We should style the output with `react-html-parser` and the display inside of the editor to be consistent with our theme.

ParkeBrown:
@daniel-leod can you confirm that we still need to address the unresolved comment above?

daniel-leod:
> @daniel-leod can you confirm that we still need to address the unresolved comment above?

@ParkeBrown I think the last commit would resolve that. This PR is ready for review.

ParkeBrown:
Daniel believes he's covered the requested changes

CarsonF:
I'd like to go with an editor that produces JSON instead, instead of html string. So rendering can be easily customized based on context.

Issue #722, opened 2021-02-11
Person Details | Assign & Update Roles


ParkeBrown:
Administrators should be able to add, change and remove roles from a Person on the People Details page.

Leverage the existing UI for adding/editing the roles of a Team Member.

<img width="444" alt="Screen Shot 2021-07-09 at 12 43 33 PM" src="https://user-images.githubusercontent.com/7377802/125122920-49fcbc80-e0b3-11eb-8194-d37c42cac9f1.png">

CarsonF:
Cherry picked to master. Going live now.

PR #893, opened 2021-07-15
User.roles form field

tmpacifitech:
https://github.com/SeedCompany/cord-field/issues/722

BE changes: https://github.com/SeedCompany/cord-api-v3/pull/2028


CarsonF:
The issue is for users not team members.

CarsonF (changes requested):
Use user.roles.canEdit to determine if the user can change the roles of the user.

This is completely separate from team members.

CarsonF (approved):

PR #901, opened 2021-07-24
Periodic Reports | Add received date to periodic report

blakekrammes:
API PR: https://github.com/SeedCompany/cord-api-v3/pull/1972


src/components/PeriodicReports/Upload/useUploadPeriodicReport.ts
```diff
@@ -1,33 +1,45 @@
...
         },
-      },
-    });
-  };
-
-  return (files: File[], reportIdOverride?: string) => {
-    const id = reportIdOverride ?? reportId;
+      });
+    };
     if (id) {
```
blakekrammes, line 31: We needed to be able to update a report with just the received date, so I made the reportId param required since it can't be derived without a file input param. I'm not sure what the function of the report id override was and if we need to maintain that.

CarsonF (approved):

PR #914, opened 2021-08-03
Show final report on reports list

tmpacifitech:
https://github.com/SeedCompany/cord-api-v3/pull/2047

### DIFF
--- a/src/components/PeriodicReports/PeriodicReportsTable.tsx
+++ b/src/components/PeriodicReports/PeriodicReportsTable.tsx
@@ -55,6 +55,8 @@ export const PeriodicReportsTableInContext = ({
       title: 'Period',
       defaultSort: 'desc',
       render: ({ report }) => <ReportLabel report={report} />,
+      customSort: (row1, row2) =>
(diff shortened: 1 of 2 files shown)


src/components/PeriodicReports/PeriodicReportsTable.tsx
```diff
@@ -55,6 +55,8 @@ export const PeriodicReportsTableInContext = ({
       title: 'Period',
       defaultSort: 'desc',
       render: ({ report }) => <ReportLabel report={report} />,
+      customSort: (row1, row2) =>
+        row1.period.toMillis() - row2.period.toMillis(),
```
CarsonF, line 59: I don't think this is necessary..

tmpacifitech: Because of the custom render method which returns react element instead of raw string, the table wasnt able to be sorted by this column. That's why I had to add a new custom sort.

CarsonF: Right of course 👍

CarsonF (approved):
🎉

PR #915, opened 2021-08-03
Add placeholder text to verses field

blakekrammes:
Text taken from Figma example. 

Screen Shot 2021-08-03 at 12 36 20 PM


CarsonF (commented):
Can we add helper text to hint at pressing enter to add selection?

blakekrammes:
So text that's there before anything has been entered?

CarsonF:
No it's not the placeholder it's below the input. 
https://github.com/SeedCompany/cord-field/blob/63221be5bfe822c18a399ef9a24607ba330e0c81/src/scenes/Engagement/EditEngagement/EditEngagementDialog.tsx#L75-L75

blakekrammes:
Right right, but it will only be there before the user enters anything. Or should it always be there until an error occurs?

CarsonF:
It can always be there unless there's an error, which is exactly what's coded now

CarsonF:
Thanks Blake!

PR #961, opened 2021-10-12
Add progress measurement and target

tmpacifitech:
Fixes https://github.com/SeedCompany/cord-field/issues/925


CarsonF (approved):
Thanks @tmpacifitech!

PR #995, opened 2021-11-01
Change progress measurement from Boolean to Done

daniel-leod:
Closes [#2245](https://github.com/SeedCompany/cord-api-v3/issues/2245)


CarsonF (changes requested):
Just want to change the label in the UI

CarsonF:
You should only need to modify the `displayProgressMeasurement` function

PR #998, opened 2021-11-02
Auto-select the single step of Completed for any Goal of type Other

### LINKED ISSUE #992: G&P | Auto-select the single step of `Completed` for any Goal of type `Other`

daniel-leod:
Closes #992


CarsonF (changes requested):
This only affects what's shown in this collapsed UI. We need to change the actual form value so it applies everywhere. This is done with "decorators" [here](https://github.com/SeedCompany/cord-field/blob/88acf714af8c82a361bd9081ae2563ca265a0ee0/src/scenes/Products/ProductForm/ProductForm.tsx#L59-L59).

CarsonF:
Looks like this has already been completed by #1003

PR #1024, opened 2021-11-19
Fix auto select goals available steps

### LINKED ISSUE #1019: Project > Goals > "Steps" are not changed after clicking on "Visual-other" methodology
**Roles:** All

**Describe the bug**
Project > Goals > `Steps` are not changed after clicking on `Visual-other` methodology

**Steps to Reproduce**
1.  Login as a root user
2.  Create a project and open a project detail page
3.  Add `LanguageEngagement` and click on the `View Details`
4.  Click on the `Add Goal` button and click on the `methodology` dropdown
5.  Select `Visual - film` methodology
6.  Select `Visual - Sign Language` methodology
7.  Select `Visual - Other` methodology
8.  Select `Visual - film` methodology

**Expected Result**
`Steps` should be changed for `Visual- Sign Language` &  `Visual- film` methodology type after clicking on  `Visual-other` methodology

**Actual Result**
`Steps` are not changed after clicking on `Visual-other` methodology

See video - 

https://user-images.githubusercontent.com/5212585/142194414-edfce8b1-1679-4ffd-951c-586c7756b910.mp4

**FYI:** @ParkeBrown @michaelmarshall 

cord-field# 7b2519a

daniel-leod:
Fixes #1019

### DIFF
--- a/src/scenes/Products/ProductForm/StepsSection.tsx
+++ b/src/scenes/Products/ProductForm/StepsSection.tsx
@@ -1,8 +1,7 @@
 import { useQuery } from '@apollo/client';
 import { ToggleButton } from '@material-ui/lab';
-import { intersection } from 'lodash';
 import React, { useEffect } from 'react';
-import { displayProductStep, ProductStep } from '../../../api';
+import { displayProductStep } from '../../../api';
 import { EnumField } from '../../../components/form';
 import { AvailableProductStepsDocument as AvailableSteps } from './ProductForm.generated';
 import { SectionProps } from './ProductFormFields';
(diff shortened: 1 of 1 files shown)


CarsonF:
Isn't this the same as https://github.com/SeedCompany/cord-field/pull/998?

daniel-leod:
> Isn't this the same as #998?

@CarsonF No, this fixes available steps auto selection when changing methodologies in visual approach.

CarsonF (approved):
I think the idea previously was that we were just trying to limit the steps selected to those now available, but defaulting to all available wasn't a thing. So we just wanted to keep what was selected by the user.

I see where this flow is confusing though if the user goes back and changes the product type. And I think that specifically could be considered "starting over" with the product. I agree this makes more sense over all - If the type is changed then the steps selected are reset, regardless of what the user has previously done.

Thanks @daniel-leod

PR #1034, opened 2021-11-24
Add posts to language details page

### LINKED ISSUE #946: Posts | Add Posts to the Language Details page
This will follow the same implementation as Posts on Projects do. The only difference is the only "shareabilities" are:
- External
- Ask to Share Externally
- Internal

daniel-leod:
Closes #946

### DIFF
--- a/src/scenes/Languages/Detail/LanguageDetail.graphql
+++ b/src/scenes/Languages/Detail/LanguageDetail.graphql
@@ -76,3 +76,14 @@ mutation RemoveLocationFromLanguage($languageId: ID!, $locationId: ID!) {
     ...LanguageForm
   }
 }
+
+query LanguagePostListOverview($language: ID!, $input: PostListInput) {
+  language(id: $language) {
+    id
+    ... on Postable {
+      posts(input: $input) {
+        ...PostList
+      }
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


src/scenes/Languages/Detail/LanguageDetail.graphql
```diff
@@ -75,3 +75,14 @@ mutation RemoveLocationFromLanguage($languageId: ID!, $locationId: ID!) {
...
 }
+
+query LanguagePostListOverview($language: ID!, $input: PostListInput) {
+  language(id: $language) {
+    id
+    ... on Postable {
+      posts(input: $input) {
+        ...PostList
+      }
+    }
```
CarsonF, line 86:
```suggestion
    posts(input: $input) {
      ...PostList
    }
```
The specific call out on the interface isn't needed since `Language` implements `Postable`.

This was previously needed on the project query because `Project` interface didn't implement `Postable`.
I fixed the API yesterday to have that, so it's no longer needed on that query now either.

CarsonF (changes requested):
Same comments as on other PR. Assuming this can be rebased after the other one is merged.

CarsonF (approved):

PR #1038, opened 2021-11-24
Tweak goals section

### LINKED ISSUE #1035: Project > Goal > Previously created goal name is displayed in "Other" goal type button
**Roles:** All

**Describe the bug**
Project > Goal > Previously created goal name is displayed in `Other` goal type button

**Steps to Reproduce**
1.  Login as a root user
2. Create a project and add language engagement
3. Create a goal with `ethnoArt` goal type
4. Click on the `Other` goal type and collapse the dropdown

**Expected Result**
Only the `Other` goal type should be displayed

**Actual Result**
Previously created goal name is displayed in `Other` goal type button

See video - 

https://user-images.githubusercontent.com/5212585/143193816-d2a7ee36-14d1-4a5c-aa79-c8884e64d1f9.mp4

***FYI:*** @ParkeBrown @michaelmarshall 

cord-field# 5c78c6f

daniel-leod:
Fixes #1035

### DIFF
--- a/src/scenes/Products/ProductForm/GoalsSection.tsx
+++ b/src/scenes/Products/ProductForm/GoalsSection.tsx
@@ -25,9 +25,7 @@ export const GoalsSection = ({
           {productType && (
             <ToggleButton selected value={produces || ''}>
               {`${displayProductTypes(productType)} ${
-                (productType !== 'DirectScriptureProduct' &&
-                  produces?.name.value) ||
-                ''
+                produces?.name.value || ''
               }`}
             </ToggleButton>
           )}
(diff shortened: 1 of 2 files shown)


CarsonF (commented):
Maybe it would be better to clear the `produces` field if the type is changed to other or scripture? Or maybe it doesn't matter because the submit handler already ignores produces when it needs to.

CarsonF (commented):
Ok well now this is a side effect, so it needs to be in an `useEffect` hook 😬

CarsonF (approved):

PR #1043, opened 2021-12-01
Add language pin feature

tmpacifitech:
API: https://github.com/SeedCompany/cord-api-v3/pull/2302

Fixes: https://github.com/SeedCompany/cord-api-v3/issues/2276

### DIFF
--- a/src/components/LanguageListItemCard/LanguageListItem.graphql
+++ b/src/components/LanguageListItemCard/LanguageListItem.graphql
@@ -21,4 +21,5 @@ fragment LanguageListItem on Language {
   presetInventory {
     value
   }
+  ...TogglePin
 }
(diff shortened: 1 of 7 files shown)


CarsonF (changes requested):
https://user-images.githubusercontent.com/932566/144680597-4e2c4f6d-3276-442e-a312-c0e756fef2e7.mov

Toggle appears to do nothing. Also can we get some spacing in-between the two icons

tmpacifitech:
> Screen.Recording.2021-12-03.at.4.21.09.PM.mov 
> Toggle appears to do nothing. Also can we get some spacing in-between the two icons

@CarsonF This works fine on my side. I just tested again with the latest api develop branch.

Issue #1022, opened 2021-11-18
G&P | Show on the Progress Reports table that a specific Step for a Goal was not planned


sethmcknight:
Use an en dash for cells that are not planned

PR #1053, opened 2021-12-16
Use dash for not planned cells

tmpacifitech:
Fixes https://github.com/SeedCompany/cord-field/issues/1022


CarsonF (changes requested):
Not all cells missing a value; only for cells whose columns aren't planned steps.

tmpacifitech:
Updated PR.

CarsonF (approved):
Thanks @tmpacifitech!

Issue #1010, opened 2021-11-09
G&P | Product form validations errors are not shown in collapsed accordion state


iPiyushPatel:
**Roles:** All

**Describe the bug**
Project > Goals > Ethno Art > Required field message is not displayed at front of the UI when user leave blank `Progress Target` field

**Steps to Reproduce**
1.  Login as a root user
2.  Create a project and open a project detail page
3.  Add `LanguageEngagement` and click on the `View Details`
4.  Click on the `Add Goal` button and select `Ethno Art` from the goal dropdown
5.  Add other details excluding `Progress Target` and click on the `Save Goal` button

**Expected Result**
Required field message should be displayed somewhere on the top or bottom of the form along with missing field name.

**Actual Result**
Required field message is not displayed at front of the UI when user leave blank `Progress Target` field

See video - 

https://user-images.githubusercontent.com/5212585/140894563-272e9e1a-5444-497f-93e7-ab270e656ca3.mp4

**FYI:** @ParkeBrown @michaelmarshall 

cord-field# d8b9268

PR #1056, opened 2021-12-21
Open progress target whilte its not validated

tmpacifitech:
Fixes https://github.com/SeedCompany/cord-field/issues/1010


CarsonF (commented):
Can this be done in a more generic way? I think this use case also applies to `produces` field

CarsonF (changes requested):
`props.errors` could be passed here
https://github.com/SeedCompany/cord-field/blob/f16352850899cb5c0e50421d494978a427ce0e7f/src/scenes/Products/ProductForm/ProductFormFields.tsx#L81
Then `DefaultAccordion` / `SecuredAccordion` could check this prop with the name field to determine if the accordion is an error state.
This would transparently apply to all (most?) the form accordions.

CarsonF (approved):

Issue #1015, opened 2021-11-11
Develop "Pin" Animation


AmyRickwartz:

AmyRickwartz:
Designs can be found here:
[FOR DEVELOPMENT - Cord Field > Projects - Favor Subscribe](https://www.figma.com/file/IfEmcJeuNY6ucgjYctlKB7/FOR-DEVELOPMENT---Cord-Field?node-id=4623%3A786)

When I designed this I didn't use the "ripple" effect that is live on CF currently. I need to update this file sightly to reflect that. However, if development picks this up in the mean time—the animation action is moving the "pin" -45° to show if an item has been pinned.

ParkeBrown:
@sethmcknight FYI

tmpacifitech:
> Designs can be found here: [FOR DEVELOPMENT - Cord Field > Projects - Favor Subscribe](https://www.figma.com/file/IfEmcJeuNY6ucgjYctlKB7/FOR-DEVELOPMENT---Cord-Field?node-id=4623%3A786)
> 
> When I designed this I didn't use the "ripple" effect that is live on CF currently. I need to update this file sightly to reflect that. However, if development picks this up in the mean time—the animation action is moving the "pin" -45° to show if an item has been pinned.

@AmyRickwartz I need permission to access this figma file.

AmyRickwartz:
Hi Thain @tmpacifitech ,
The Figma file says you have viewing access. Perhaps you are not using the same email that you had used previously in Figma? I can’t see what email it says you are registered with…I had the issue with Carson once.

PR #1062, opened 2021-12-23
Add pin icon animation

tmpacifitech:
Fixes https://github.com/SeedCompany/cord-field/issues/1015

### DIFF
--- a/src/components/TogglePinButton/TogglePinButton.tsx
+++ b/src/components/TogglePinButton/TogglePinButton.tsx
@@ -1,5 +1,6 @@
 import { useMutation } from '@apollo/client';
 import { makeStyles, Theme, Tooltip, TooltipProps } from '@material-ui/core';
+import clsx from 'clsx';
 import * as React from 'react';
 import { Except } from 'type-fest';
 import { addItemToList, removeItemFromList } from '../../api';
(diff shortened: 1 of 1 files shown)


CarsonF (changes requested):
Position is messed up in list
Screen Shot 2021-12-29 at 3 02 32 PM

src/components/TogglePinButton/TogglePinButton.tsx
```diff
@@ -11,6 +11,15 @@ import {
   TogglePinnedDocument,
 } from './TogglePinButton.generated';
 
+const useStyles = makeStyles<Theme, { pinned?: boolean }>(() => ({
+  button: ({ pinned }) => ({
+    '& > span': {
+      transition: 'transform .2s ease',
```
CarsonF, line 17: Could you use a standard transition from the theme?
https://github.com/SeedCompany/cord-field/blob/b6f501a39f3420e5f6af9569dcb63c784e11dace/src/scenes/Projects/Files/DirectoryBreadcrumb.tsx#L24

There's some preset durations as well if you don't want the default.
```tsx
transition: transitions.create('transform', {
  duration: transitions.duration.short,
}),
```

CarsonF (approved):

PR #1091, opened 2022-02-22
Create financial reports guide

### LINKED ISSUE #1088: Guide to create financial reports
Financial reports are only created after:
1. Project date range is set
2. There is a managing partner
3. There is a financial reporting frequency set on that partner

In the report card, we should look and see if the reports are empty (and `canRead`). If so, we should suggest that the user do the next requirement that is not fulfilled.

Maybe there's a button on the bottom of the card (CardAction) that has one of these:
1. "Set date range" -> open `UpdateProjectDialog`
2. "Add managing partner" -> go to partnerships list
3. "Set reporting frequency" -> go to managing partner or partnerships list

Maybe the card body can list these requirements?

zkhin:
Fixes https://github.com/SeedCompany/cord-field/issues/1088

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466506086) by [Unito](https://www.unito.io)
┆Link To Item: https://seed-company-squad.monday.com/boards/3451697530/pulses/3466506086


CarsonF (changes requested):
Seeing it in the UI now, I don't think this gives enough context.
Screen Shot 2022-02-22 at 8 44 31 AM
 
The new button just seems out of place, since there's no info to describe why it's there.
Also, both the main card action area & the view reports button make up the primary action of this card - to go to the reports list. But this doesn't add any value as the list will be empty.

I think instead we should remove all of the card actions and the card body under the title, with a check list. Maybe with some copy too that says something like, "complete these requirements to access financial reports".

I just tried to mock up something, and I think that puts too much info on this card.
Let's just add some copy to the card & remove the card actions, and then put dependency details on the report list page.
Something like this:
Screen Shot 2022-02-22 at 9 24 33 AM
Screen Shot 2022-02-22 at 9 25 01 AM

zkhin:
Here are updated UI screenshots.
image
image
image

src/components/PeriodicReports/FinancialReportCheckList.tsx
```diff
@@ -0,0 +1,57 @@
...
+  NarrativeReportsQuery,
+} from '../../scenes/Projects/Reports/ProjectReports.generated';
+
+const ReportGuideAction = {
+  SET_RANGE: 'SET_RANGE',
+  ADD_MANAGING_PARTNER: 'ADD_MANAGING_PARTNER',
+  SET_REPORT_FREQUENCY: 'SET_REPORT_FREQUENCY',
+};
+
+export const FinancialReportCheckList = ({
```
CarsonF, line 14: It still would be good to have these link to or open a dialog to take the action needed.
FWIW I used a [`<List>`](https://v4.mui.com/components/lists/#checkbox) component in my screenshot

src/components/PeriodicReports/PeriodicReportCard.tsx
```diff
@@ -119,63 +127,72 @@ const PeriodicReportCardInContext = (props: PeriodicReportCardProps) => {
...
+                  title="Next"
+                  report={dueNext}
+                  className={classes.relevantReport}
+                />
+              </div>
+            ) : (
+              <div>
+                <span>Not all requirements have been met</span> <br />
+                <span>Click here to view remaining list</span>
+              </div>
```
CarsonF, line 148: Does this have the appropriate typography or does it need to use the `Typography` component?

src/scenes/Projects/Reports/ProjectReports.graphql
```diff
@@ -6,6 +6,17 @@ query NarrativeReports($projectId: ID!, $changeset: ID) {
...
+    mouRange {
+      ...securedDateRange
+    }
+    financialReportPeriod {
+      canRead
+      canEdit
+      value
+    }
+    partnerships {
+      ...PartnershipSummary
```
CarsonF, line 18: We don't need all this data here.
```suggestion
      canRead total
```

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -159,6 +160,10 @@ export const ProjectOverview: FC = () => {
     },
   });
 
+  const { data: financialReportsData } = useQuery(FinancialReportsDocument, {
+    variables: { projectId, changeset: changesetId },
+  });
```
CarsonF, line 165: Let's just add the data we need to the existing ProjectOverview query instead of adding another one.

src/components/PeriodicReports/FinancialReportCheckList.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {
+  FinancialReportsQuery,
+  NarrativeReportsQuery,
+} from '../../scenes/Projects/Reports/ProjectReports.generated';
+
+const ReportGuideAction = {
+  SET_RANGE: 'SET_RANGE',
+  ADD_MANAGING_PARTNER: 'ADD_MANAGING_PARTNER',
+  SET_REPORT_FREQUENCY: 'SET_REPORT_FREQUENCY',
+};
```
CarsonF, line 12: This doesn't add much value, after enumerating we still check each value and do individual logic based on it. `checkList` could be simplified without it.

src/components/PeriodicReports/PeriodicReportsList.tsx
```diff
@@ -26,12 +26,14 @@ export const PeriodicReportsList = ({
...
   pageTitleSuffix,
   reports,
+  requirementCheckList,
   onRowClick,
 }: {
   type: ReportType;
   breadcrumbs?: ReactNode[];
   pageTitleSuffix?: string;
   reports?: readonly PeriodicReportFragment[];
+  requirementCheckList?: ReactNode;
```
CarsonF, line 36: Maybe this could just be `children` instead? I think that would be a bit more idiomatic for React.
As in the caller can override the children or we'll show the table by default.

CarsonF (changes requested):

sethmcknight:

PR #1093, opened 2022-03-07
Add product placeholder field

### LINKED ISSUE #1055: Product Placeholder Field
Add form field for new `Product.placeholderDescription` API field.

When using this field all other required fields, should not be required.

For produces field we should default to `"Unknown"`...but maybe that should be done in the API.

zkhin:
Fixes https://github.com/SeedCompany/cord-field/issues/1055

### UI
image
image

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466505085) by [Unito](https://www.unito.io)
┆Link To Item: https://seed-company-squad.monday.com/boards/3451697530/pulses/3466505085

### DIFF
--- a/src/scenes/Products/Detail/ProductDetail.graphql
+++ b/src/scenes/Products/Detail/ProductDetail.graphql
@@ -24,6 +24,9 @@ fragment ProductDetail on Product {
   describeCompletion {
     value
   }
+  placeholderDescription {
+    value
+  }
   engagement {
     ...EngagementBreadcrumb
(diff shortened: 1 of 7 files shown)


zkhin:
@CarsonF

> When using this field all other required fields, should not be required.

I checked the `createOtherProduct` mutation and the `title` field is required on the API side. How can we avoid this required field when using `placeholderDescription`

sethmcknight:

PR #1096, opened 2022-03-15
Altered .gitignore and removed .vscode/settings.json from repo

bryanjnelson:


.gitignore
```diff
@@ -26,7 +26,7 @@ package-lock.json
 
 # IDE - VSCode
 .vscode/*
-!.vscode/settings.json
+.vscode/settings.json
```
CarsonF, line 29:
```suggestion
```

CarsonF (changes requested):

CarsonF (commented):
Do you want to ignore tasks or launch while we are here too or you want to share those?

bryanjnelson (commented):
Nah, those can be left alone for now.

CarsonF (approved):
LGTM

PR #1107, opened 2022-04-21
Update organization name

zkhin:
### Actual
image

### DIFF
--- a/src/scenes/Partners/Edit/EditPartner.tsx
+++ b/src/scenes/Partners/Edit/EditPartner.tsx
@@ -175,27 +175,27 @@ export const EditPartner = ({
       ) => {
         const { 'partner.organizationName': nameDirty, ...dirty } =
           form.getState().dirtyFields;
-        if (nameDirty) {
-          await updateOrganizationName({
-            variables: {
-              id: partner.organization.value!.id,
-              name: organizationName,
-            },
-          });
-        }
-        if (Object.keys(dirty).length > 0) {
-          await updatePartner({
-            variables: {
-              input: {
-                partner: {
-                  ...rest,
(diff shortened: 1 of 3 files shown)


.yarn/build-state.yml
```diff
@@ -0,0 +1,42 @@
+# Warning: This file is automatically generated. Removing it is fine, but will
```
CarsonF, line 1: You can delete this file, it's no longer used.

src/scenes/Partners/Edit/EditPartner.tsx
```diff
@@ -155,8 +163,16 @@ export const EditPartner = ({
       decorators={decorators}
       initialValues={initialValues}
       onSubmit={async ({
-        partner: { pointOfContactId, pmcEntityCode, ...rest },
+        partner: { pointOfContactId, pmcEntityCode, organizationName, ...rest },
       }) => {
+        if (organizationName && partner.organization.value) {
```
CarsonF, line 168: The org name will always be truthy, even if it's not changed. You should check if this field is dirty instead. Here are some examples:
https://github.com/SeedCompany/cord-field/blob/f6001e70a1c304d64ef20ad30be88a7ae68645ec/src/scenes/Products/Edit/EditProduct.tsx#L259-L267
https://github.com/SeedCompany/cord-field/blob/f6001e70a1c304d64ef20ad30be88a7ae68645ec/src/scenes/Engagement/EditEngagement/EditEngagementDialog.tsx#L275-L281

CarsonF (changes requested):

zkhin:
@CarsonF Give me some minutes, I am resolving the type error on my end.

CarsonF (approved):

PR #1108, opened 2022-04-22
Add yarn sdk settings

zkhin:


.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+  "recommendations": [
+    "arcanis.vscode-zipfs",
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode"
```
CarsonF, line 5: This shouldn't be needed right? Running eslint invokes prettier plus more.
```suggestion
```

zkhin: Correct

CarsonF (approved):
Thanks!

PR #1120, opened 2022-05-10
Fix DateField initialValue handling

### LINKED ISSUE #1116: Periodic report received date previous values bit reinitialized
The periodic report's received date field should be re-initialized to the existing value when the form is created & rendered.

atGit2021:
Fixes #1116; Modified the UpdatePeriodicReportDialog component; enabled the received date value to be passed down to the DateField component

### DIFF
--- a/src/components/form/DateField.tsx
+++ b/src/components/form/DateField.tsx
@@ -204,13 +204,17 @@ export const humanFormat = inputFormat.toLowerCase();
 
 type DateInput = string | CalendarDate | null | undefined;
 
+const uninitialized = Symbol('uninitialized');
+
 /**
  * Memoizes date objects to prevent re-renders & parses ISO strings
  */
-const useDate = (valIn: DateInput) => {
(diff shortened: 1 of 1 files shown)


CarsonF (commented):
We have yet to need to use a _field's_ `initialValue` property. So far, the _form's_ `initialValues` property has been sufficient. Using the fields brings up the confusion of which property takes precedence.

Ah I actually just had a thought, and checked. It looks like all date fields in dialogs aren't getting their values initialized. So it's a problem with the `DateField`.

CarsonF:
I think it's because the `null` is being treated as an actual value
https://github.com/SeedCompany/cord-field/blob/fd8138d9177059d137f818ed3a77a47e8ff9f3c0/src/components/form/DateField.tsx#L218-L218

If we change it to `undefined` I think it'll work. 
We'll just need to double check that this doesn't affect submissions. 
Sending `undefined` to API just omits the field, aka "unchanged". Sending `null` actually attempts to set the value as null, aka "clear it".

CarsonF (approved):

PR #1125, opened 2022-05-19
Do not allow users to upload a planning file when there are no products

### LINKED ISSUE #1092: Prevent Progress Report from being uploaded before a Planning Spreadsheet exists
The ability to upload a Progress Report _before_ uploading a Planning Spreadsheet currently exists and should be eliminated.  It is not a viable option from a business perspective and can cause issues and confusion for users if the uploads are done out of order.

A note from Carson concerning implementation:
> We should just check if any products have been created though, don't look at the pnp property (planning file).

elisabeth-sorrell:
Closes https://github.com/SeedCompany/cord-field/issues/1092

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466497735) by [Unito](https://www.unito.io)
┆Link To Item: https://seed-company-squad.monday.com/boards/3451697530/pulses/3466497735

### DIFF
--- a/src/components/EngagementBreadcrumb/EngagementBreadcrumb.graphql
+++ b/src/components/EngagementBreadcrumb/EngagementBreadcrumb.graphql
@@ -4,6 +4,9 @@ fragment EngagementBreadcrumb on Engagement {
     ...ProjectBreadcrumb
   }
   ... on LanguageEngagement {
+    products {
+      total
+    }
     language {
       value {
         id
--- a/src/components/PeriodicReports/PeriodicReportCard.tsx
+++ b/src/components/PeriodicReports/PeriodicReportCard.tsx
@@ -69,7 +69,7 @@ export interface PeriodicReportCardProps {
   dueNext?: SecuredPeriodicReportFragment;
   disableIcon?: boolean;
   hasDetailPage?: boolean;
-  hasGoals: boolean;
+  hasGoals?: boolean;
 }
 
 const PeriodicReportCardInContext = (props: PeriodicReportCardProps) => {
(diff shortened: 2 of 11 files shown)


elisabeth-sorrell:
No products created:
Screen Shot 2022-05-19 at 3 24 58 PM

After a product is created
Screen Shot 2022-05-19 at 3 27 12 PM

bryanjnelson (changes requested):
This is close, @elisabeth-sorrell, but I think there was a misunderstood requirement here.  And I'll take that upon myself for not being clear enough in the issue. ☹️ 

You wrote the case that won't allow a Planning Spreadsheet upload if the Goals are 0.  However, that case is okay...one of the ways that the user will get Goals into Cord is through a PnP Planning Spreadsheet upload.

What we want to do here is not allow a Progress Report upload (Progress Tab of the PnP) if the Goals are 0.  This will force them to either a) add a goal manually through the UI _or_ upload a Planning Spreadsheet.

elisabeth-sorrell:
Okay, I have this:
1. The 'Upload Report' button is disabled. I was thinking of just making it disappear but thought that it might be a little confusing. This way clues in the users that something's up.
Screen Shot 2022-05-27 at 4 19 43 PM

2. There is no "New Version" button for each report row, so it's not an option to upload a new version of report as a shortcut when there are no products
Screen Shot 2022-05-27 at 4 21 18 PM

3. There is a disabled card explaining that a planning info file cannot be uploaded unless there's at least one goal/product
Screen Shot 2022-05-27 at 4 22 30 PM

bryanjnelson (commented):
Since I'm not familiar enough with the UI Code to approve anything I'll let that piece up to Carson.  However, I agree that disabling the button instead of removing it is better.  And the disabled card verbiage make sense.  I'm not sure about the removal of the "New Version" option; I'll let @CarsonF chime in on that.

bryanjnelson:
But it looks good...thanks @elisabeth-sorrell!

sethmcknight:

PR #1133, opened 2022-06-14
Fix partner address to be clearable

daniel-leod:
API PR: https://github.com/SeedCompany/cord-api-v3/pull/2468

### DIFF
--- a/src/scenes/Partners/Edit/EditPartner.tsx
+++ b/src/scenes/Partners/Edit/EditPartner.tsx
@@ -169,6 +169,7 @@ export const EditPartner = ({
             pointOfContactId,
             pmcEntityCode,
             organizationName,
+            address,
             ...rest
           },
         },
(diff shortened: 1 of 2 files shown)


src/scenes/Partners/Edit/EditPartner.tsx
```diff
@@ -190,6 +191,7 @@ export const EditPartner = ({
                 input: {
                   partner: {
                     ...rest,
+                    address: address?.trim().length ? address : null,
```
CarsonF, line 194: I think with https://github.com/SeedCompany/cord-api-v3/pull/2469 the field will be cleared now without this. Can you try?

daniel-leod: I tried yesterday but I think it's still needed. Let me double-check.

daniel-leod: Ok, just tried and to be able to clear the partner address it's needed to submit a blank space `' '` or more. That doesn't feel very intuitive to me.

CarsonF (approved):
Thanks @daniel-leod!

PR #1164, opened 2022-09-14
removed tss-react from cord-field - 3466480946 

andrewdavid48:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466487404) by [Unito](https://www.unito.io)
┆Link To Item: https://seed-company-squad.monday.com/boards/3451697530/pulses/3466487404

### DIFF
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -69,13 +69,13 @@ export const Error = ({
 
   return (
     <Box
-      // TODO: use page based sx props
+      component={Component}
       sx={(theme) => ({
         overflow: 'auto',
         padding: theme.spacing(4, 0, 0, 4),
       })}
     >
-      <Component>
+      <>
         {/* Default status code to be helpful for the most common ones. The
       children can still override this by rendering <StatusCode /> themselves */}
         <StatusCode code={statusCode} />
@@ -105,7 +105,7 @@ export const Error = ({
             </Grid>
           </Grid>
         )}
-      </Component>
+      </>
     </Box>
   );
 };
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -45,6 +45,7 @@ export const Error = ({
   children,
   show,
   disableButtons,
+  page,
   component: Component = 'div',
 }: ErrorProps) => {
   const navigate = useNavigate();
@@ -70,10 +71,7 @@ export const Error = ({
   return (
     <Box
       component={Component}
-      sx={(theme) => ({
-        overflow: 'auto',
-        padding: theme.spacing(4, 0, 0, 4),
-      })}
+      sx={[!!page && { overflow: 'auto', pt: 4, pl: 4 }]}
     >
       <>
         {/* Default status code to be helpful for the most common ones. The
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -1,7 +1,7 @@
 import { ApolloError } from '@apollo/client';
 import { Box, Button, Grid, Typography } from '@mui/material';
 import { isPlainObject } from 'lodash';
-import { isValidElement, ReactNode } from 'react';
+import { ElementType, isValidElement, ReactNode } from 'react';
 import { getErrorInfo } from '~/api';
 import { ButtonLink, StatusCode, useNavigate } from '../Routing';
 import { ErrorRenderers, renderError } from './error-handling';
@@ -26,6 +26,7 @@ export interface ErrorProps {
   page?: boolean;
   /** Turn off back & home buttons */
   disableButtons?: boolean;
+  component?: ElementType;
 }
 
 /**
@@ -44,6 +45,7 @@ export const Error = ({
   children,
   show,
   disableButtons,
+  component: Component = 'div',
 }: ErrorProps) => {
   const navigate = useNavigate();
 
@@ -73,7 +75,7 @@ export const Error = ({
         padding: theme.spacing(4, 0, 0, 4),
       })}
     >
-      <>
+      <Component>
         {/* Default status code to be helpful for the most common ones. The
       children can still override this by rendering <StatusCode /> themselves */}
         <StatusCode code={statusCode} />
@@ -103,7 +105,7 @@ export const Error = ({
             </Grid>
           </Grid>
         )}
-      </>
+      </Component>
     </Box>
   );
 };
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -1,21 +1,10 @@
 import { ApolloError } from '@apollo/client';
-import { Button, Grid, Typography } from '@mui/material';
+import { Box, Button, Grid, Typography } from '@mui/material';
 import { isPlainObject } from 'lodash';
-import { ElementType, isValidElement, ReactNode } from 'react';
-import { makeStyles } from 'tss-react/mui';
+import { isValidElement, ReactNode } from 'react';
 import { getErrorInfo } from '~/api';
 import { ButtonLink, StatusCode, useNavigate } from '../Routing';
 import { ErrorRenderers, renderError } from './error-handling';
-
-const useStyles = makeStyles()(({ spacing }) => ({
-  page: {
-    overflow: 'auto',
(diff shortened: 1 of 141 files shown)


src/components/Sensitivity/Sensitivity.tsx
```diff
@@ -59,17 +59,19 @@ export interface SensitivityProps {
   value?: SensitivityType;
   loading?: boolean;
   className?: string;
+  sx?: SxProps<Theme>;
```
CarsonF, line 62: I've got a shortcut for this here
https://github.com/SeedCompany/cord-field/blob/2f3b79721181e77a674a7ffa1d741d9edf7f51fb/src/common/sx.ts#L5-L8

src/components/LanguageListItemCard/LanguageListItemCard.tsx
```diff
@@ -10,41 +16,39 @@ import { Sensitivity } from '../Sensitivity';
...
-      marginTop: spacing(1),
-    },
-    pin: {
-      position: 'absolute',
-      top: 10,
-      right: 10,
-    },
-  };
-});
+const classes = {
```
CarsonF, line 19: Hmm I'm not sure I like this. Downsides I can first think of is not having type completion on these css props and if there's an error it would be shown at the usage instead of this declaration. What do you think about just inlining these objects?

andrewmurraydavid: Inlining them makes sense. This first commit was mainly to see what direction we want to go with the and I agree it's nicer to get and error on declaration rather than usage

CarsonF (approved):
I think I like this. It does "clutter" the jsx a bit more, but I think it's better to have those styles close to their usage. I've often found myself asking "what does this class do" and having to jump back and forth between styling at top and usage in jsx tree.

So I'm good with this if you like it 👍

CarsonF:
Will you squash/fixup that second commit as well?

andrewmurraydavid:
@CarsonF 
What second commit? If you're talking about the one with the yarn updates, that has been removed.These 2 commits here are only with the tss-react changes

CarsonF:
> These 2 commits here are only with the tss-react changes

Yeah I'd just prefer that "clean up" commits get squashed so we are left with single purpose commits.

src/components/Error/Error.tsx
```diff
@@ -73,7 +75,7 @@ export const Error = ({
         padding: theme.spacing(4, 0, 0, 4),
       })}
     >
-      <>
+      <Component>
```
CarsonF, line 78: You can use the component prop on Box instead of adding another element to DOM

src/components/List/List.tsx
```diff
@@ -61,23 +52,34 @@ export const List = <Item extends Entity>(props: ListProps<Item>) => {
...
-  });
 
   const scrollRef = useRef<HTMLDivElement | null>(null);
   usePersistedScroll(scrollRefProp ?? scrollRef);
   const determineChangesetDiff = useDetermineChangesetDiffItem();
 
   return (
-    <div className={cx(classes.root, className)} ref={scrollRef}>
+    <Box
+      component="div"
```
CarsonF, line 65: Pretty sure this is the default
```suggestion
```

src/components/List/List.tsx
```diff
@@ -61,23 +52,34 @@ export const List = <Item extends Entity>(props: ListProps<Item>) => {
...
+      className={className}
+      ref={scrollRef}
+      css={(theme) => [
+        {
+          overflow: 'auto',
+          marginLeft: theme.spacing(-2),
+          padding: theme.spacing(2),
+        },
+      ]}
+      sx={containerSx}
```
CarsonF, line 75: Let's stick with the single declaration so we can be sure which css values are applied in which order. Since we have `sx` here, let's use it. I have a helper for this case.
```ts
sx={[
   { overflow: 'auto', ml: -2, p: 2 },
   ...extendSx(containerSx),
]}
```
Since `containerSx` is listed after definitions, it'll override our defaults.

src/components/List/List.tsx
```diff
@@ -61,23 +52,34 @@ export const List = <Item extends Entity>(props: ListProps<Item>) => {
...
+    >
       <Grid
         direction="column"
         spacing={spacing}
         {...ContainerProps}
         container
-        className={classes.container}
+        sx={sx}
```
CarsonF, line 82: Isn't it a bit confusing to have `ContainerProps.sx` and `sx` and then `containerSx` which doesn't go to the "container"?

What if `sx` goes to the Box above and then `containerSx` just gets dropped in favor of `ContainerProps.sx`?

src/scenes/Users/List/UserList.tsx
```diff
@@ -23,23 +23,9 @@ import { UserSortOptions } from './UserSortOptions';
...
-    overflowY: 'auto',
-    // allow card shadow to bleed over instead of cutting it off
-    padding: spacing(0, 0, 0, 2),
-    margin: spacing(0, 0, 0, -2),
-  },
-  total: {
-    marginTop: spacing(2),
-  },
-}));
+const itemsSx = (theme: Theme) => ({
```
CarsonF, line 26: I wonder if it would be better to name this like it's purpose, "max width", instead of where it's used. Or maybe "itemWidth"?

src/scenes/Users/List/UserList.tsx
```diff
@@ -80,18 +71,28 @@ export const UserList = () => {
         <TabList
           onChange={(_e, tab) => setFilters({ ...filters, tab })}
           aria-label="user navigation tabs"
-          className={classes.items}
+          sx={(theme) => itemsSx(theme)}
```
CarsonF, line 74:
```suggestion
          sx={itemsSx}
```

CarsonF (changes requested):

src/components/List/List.tsx
andrewmurraydavid: The reason `sx` was added was because `List` component does not have and `sx` and does not inherit it either. `containerSx` will be removed :)

CarsonF (changes requested):
Here's some comments to get started on. The idea behind each comment could be applied across the board so take a look at the meaning and see where else it could be applied.

src/components/Error/Error.tsx
willdch: `    <Box
      component={Component}
      sx={(theme) => ({
        overflow: 'auto',
        padding: theme.spacing(4, 0, 0, 4),
      })}
    >
      <>
        {/* Default status code to be helpful for the most common ones. The
      children can still override this by rendering <StatusCode /> themselves */}
        <StatusCode code={statusCode} />
        <Typography gutterBottom>Oops, Sorry.</Typography>
        {rendered}
        {!disableButtons && (
          <Grid
            container
            spacing={3}
            sx={{
              marginTop: 3,
            }}
          >
            <Grid item>
              <Button
                onClick={() => navigate(-1)}
                variant="contained"
                color="secondary"
              >
                Back
              </Button>
            </Grid>
            <Grid item>
              <ButtonLink to="/" variant="contained" color="secondary">
                Home
              </ButtonLink>
            </Grid>
          </Grid>
        )}
      </>
    </Box>`

removing the <Component> and just using the Box `component={Component}` prop throws the following error.
`This JSX tag's 'children' prop expects a single child of type 'ReactNode', but multiple children were provided.`

To get around this I can use a react Fragment as a parent container and it doesn't appear to add a new element to the DOM.  Will this work?

src/components/Error/Error.tsx
```diff
@@ -81,30 +68,44 @@ export const Error = ({
...
 
   return (
-    <Component className={cx(page && classes.page)}>
-      {/* Default status code to be helpful for the most common ones. The
+    <Box
+      component={Component}
+      sx={(theme) => ({
+        overflow: 'auto',
+        padding: theme.spacing(4, 0, 0, 4),
+      })}
```
CarsonF, line 76: Re-introduce this style behind the page flag.
```suggestion
      sx={[!!page && { overflow: 'auto', pt: 4, pl: 4, }]}
```

sethmcknight:

PR #1166, opened 2022-10-03
Comments functionality

### LINKED ISSUE #1150: RichText via Editor.js
I want to integrate with https://editorjs.io/. 

How does rendering work? Can each block type use a customized component that fits our style?
It's possible to render differently for view vs edit. Edit gets the full editor, but view just renders the markup. Sounds more performant.

https://github.com/editor-js/awesome-editorjs
https://github.com/Jungwoo-An/react-editor-js
https://github.com/moveyourdigital/editorjs-blocks-react-renderer

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466492736) by [Unito](https://www.unito.io)
┆Link To Item: https://seed-company-squad.monday.com/boards/3451697530/pulses/3466492736

andrewdavid48:
Created the comments feature to enable the use of Rich text using the EditorJs library.

Currently the only Commentable we have is the Project object and we're possibly looking into making other object commentable as well. Because this PR is not as much focused on the Comments feature but more on the Rich text, we put aside some API work which was not critical to the Rich text feature.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466486753) by [Unito](https://www.unito.io)
┆Link To Item: https://seed-company-squad.monday.com/boards/3451697530/pulses/3466486753


src/scenes/Root/CommentsBar/CommentItem/CommentItem.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+export interface CommentProps {
+  comment: CommentPropsFragment;
+  repliesCount?: number;
+  isChild?: boolean;
+  resourceId: string;
+  isExpanded?: boolean;
+  parent: CommentThreadPropsFragment;
+  handleDeleteComment?: (comment: CommentPropsFragment) => Promise<void> | void;
+  handleEditComment?: (comment: CommentPropsFragment) => Promise<void> | void;
```
CarsonF, line 25: callback properties always start with `on` not `handle`. Handle is the prefixed used to declare the handler function for the callback property, though I don't really care if this is followed. I think I define mine inline a lot of times anyways.

Since this is a comment as well I think the comment suffix is superfluous. So `onDelete`/`onEdit` would be better. Though it looks like edit isn't even used?

Also the `Promise` return type isn't the best. This code that calls the function doesn't await the promise before doing anything; it just calls it and leaves the promise hanging. So it would be better to "hang" that promise in the calling code. It's more explicit that it is async logic that is not awaited.

src/scenes/Root/CommentsBar/CommentItem/CommentItem.tsx
```diff
@@ -0,0 +1,179 @@
...
+      variables: {
+        id: commentId,
+      },
+      update: removeItemFromList({
+        listId: [parent, 'comments'],
+        item: comment,
+      }),
+    });
+
+    if (data?.deleteComment) {
```
CarsonF, line 57: I believe the mutation function call throws on error (with the default error policy at least), so there's no need to check for success here.

src/scenes/Root/CommentsBar/CommentReply/CommentReply.tsx
```diff
@@ -0,0 +1,176 @@
...
+  threadId?: string;
+  resourceId: string;
+  commentId?: string;
+  placeholder?: string;
+  blocks?: JSON;
+  isEditing?: boolean;
+  handleClose?: () => void;
+}
+
+export const CommentReply = ({
```
CarsonF, line 26: I think this should be split. Editing a comment is a different use case than creating a new comment on a thread.

`EditComment` can receive a `commentId`.
`CommentReply` could receive a `parent` being a `ThreadFragment` or `CommentableFragment`.

This will help limit the prop surface area.

src/scenes/Root/CommentsBar/CommentItemMenu/CommentMenuItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+  commentId,
+  threadId,
+  anchorEl,
+  open,
+  onDelete,
+  onEdit,
+  ...rest
+}: CommentItemMenuProps) => {
+  const listItemSx = (theme: Theme) => ({
+    marginRight: theme.spacing(2),
```
CarsonF, line 28: FWIW
```suggestion
  const listItemSx = {
    marginRight: 2,
```

src/scenes/Root/CommentsBar/CommentThread/CommentThreadAccordion.tsx
```diff
@@ -0,0 +1,66 @@
...
+          margin: 0,
+        },
+        padding: 0,
+      }}
+      elevation={0}
+    >
+      {children}
+      <AccordionDetails
+        sx={{
+          '&': {
```
CarsonF, line 36: Is this doing anything?

src/scenes/Root/CommentsBar/CommentReply/CommentReply.tsx
```diff
@@ -0,0 +1,176 @@
...
+          >
+            <EditorJsWrapper
+              holder={`reply-to-${commentId}`}
+              onInitialize={handleInitialize}
+              placeholder={placeholder}
+              autofocus
+              blocks={isEditing ? blocks : undefined}
+              customTools={{
+                paragraph: true,
+                header: true,
```
CarsonF, line 151: I don't think we should have headers here.
```suggestion
```

CarsonF (changes requested):
Progress!

## UI Comments

Screen Shot 2022-10-18 at 9 26 02 AM
The title/close button need some work. The icon is stretched instead of giving it its proper space.
The title style seems a little weird too... maybe an `h3` would be better.

---
Screen Shot 2022-10-18 at 9 29 02 AM
I think the title could be dropped here and the placeholder changed to "Write a comment..."

---
Screen Shot 2022-10-18 at 9 39 47 AM
These shadows look weird. Each thread should be at the same "0" elevation in the list, so just like a line border appearance. It was just the thread replies that were supposed to look like a "sub-drawer" that was "further below" the list of threads.

---
Screen Shot 2022-10-18 at 9 44 10 AM
This doesn't seem the best. Perhaps we can just have a reply button in the bottom right of the comment thread instead?

---
Screen Shot 2022-10-18 at 9 48 15 AM

- "Hide Replies" doesn't look that good. What if we did an expand icon in the bottom right, below the context button?
   Screen Shot 2022-10-18 at 9 51 27 AM
  Maybe a "Reply" button could sit to the left of that?
- I'm not happy with the indentation of the replies. I think that's valuable real estate to be consuming. I think if we can make the "sub drawer" appear "under" more that would be better. It's also two competing designs trying to communicate the same thing: shadows and indentation. If the shadows are not enough, maybe a different subtle background shade could help?

src/TestContext.tsx
```diff
@@ -0,0 +1,35 @@
+import { ApolloProvider } from '@apollo/client';
```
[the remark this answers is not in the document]
andrewmurraydavid: Yes, It was meant to be reverted when I reverted the initial tests idea, but was hidden in git cache locally. I ended up reverting it in one of my last commits

sethmcknight:

PR #1171, opened 2022-10-19
Fix BetaFeatures so that users with betafeatures on can see change requests

elisabeth-sorrell:


src/components/Session/Session.tsx
```diff
@@ -42,7 +43,9 @@ export const updateSessionCache = <T extends LoginMutation | RegisterMutation>(
 };
 
 export const useBetaFeatures = () =>
-  useSession().powers?.includes('BetaFeatures');
+  Object.entries(useSession().betaFeatures || {}).flatMap(([key, value]) =>
+    value ? key : []
+  );
```
CarsonF, line 48: I'd prefer all this wrapped in a `new Set(...)` as that is a better shape to output.

The return type should be specified as
```ts
(): ReadonlySet<keyof BetaFeaturesFragment>
```

src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -107,6 +107,7 @@ export const ProjectOverview = () => {
   const { classes, cx } = useStyles();
   const { projectId, changesetId } = useProjectId();
   const beta = useBetaFeatures();
+  console.log(beta);
```
CarsonF, line 110:
```suggestion
```

CarsonF (changes requested):
Thanks @elisabeth-sorrell! just a couple things...

src/components/Changeset/ChangesetDiff.graphql
```diff
@@ -1,7 +1,7 @@
-query ChangesetDiff($changeset: ID!) {
+query ChangesetDiff($changeset: ID!, $resource: ID) {
```
CarsonF, line 1: Was this added by mistake?

src/components/Session/Session.tsx
```diff
@@ -2,7 +2,7 @@ import { ApolloCache, useQuery } from '@apollo/client';
 import { pickBy } from 'lodash';
 import LogRocket from 'logrocket';
 import { useEffect } from 'react';
-import { SessionOutput } from '~/api/schema.graphql';
+import { BetaFeatures, SessionOutput } from '~/api/schema.graphql';
```
CarsonF, line 5: Can you use the type from the fragment you added instead of this schema type? It'll prevent checking for a feature that hasn't been fetched in future.

src/components/Changeset/ChangesetDiff.graphql
elisabeth-sorrell: yeah something slipped in from other work I was doing

CarsonF (approved):
Thanks @elisabeth-sorrell. Looks great!

PR #1188, opened 2022-11-03
Added TeamHighlight step

andrewmurraydavid:
Created TeamHighlights step for the new Progress report feature.

API PR: https://github.com/SeedCompany/cord-api-v3/pull/2554


src/scenes/ProgressReports/Detail/ProgressReportDrawer/ProgressReportSidebar.tsx
```diff
@@ -1,20 +1,27 @@
...
+import { useProgressReportContext } from '../../ProgressReportContext';
 import { InstructionsDialog } from './InstructionsDialog';
 import { ProgressReportStepper } from './ProgressReportStepper';
 
-export const ProgressReportSidebar = ({ step }: { step: number }) => {
+export const ProgressReportSidebar = () => {
+  const { step, currentReport } = useProgressReportContext();
   const [instructionsState, showInstructions] = useDialog();
 
+  const daysLeft = Math.floor(currentReport?.due.diffNow('days').days ?? 0);
```
CarsonF, line 12:
```suggestion
  const daysLeft = currentReport?.due.toRelative({ options if needed });
```

src/scenes/ProgressReports/Detail/ProgressReportDrawer/ProgressReportSidebar.tsx
```diff
@@ -1,20 +1,27 @@
...
       sx={{
         display: 'initial',
         flexDirection: 'column',
-        minWidth: 350,
+        width: 300,
         padding: 2,
         pt: 4,
+        right: 0,
+        top: 0,
+        position: 'fixed',
```
CarsonF, line 24: This has me 😬 a little bit. Esp the part where the main content separately has to declare `mr: 300px`. Could we use flexbox or grid here instead?

src/scenes/ProgressReports/Detail/ProgressReportDrawer/ProgressReportSidebar.tsx
```diff
@@ -35,8 +42,22 @@ export const ProgressReportSidebar = ({ step }: { step: number }) => {
...
       </Typography>
-      <Typography variant="body2" color="info.light" sx={{ marginBottom: 2 }}>
-        36 DAYS before due
+      <Typography
+        variant="body2"
+        color={daysLeft > 0 ? 'info.light' : 'error.dark'}
+        sx={{ marginBottom: 2 }}
+      >
+        <>
+          {Math.abs(daysLeft)} {daysLeft > 0 ? 'DAYS before' : 'DAYS over'} due
```
CarsonF, line 51:
```suggestion
          {Math.abs(daysLeft)} {due >= DateTime.now() ? 'DAYS before' : 'DAYS over'} due
```

src/scenes/ProgressReports/Detail/ProgressReportDrawer/ProgressReportStepper.tsx
```diff
@@ -72,44 +70,11 @@ export const ProgressReportStepper = ({ step }: ProgressReportStepperProps) => {
...
-        />
-        <Typography
-          sx={{
-            padding: 1,
-            flexGrow: 2,
-          }}
-        >
-          Partner Steps
-        </Typography>
+    <Paper elevation={4} sx={{ mr: 2, overflow: 'hidden' }}>
```
CarsonF, line 73: How come overflow hidden?

src/scenes/ProgressReports/Detail/ProgressReportDrawer/RoleIcon.tsx
```diff
@@ -0,0 +1,56 @@
+import { Grading, Translate } from '@mui/icons-material';
+import { Box } from '@mui/material';
+import { Role } from '~/api/schema.graphql';
+import { PeopleJoinedIcon } from '~/components/Icons';
+import { ProjectManagerIcon } from '~/components/Icons/ProjectManagerIcon';
+import { colorPalette } from './colorPalette';
+
+interface StepIconProps {
+  roleStep?: Role | null;
```
CarsonF, line 9:
```suggestion
  role?: Role | null;
```

src/scenes/ProgressReports/Detail/ProgressReportDrawer/RoleIcon.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+const variantToIconMapper: {
+  [key in Role]: React.FC | null;
```
CarsonF, line 13: I'd be ok with this then we could omit all the roles that we don't want to declare icons for. And it doesn't need adjusting if a role is added.

We could also enforce that the map is `SvgIcons`, then we could skip the `Box` below.
```suggestion
  [key in Role]?: SvgIconComponent;
```

```ts
const Icon = variantToIconMapper[role];
return Icon && <Icon sx={...} />;
```

src/scenes/ProgressReports/Detail/ProgressReportDrawer/RoleIcon.tsx
```diff
@@ -0,0 +1,56 @@
...
+  LeadFinancialAnalyst: null,
+  Leadership: null,
+  Liaison: null,
+  Mentor: null,
+  RegionalCommunicationsCoordinator: null,
+  RegionalDirector: null,
+  StaffMember: null,
+};
+
+export const RoleIcon = ({ roleStep = 'FieldPartner' }: StepIconProps) => {
```
CarsonF, line 39: I don't think this component should assume any default. It's a business-like decision that should be made in a smarter component.

src/scenes/ProgressReports/Detail/ProgressReportDrawer/RoleIcon.tsx
```diff
@@ -0,0 +1,56 @@
...
+  if (!roleStep) {
+    return null;
+  }
+
+  return (
+    <Box
+      component={variantToIconMapper[roleStep] ?? PeopleJoinedIcon}
```
CarsonF, line 46: I wouldn't assume this default either. If we really want a fallback icon, I'd suggest adding a `fallbackRole` prop.

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/NextStepButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  '&:hover': {
+    textDecoration: 'underline',
+  },
+});
+
+export const NextStepButton = ({ sx }: StyleProps) => {
+  const { nextProgressReportStep } = useProgressReportContext();
+
+  return (
+    <Box
```
CarsonF, line 20: Can we just use an MUI `<Button>` here instead? This is a lot of work to make this half way between a link and a button. IMO it should just be a normal button style. If truly desired we could disableRipple and change hover to underline. It would just be less consistent with the rest of the app.

---

Also another pet peeve of mine is doing navigation in an `onClick`. Navigation should always be done with Link-type components (`Link`, `ButtonLink`, etc.) Those use anchor tags which allow "open in new tab/window" to work normally (via modifier key or from right click context menu).

The downside for this is that the `to` path needs to be determined ahead of time, which requires a slight refactor. The end result though is just swapping the `nextStep` function with a string.

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/PromptVariant/PromptsForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+                    />
+                  );
+                })}
+              </EnumField>
+            )}
+
+            <SubmitButton
+              variant="outlined"
+              color="secondary"
+              disabled={validating}
```
CarsonF, line 72: This is handled internally.
```suggestion
```

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/PromptVariant/PromptsForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <Typography variant="body2" sx={{ mb: 2 }}>
+              As you reflect on the past three months, select ONE question to
+              answer.
+            </Typography>
+
+            {stepData?.prompts.length && (
+              <EnumField
+                name="prompt"
+                label="Select a question"
+                validate={[required]}
```
CarsonF, line 51: The `required` flag does this interally.
```suggestion
```

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/PromptVariant/PromptsForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+              display: 'flex',
+              flexDirection: 'column',
+              padding: 2,
+              pt: 0,
+            }}
+          >
+            <Typography sx={{ mb: 2 }} variant="h3">
+              Share a team highlight story.
+            </Typography>
+            <Typography variant="body2" sx={{ mb: 2 }}>
```
CarsonF, line 42: `gutterBottom` or maybe `paragraph` here instead of the manual margin bottom?

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/PromptVariant/PromptsForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+export const PromptsForm = ({
+  availableData: stepData,
+  reportId,
+  createItemMutation,
+}: PromptsFormProps) => {
+  return (
+    <Form
+      onSubmit={createItemMutation}
+      initialValues={{
+        reportId,
```
CarsonF, line 26: I typically just grab this from scope. I know this is how it's done in traditional html/server-side forms. I just don't see the need here. This probably makes sense for the current way you are connecting the submit handler. It'll probably be less needed once looking at my other comments.
[An example](https://github.com/SeedCompany/cord-field/blob/2667b2d8c09f382454516be3ac56579195262df9/src/scenes/Products/Edit/EditProduct.tsx#L171-L174).

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/PromptVariant/PromptsForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+            </Typography>
+            <Typography variant="body2" sx={{ mb: 2 }}>
+              As you reflect on the past three months, select ONE question to
+              answer.
+            </Typography>
+
+            {stepData?.prompts.length && (
+              <EnumField
+                name="prompt"
+                label="Select a question"
```
CarsonF, line 50: We should be consistent with our noun. Since we talked this morning about using "Selected Prompt" in other view, we should use that noun here too.
```suggestion
                label="Select a prompt"
```

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/PromptVariant/VariantResponsesAccordion.tsx
```diff
@@ -0,0 +1,89 @@
...
+  expanded?: boolean;
+  onSubmit: (input: any) => void | SubmissionErrors | Promise<SubmissionErrors>;
+}) => {
+  const [expanded, setExpanded] = useState(_expanded ?? false);
+  const [savedAt, setSavedAt] = useState<DateTime | null>(null);
+
+  if (response.response.canRead) {
+    return (
+      <Accordion
+        key={response.variant.responsibleRole}
```
CarsonF, line 32:
```suggestion
        key={response.variant.key}
```

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/PromptVariant/VariantResponsesAccordion.tsx
```diff
@@ -0,0 +1,89 @@
...
+  onSubmit,
+}: {
+  response: ProgressReportItemResponseEditFragment;
+  expanded?: boolean;
+  onSubmit: (input: any) => void | SubmissionErrors | Promise<SubmissionErrors>;
+}) => {
+  const [expanded, setExpanded] = useState(_expanded ?? false);
+  const [savedAt, setSavedAt] = useState<DateTime | null>(null);
+
+  if (response.response.canRead) {
```
CarsonF, line 29: Invert.

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/PromptVariant/VariantResponsesAccordion.tsx
```diff
@@ -0,0 +1,89 @@
...
+              {({ handleSubmit }) => (
+                <form onSubmit={handleSubmit}>
+                  <RichTextField name="response" label="Response" />
+                  {savedAt && (
+                    <Typography
+                      variant="caption"
+                      sx={{ mb: 1 }}
+                      component="div"
+                    >
+                      Saved at {savedAt.toISO()}
```
CarsonF, line 68:
```suggestion
                      Saved at <FormattedDateTime date={savedAt} />
```
I'll add this property to the API as well.

CarsonF (changes requested):
Looking great @andrewmurraydavid!

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/TeamHighlight/TeamHighlightStep.graphql
```diff
@@ -0,0 +1,35 @@
...
+      ... on ProgressReport {
+        id
+        highlights {
+          ...Pagination
+          items {
+            id
+          }
+        }
+      }
+    }
```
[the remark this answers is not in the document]
andrewmurraydavid: Actually it does need to update the list of highlights as that is how we are **currently** checking if it needs to display a prompt or the variant responses, just fyi @CarsonF 
https://github.com/SeedCompany/cord-field/pull/1188/files/da7f7d13c4e133a74c6570534aa6a0be89c8a45c#diff-62977201f87c94179cc270364dadd3d3ce0941a77208b06a3a81e558595199a5R12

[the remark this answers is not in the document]
CarsonF: Well what I mean is it doesn't directly update the list. Since the list is normalized, it doesn't matter here.
So
```
highlights: ['PVR:abc', 'PVR:def']
```
Now we modify `PVR:def` by submitting a response, and we receive back the that updated object.
So `PVR:def` is now includes a new/updated response, and that cache is updated. Since the `highlights` list is just referencing that normalized cache key it will get the new data when querying the cache.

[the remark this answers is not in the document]
andrewmurraydavid: I am sorry, I just realized it was for the UpdateResponse, not CreateHightlight! This will be removed, yes!

src/scenes/ProgressReports/Detail/ProgressReportDrawer/ProgressReportSidebar.tsx
andrewmurraydavid: Not really because of the position fixed :( unfortunately that breaks the standard grids. If you get something working, I'm happy to change it, but from what I've tried, the built-in solutions didn't work. Also the main container does not declare `mr: 300px`, it declares `width: 'calc(100% - 300px)'`

src/scenes/ProgressReports/Detail/ProgressReportDrawer/ProgressReportDrawer.tsx
```diff
@@ -1,29 +1,42 @@
+import { useQuery } from '@apollo/client';
```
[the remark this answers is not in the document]
andrewmurraydavid: This will create major conflicts for the branches that are **based** on this PRs branch, just FYI. I don't feel comfortable making this change in this PR. I can make this change in the Progress step PR

src/scenes/ProgressReports/Detail/ProgressReportDrawer/ProgressReportStepper.tsx
andrewmurraydavid: Not anymore. There was a weird issue back when we had a header on the stepper. Now it displays properly

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/PromptVariant/VariantResponsesAccordion.tsx
andrewmurraydavid: The reason I was using ISO (and I agree that we can move away from it) was because it was providing seconds and milliseconds as well. Using the `FormattedDateTime` component does not provide that level of detail: `Saved at 11/8/2022, 9:20 AM`

src/scenes/ProgressReports/Detail/ProgressReportDrawer/Steps/NextStepButton.tsx
CarsonF: Let's hold off on Links, since that would move us away from `makeQueryHandler`. How we handle the query params could need to be customized at usage of the Link. Like do we merge or replace the params? Is it all or just our own? what are our own? etc.

andrewmurraydavid:
As discussed, we will be making the fragment, context and moving changes in the Progress step PR to avoid conflicts in the PRs based on this branch

CarsonF (approved):
Thanks @andrewmurraydavid!

PR #1196, opened 2022-11-15
Tweak types in RichTextField & DRY check for empty data

CarsonF:
It is still unclear whether the `blocks` property can be nullable. If we figure out in the future that it is, then we can adjust the check for that in the now one spot.


andrewmurraydavid (approved):
Thanks for making this extra check!

PR #1199, opened 2022-11-21
Add option to show count characters for RichTextField

CarsonF:
```tsx
<RichTextField showCharacterCount />
```
Screen Shot 2022-11-21 at 4 16 35 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,8 @@
     "ts-essentials": "^9.2.0",
     "tss-react": "^3.7.1",
     "validator": "^13.7.0",
-    "xlsx": "^0.18.5"
+    "xlsx": "^0.18.5",
+    "xss": "^1.0.14"
   },
   "devDependencies": {
     "@babel/core": "^7.18.10",
--- a/src/components/Formatters/useNumberFormatter.tsx
+++ b/src/components/Formatters/useNumberFormatter.tsx
@@ -13,3 +13,8 @@ export const useNumberFormatter = (options?: Intl.NumberFormatOptions) => {
     [formatter]
   );
 };
+
+export const FormattedNumber = ({ value }: { value: Nullable<number> }) => {
+  const formatNumber = useNumberFormatter();
+  return <>{formatNumber(value)}</>;
+};
--- a/src/components/RichText/RichTextField.tsx
+++ b/src/components/RichText/RichTextField.tsx
@@ -9,6 +9,7 @@ import {
   ClickAwayListener,
   FormControl,
   FormHelperText,
+  Grid,
   InputLabel,
   OutlinedInput,
   TextField,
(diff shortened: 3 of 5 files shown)


src/components/Formatters/useNumberFormatter.tsx
```diff
@@ -13,3 +13,8 @@ export const useNumberFormatter = (options?: Intl.NumberFormatOptions) => {
     [formatter]
   );
 };
+
+export const FormattedNumber = ({ value }: { value: Nullable<number> }) => {
```
CarsonF, line 17: We can add more options here later.

package.json
```diff
@@ -108,7 +108,8 @@
     "ts-essentials": "^9.2.0",
     "tss-react": "^3.7.1",
     "validator": "^13.7.0",
-    "xlsx": "^0.18.5"
+    "xlsx": "^0.18.5",
+    "xss": "^1.0.14"
```
CarsonF, line 112: `xss` was the smallest and tied for most-downloaded, and it's recently updated. 

https://npmtrends.com/dompurify-vs-sanitize-html-vs-xss
https://bundlephobia.com/package/xss@1.0.14

src/components/RichText/RichTextField.tsx
```diff
@@ -50,13 +59,14 @@ export function RichTextField({
   sx,
   className,
   placeholder,
+  showCharacterCount,
   ...props
 }: RichTextFieldProps) {
-  const instanceRef = useRef<EditorCore>();
+  const instanceRef = useRef<EditorJS>();
```
CarsonF, line 65: EditorCore is a stripped down version meant to be compatible with SSR. The library's implementation of SSR was lacking and so I'm just avoiding it all together. This actual object gives better types in case we want to do something else with the editor later.

src/components/RichText/RichTextField.tsx
```diff
@@ -234,7 +248,16 @@ export function RichTextField({
...
-                        {getHelperText(meta, helperText)}
+                        {getHelperText(
+                          meta,
+                          helperText ??
+                            (showCharacterCount && val ? (
+                              <>
+                                <FormattedNumber value={val.characterCount} />{' '}
+                                characters
+                              </>
+                            ) : undefined)
```
CarsonF, line 259: This was a first pass. It might be good to right align. It might be good to separate out from actual helper text so both can exist at once.

src/components/RichText/RichTextField.tsx
```diff
@@ -234,7 +248,16 @@ export function RichTextField({
                       />
 
                       <FormHelperText id={`${id}-helper-text`}>
```
andrewmurraydavid, line 250: Because `getHelperText` supports a ReactNode, we can do this:

```suggestion
                      <FormHelperText id={`${id}-helper-text`} component="div">
                        {getHelperText(
                          meta,
                          <Box
                            sx={{
                              display: 'flex',
                              justifyContent: 'space-between',
                              mb: 1,
                            }}
                          >
                            <div>{helperText}</div>
                            {showCharacterCount && val ? (
                              <div>
                                <FormattedNumber value={val.characterCount} />{' '}
                                characters
                              </div>
                            ) : undefined}
                          </Box>
```

src/components/RichText/RichTextField.tsx
andrewmurraydavid: See my comments above 😄

andrewmurraydavid (approved):
@CarsonF overall looks good. I added my input on how the helper text and the chars counter could live together

src/components/RichText/RichTextField.tsx
CarsonF: Ok you convinced me lol

andrewmurraydavid (approved):
Looks good @CarsonF! Thanks!
Let's ship it 😛

PR #1337, opened 2022-12-06
Added engagement description on the detail page

andrewmurraydavid:
Added the description box on the detail page.

A/C: 
- [x] can view if proper permissions
- [x] can edit if proper permissions

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3629677427) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/RichText/RichTextField.tsx
+++ b/src/components/RichText/RichTextField.tsx
@@ -300,7 +300,7 @@ const Loading = ({
 }) => {
   // eslint-disable-next-line react/display-name
   const Input = forwardRef(({ className }: StyleProps, ref) => (
-    <Box ref={ref as any} className={className} sx={{ minHeight: 79 }}>
+    <Box ref={ref as any} className={className} sx={{ minHeight: 45 }}>
       <RichTextView data={value} />
     </Box>
   ));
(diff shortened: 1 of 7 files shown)


CarsonF (approved):
Thanks @andrewmurraydavid! I did some polishing that...kinda snowballed. Here are some notes of changes and why.

- I moved the edit & save buttons to icons next to the title. I think this looks nicer, and takes up less space.
- Tweaked the empty text. This was mainly because I was seeing "Description" "description" "description", which felt clunky.
- Moved header into the component. These are details the owner shouldn't be concerned with.
- Gave a type for the form shape. values are `any` without it.
- Wrapped the `initialValues` object in a memo. See #421 
- Awaited mutation call. The form was closing and showing previous values while the submission was happening. mutations should always be awaited.
- Only calling the mutation if the form is dirty. This is something I abstracted in the `DialogForm`, but that seems to have fallen out of favor.
- Using `Stack` instead of `Grid` as the API is simpler for these cases where..we just need to stack. It's new in V5 I think which is why it's not used much in the app.


Screen Shot 2022-12-19 at 7 15 52 PM
Screen Shot 2022-12-19 at 7 16 05 PM
Screen Shot 2022-12-19 at 7 16 13 PM

andrewmurraydavid:
Looks good to me! Thanks for these updates @CarsonF!

PR #1338, opened 2022-12-07
added the marketing location field

willdch:
Started the process of adding the marketing location field.  For now it is just a field added within the project region.


CarsonF (approved):
Thanks @willdch!

PR #1340, opened 2022-12-10
moved the marketing location to the engagement level

willdch:
This is not complete but I would like to get some eyes on it. 
This was requested to be moved from the project level to the engagement level.  
I still need to figure out the `Marketing Region` field.  
Seth said the data for the `Marketing Location`  and `Field Location` should be the same but the `Marketing Region` and the `Field Region` are different.  He was not sure if we collect the `Marketing Region` data anywhere.


CarsonF:
Closing since we decided to keep it at the project level

PR #1343, opened 2022-12-14
Refactor Steps to single declaration & untangle usages

CarsonF:


andrewmurraydavid:
@CarsonF just FYI, not sure what it is, but something doesn't feel right. I haven't ignored or forgotten of this PR. I've been sleeping on it until I can think of an approach I feel comfortable with or I can understand why I don't feel comfortable with it

CarsonF:
Yeah I had to table this as well for more pressing things.

My goal here is to unify the steps, their labels, and list & group order.
Which I've done here very concisely [here](https://github.com/SeedCompany/cord-field/pull/1343/files#diff-64d3d7c8c2418f8496a65749e3f80e2250dc5230ccb70279f4272c4cea5a9b73R11-R19) that replaces [this](https://github.com/SeedCompany/cord-field/pull/1343/files#diff-41de3c5c95625be27d6e05ef00212ed49b4acb95e7373c707cf3abf02b4a855aL13-L25) and [this](https://github.com/SeedCompany/cord-field/pull/1343/files#diff-aaa34bf1343ed42c590d7d68dbce2c265dfe9e6ee2dda5f92df5030d12452375L35-L39).

The kebab case version of the step can just be [calculated](https://github.com/SeedCompany/cord-field/pull/1343/files#diff-41de3c5c95625be27d6e05ef00212ed49b4acb95e7373c707cf3abf02b4a855aR36) from the label.

The second goal was to stop saving the [step index state](https://github.com/SeedCompany/cord-field/pull/1343/files#diff-41de3c5c95625be27d6e05ef00212ed49b4acb95e7373c707cf3abf02b4a855aL52), since is only a reflection of the step label from the url. That removes the need for this [useEffect](https://github.com/SeedCompany/cord-field/pull/1343/files#diff-41de3c5c95625be27d6e05ef00212ed49b4acb95e7373c707cf3abf02b4a855aL84), which the community is beginning to understand is bad.

The other biggest change was to not bother with `Stepper.activeStep`. Now instead of having to calculate the flat index of the step, we just use the label. This is easier since each step knows its label and it's independent of the other steps.
https://github.com/SeedCompany/cord-field/pull/1343/files#diff-128a5417f9a27cfd5db6128a6e6c67fd9d5590b78656442318532a83f5aab4b3R83-R84

All of this I'm happy with and is finished IMO.

---

CI however, now informs us of the circular dependency.
This is because the `ProgressReportContext` is now aware/depends on the each step. But each step also need the context.

`ProgressReportContext` is really doing two things: holding the report object, providing helpers for the current step state. 
Breaking this apart would fix the circular dependency.

Each step doesn't need be aware of how to move between steps, they just need the report.
And the navigation components probably don't need the report, they just need to navigate between steps.
Both of these assumptions could be false, but it still stands that the context is breaking _single responsibility principle_ and so this is still a good direction.

My first thought is just to pass the report directly as a prop. I don't think we need to forward it more than 1/2 layers and it's just a single prop so this seems the simplest.
`ProgressReportContext` could be renamed to `ProgressReportStepContext` then.

CarsonF:
I fixed the circular dependencies here by:
- Doing what I said above - moving the report to a straight prop instead of storing in context
- Moving the NextStepButton out of steps and into StepContainer.

Now the steps rightfully don't know anything about the outer layout / other steps

PR #1344, opened 2022-12-15
More robust MIME determination

CarsonF:
Fall back on the file name & even binary data to determine.

I've noticed these coming cross blank for even `.xlsm` files so clearly the OS is unreliable in giving us this info correctly.


andrewmurraydavid (approved):
LGTM 👍

PR #1346, opened 2022-12-19
Added ability to change prompts after selection

andrewmurraydavid:
Added a button under the prompt to be able to change the prompt. That button will take you to the prompt selection which will behave as before.


CarsonF (approved):
Thanks @andrewmurraydavid!

PR #1349, opened 2022-12-20
Reverse sort response variants & conditionally expand

andrewmurraydavid:


CarsonF (approved):
Thanks @andrewmurraydavid! I tacked on a few of my own

PR #1351, opened 2022-12-20
Refactor prompt response steps

CarsonF:
- Moved fragments out of `ProgressReportEdit` and into common spot. (If they were more specific I would've moved them into the the specific step folder).
- Renamed the fragments to not be tied to `ProgressReport` as they are completely independent. 
- Split `PromptVariantStep` into `Prompt` and `VariantResponses` components. The former had a large number of inputs and was only forwarding to these two, so it was the wrong component boundary.
- Moved prompt editing from `VariantResponsesForm` to `Prompt`. I think it's fine to show prompts while still showing the responses.
- Forms do dirty check before API call
- Forms have strict types now
- The existing prompt is now selected on edit
- The prompt edit button is a pencil next to the prompt, which is more consistent with the rest of the app and takes up less space.


andrewmurraydavid (approved):
@CarsonF, thanks for these refactors!
Looks good to me!

PR #1361, opened 2023-01-09
Log when there's an API error

CarsonF:
This way LogRocket can get more info


CarsonF:
Cherry picked 3116e77aaad92491aee3d50a6c2699fdaca0dede

PR #1381, opened 2023-01-19
Skip progress report steps if user cannot read step data

CarsonF:


andrewmurraydavid (approved):
Thanks for making this cleaner @CarsonF

PR #1384, opened 2023-01-19
Add real Progress Report Instructions

sethmcknight:
Replaced the Lorem Ipsom in the Instructions Dialog with a draft of real Quarterly Report instructions.

### DIFF
--- a/src/scenes/ProgressReports/EditForm/InstructionsDialog.tsx
+++ b/src/scenes/ProgressReports/EditForm/InstructionsDialog.tsx
@@ -4,52 +4,176 @@ import {
   DialogActions,
   DialogContent,
   DialogProps,
+  DialogTitle,
   Typography,
 } from '@mui/material';
+import { useSession } from '../../../components/Session';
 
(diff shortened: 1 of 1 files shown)


src/scenes/ProgressReports/EditForm/InstructionsDialog.tsx
```diff
@@ -20,31 +20,75 @@ export const InstructionsDialog = ({ open, onClose }: DialogProps) => {
...
+          Community Story
+            - Review and scrub the Field Partner and Translation variants.
+            - Please provide any additional context, correction, or clarity on 
+              the story submitted by the Field Partner - write out acronyms, 
+              clarify names & their role with the project, and add context that 
+              was assumed in the Partners submission.
+            - This information will provide context for the Communications Writer 
+              to write and edit the story.
+            - No input is needed if the story is clear in the Field Partner or 
+              Translation variant.
```
CarsonF, line 64: @sethmcknight We do not assume nothing means "take from previous". If the story is clear the expectation is they copy it into the FPM variant. Is that ok?

PR #1402, opened 2023-01-26
Progress Report Edit layout tweaks

willdch:
[Monday.com Ticket](https://seed-company-squad.monday.com/boards/3451697530/views/87873996/pulses/3830491648)

### updates made:  
- Added "previous" step button
- Changed “Back” to “Back to Overview"
- Changed the order of location, sensitivity, project name and language name
- moved status pill to be under instructions link above the stepper


src/scenes/ProgressReports/EditForm/ProgressReportDrawerHeader.tsx
```diff
@@ -51,54 +39,71 @@ export const ProgressReportDrawerHeader = ({ report }: ReportProp) => {
...
+          </Box>
         </Card>
       </Box>
+      <Box sx={{ display: 'flex', mt: 2 }}>
+        <Typography variant="h5">{project.name.value}</Typography>
+        <Divider
+          orientation="vertical"
+          flexItem
+          sx={{ mx: 2, borderRightColor: 'black', borderRightWidth: 2 }}
+        />
```
CarsonF, line 94: If we want the divider to look like like the text I would just use the pipe character instead.
```tsx
<Typography variant="h5">
{project.name.value}
|
{language.value?.displayName.value}
</Typography>
```
(still might need to tweak spacing, with `{" "}` or maybe `&nbsp;`)

Otherwise I'd use the `Divider` component as is. If we don't like the way the Divider looks we can change it in the theme.
Using hardcoded colors is a code smell. The text next to it is not `"black"` either, it's a shade of black.

I'd also use a `Stack` here instead. Check out their [example which includes dividers](https://mui.com/material-ui/react-stack/#dividers).

src/scenes/ProgressReports/EditForm/ProgressReportSidebar.tsx
```diff
@@ -29,6 +30,11 @@ export const ProgressReportSidebar = ({ report }: ReportProp) => {
       >
         Instructions
       </Button>
+      <Box mb={1}>
```
CarsonF, line 33: This leaves zero spacing between the instructions button.

Screenshot 2023-02-01 at 3 08 02 PM

Also the Box is still rendered even without the chip - it should be moved into the conditional.

src/scenes/ProgressReports/EditForm/Steps/PreviousStepButton.tsx
```diff
@@ -0,0 +1,33 @@
+import { ArrowBack } from '@mui/icons-material';
+import { Button } from '@mui/material';
+import { StyleProps } from '~/common';
+import { useProgressReportContext } from '../ProgressReportContext';
+
+export const PreviousStepButton = ({ sx }: StyleProps) => {
```
CarsonF, line 6: This is a copy paste of the next button. Both are fairly worthless now since they are only used in the `StepContainer` component. I'd delete both of these and render there.

You could still assign them to variables to reference twice.

```tsx 
const previousButton = !isFirst && (
  <Button
    variant="outlined"
    color="secondary"
    onClick={previousStep}
    startIcon={<ArrowBack />}
  >
    Previous
  </Button>
);

return (
  <Box ...>{previousButton}{nextButton}</Box>
  ...
  <Box ...>{previousButton}{nextButton}</Box>
);
```

CarsonF (changes requested):

src/scenes/ProgressReports/EditForm/ProgressReportDrawerHeader.tsx
```diff
@@ -51,54 +39,71 @@ export const ProgressReportDrawerHeader = ({ report }: ReportProp) => {
...
-                >
-                  |
-                </span>
-              )}
-              <Typography variant="body1">
-                {project.fieldRegion.value?.name.value}
-              </Typography>
-            </div>
-          )}
+          <Box sx={{ display: 'flex', alignItems: 'center', ml: -0.3 }}>
```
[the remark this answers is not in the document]
willdch: This one was taken care of when I replaced the Sensitivity and Location cards with the DataButton component.

src/scenes/ProgressReports/EditForm/ProgressReportDrawerHeader.tsx
```diff
@@ -51,54 +39,71 @@ export const ProgressReportDrawerHeader = ({ report }: ReportProp) => {
...
+            <MapPinIcon color="info" />
+            {project.primaryLocation.value?.name.canRead && (
+              <div css={{ display: 'flex' }}>
+                <Typography variant="body1" sx={{ ml: 0.8 }}>
+                  {project.primaryLocation.value.name.value}
+                </Typography>
+                {project.fieldRegion.value && (
```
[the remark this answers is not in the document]
willdch: This was also taken care of when I replaced the Sensitivity and Location cards with the DataButton component.

src/scenes/ProgressReports/EditForm/ProgressReportDrawerHeader.tsx
willdch: I updated this to use the Stack and divider like the link you provided shows.   I think it made the spacing look better too.

CarsonF (approved):

PR #1418, opened 2023-02-15
Progress Reports workflow history

willdch:
This PR adds the date, time, and the user who made a status change to the bottom StatusStepper. 

https://seed-company-squad.monday.com/boards/3451697530/pulses/3921175869

### DIFF
--- a/src/components/ProgressReportListBreadcrumb/ProgressReportListBreadcrumb.tsx
+++ b/src/components/ProgressReportListBreadcrumb/ProgressReportListBreadcrumb.tsx
@@ -1,0 +1,24 @@
+import { Skeleton } from '@mui/material';
+import { IdFragment } from '~/common';
+import { Breadcrumb, BreadcrumbProps } from '../Breadcrumb';
+import { idForUrl } from '../Changeset';
+
+interface ProgressReportListBreadcrumbProps extends BreadcrumbProps {
+  engagement?: IdFragment;
+}
+
+export const ProgressReportListBreadcrumb = ({
+  engagement,
+  ...rest
+}: ProgressReportListBreadcrumbProps) => (
+  <Breadcrumb
+    to={
+      engagement
+        ? `/engagements/${idForUrl(engagement)}/reports/progress`
+        : undefined
+    }
+    {...rest}
+  >
+    {engagement ? 'Quarterly Reports' : <Skeleton width={200} />}
+  </Breadcrumb>
+);
--- a/src/scenes/ProgressReports/Detail/ProgressReportDetail.tsx
+++ b/src/scenes/ProgressReports/Detail/ProgressReportDetail.tsx
@@ -12,11 +12,9 @@ import {
 } from '@mui/material';
 import { Helmet } from 'react-helmet-async';
 import { makeStyles } from 'tss-react/mui';
-import { Breadcrumb } from '../../../components/Breadcrumb';
-import {
-  idForUrl,
-  useChangesetAwareIdFromUrl,
-} from '../../../components/Changeset';
+import { ProgressReportBreadcrumb } from '~/components/ProgressReportBreadcrumb';
+import { ProgressReportListBreadcrumb } from '~/components/ProgressReportListBreadcrumb';
+import { useChangesetAwareIdFromUrl } from '../../../components/Changeset';
 import { EngagementBreadcrumb } from '../../../components/EngagementBreadcrumb';
 import { Error } from '../../../components/Error';
 import { ReportLabel } from '../../../components/PeriodicReports/ReportLabel';
@@ -91,23 +89,11 @@ export const ProgressReportDetail = () => {
           children={[
             <ProjectBreadcrumb key="project" data={engagement?.project} />,
             <EngagementBreadcrumb key="engagement" data={engagement} />,
-            <Breadcrumb
+            <ProgressReportListBreadcrumb
               key="report-list"
-              to={
-                engagement
-                  ? `/engagements/${idForUrl(engagement)}/reports/progress`
-                  : undefined
-              }
-            >
-              {!report ? <Skeleton width={200} /> : 'Quarterly Reports'}
-            </Breadcrumb>,
-            <Breadcrumb key="report" to=".">
-              {!report ? (
-                <Skeleton width={200} />
-              ) : (
-                <ReportLabel report={report} />
-              )}
-            </Breadcrumb>,
+              engagement={engagement}
+            />,
+            <ProgressReportBreadcrumb data={report} key="report" />,
           ]}
         />
 
--- a/src/scenes/ProgressReports/Detail/ProgressReportDetail.tsx
+++ b/src/scenes/ProgressReports/Detail/ProgressReportDetail.tsx
@@ -28,7 +28,7 @@ import {
 } from './ProgressReportDetail.graphql';
 import { ProgressSummaryCard } from './ProgressSummaryCard';
 import { PromptResponseCard } from './PromptResponseCard';
-import { StatusStepper } from './StatusStepper';
+import { WorkflowCard } from './WorkflowCard';
 
 const useStyles = makeStyles()(({ spacing }) => ({
   root: {
@@ -139,12 +139,7 @@ export const ProgressReportDetail = () => {
           </Grid>
         ) : (
           <Stack spacing={3} maxWidth="lg" alignItems="flex-start" mt={1}>
-            <StatusStepper
-              loading={!report}
-              current={report?.status.value}
-              sx={{ width: 1, maxWidth: 'sm' }}
-            />
-
+            <WorkflowCard report={report} sx={{ width: 1, maxWidth: 'sm' }} />
             <ProgressSummaryCard
               loading={!report}
               summary={report?.cumulativeSummary ?? null}
--- a/src/scenes/ProgressReports/Detail/StatusStepper.tsx
+++ b/src/scenes/ProgressReports/Detail/StatusStepper.tsx
@@ -1,5 +1,4 @@
 import {
-  Card,
   Skeleton,
   Step,
   StepButton,
@@ -41,7 +40,6 @@ export const StatusStepper = ({
 
   return (
     <Stepper
-      component={Card}
       activeStep={currentIndex}
       alternativeLabel={!mobile}
       orientation={mobile ? 'vertical' : 'horizontal'}
--- a/src/scenes/ProgressReports/Detail/WorkflowCard.tsx
+++ b/src/scenes/ProgressReports/Detail/WorkflowCard.tsx
@@ -1,0 +1,54 @@
+import { Card, CardActions, CardProps, Typography } from '@mui/material';
+import { RelativeDateTime } from '~/components/Formatters';
+import { ButtonLink } from '~/components/Routing';
+import { ProgressReportDetailFragment } from './ProgressReportDetail.graphql';
+import { StatusStepper } from './StatusStepper';
+
+type WorkflowCardProps = {
+  report?: ProgressReportDetailFragment;
+} & CardProps;
+
+export const WorkflowCard = ({ report, ...rest }: WorkflowCardProps) => {
+  const lastWorkflowEvent = report?.workflowEvents.at(
+    report.workflowEvents.length - 1
+  );
+  return (
+    <Card {...rest}>
+      <Typography variant="h3" sx={{ pl: 2, pt: 2 }}>
+        Status
+      </Typography>
+      <StatusStepper loading={!report} current={report?.status.value} />
+      {lastWorkflowEvent && (
+        <CardActions
+          sx={{
+            display: 'flex',
+            alignItems: 'center',
+            justifyContent: 'space-between',
+          }}
+        >
+          <ButtonLink
+            to="workflow-history"
+            sx={{ whiteSpace: 'nowrap', minWidth: 'max-content' }}
+          >
+            View History
+          </ButtonLink>
+          <Typography
+            variant="caption"
+            sx={{
+              color: 'text.secondary',
+              pr: 1,
+              pl: { sm: 5 },
+              display: { xs: 'flex', sm: 'block' },
+              flexDirection: 'column',
+              justifyContent: 'center',
+              alignItems: 'center',
+            }}
+          >
+            Updated by {lastWorkflowEvent.who.value?.fullName}{' '}
+            <RelativeDateTime date={lastWorkflowEvent.at} />
+          </Typography>
+        </CardActions>
+      )}
+    </Card>
+  );
+};
--- a/src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/TransitionProgressReport.graphql
+++ b/src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/TransitionProgressReport.graphql
@@ -6,5 +6,9 @@ mutation TransitionProgressReport(
     status {
       ...ProgressReportStatus
     }
+    # This is needed for updating the cache for the workflow events
+    workflowEvents {
+      ...workflowEvent
+    }
   }
 }
--- a/src/scenes/ProgressReports/ProgressReportRouter.tsx
+++ b/src/scenes/ProgressReports/ProgressReportRouter.tsx
@@ -2,9 +2,11 @@ import { Route, Routes } from 'react-router-dom';
 import { ChangesetContext } from '../../components/Changeset';
 import { NotFoundRoute } from '../../components/Error';
 import { ProgressReportDetail as ProgressReport } from './Detail';
+import { StatusHistory } from './StatusHistory';
 
 export const ProgressReportRouter = () => (
   <Routes>
+    <Route path=":reportId/workflow-history" element={<StatusHistory />} />
     <Route path=":reportId/*" element={<Detail />} />
     {NotFoundRoute}
   </Routes>
--- a/src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
+++ b/src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
@@ -1,0 +1,114 @@
+import { useQuery } from '@apollo/client';
+import {
+  Box,
+  Breadcrumbs,
+  FormControlLabel,
+  Skeleton,
+  Switch,
+  Typography,
+} from '@mui/material';
+import { useToggle } from 'ahooks';
+import { Helmet } from 'react-helmet-async';
+import { flexColumn } from '~/common';
+import { Breadcrumb } from '~/components/Breadcrumb';
+import { useChangesetAwareIdFromUrl } from '~/components/Changeset';
+import { EngagementBreadcrumb } from '~/components/EngagementBreadcrumb';
+import { getReportLabel } from '~/components/PeriodicReports/ReportLabel';
+import { ProgressReportBreadcrumb } from '~/components/ProgressReportBreadcrumb';
+import { ProgressReportListBreadcrumb } from '~/components/ProgressReportListBreadcrumb';
+import { ProjectBreadcrumb } from '~/components/ProjectBreadcrumb';
+import { Navigate } from '~/components/Routing';
+import { Error } from '../../../components/Error';
+import {
+  ProgressReportDetailDocument,
+  ProgressReportDetailFragment,
+} from '../Detail/ProgressReportDetail.graphql';
+import { WorkFlowEventList } from './WorkflowEventList';
+
+export const StatusHistory = () => {
+  const { id, changesetId } = useChangesetAwareIdFromUrl('reportId');
+  const [showNotes, setShowNotes] = useToggle(true);
+
+  const { data, error } = useQuery(ProgressReportDetailDocument, {
+    variables: {
+      id,
+      changesetId,
+    },
+  });
+  if (error) {
+    return (
+      <Error page error={error}>
+        {{
+          NotFound: 'Could not find quarterly report',
+          Default: "Error loading quarterly report's status history",
+        }}
+      </Error>
+    );
+  }
+
+  const report = data?.periodicReport as
+    | ProgressReportDetailFragment
+    | undefined;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- safety check. It's possible with manual user input.
+  if (report && report.__typename !== 'ProgressReport') {
+    return <Navigate to="/" replace />;
+  }
+
+  const engagement = report?.parent;
+
+  const reportTitle = getReportLabel(report);
+  const engagementTitle =
+    engagement?.language.value?.name.value ??
+    engagement?.language.value?.displayName.value;
+
+  return (
+    <Box
+      component="main"
+      css={flexColumn}
+      sx={{ p: 4, gap: 2, overflowY: 'auto' }}
+    >
+      <Helmet
+        title={
+          !report ? 'Status History' : `${reportTitle} - ${engagementTitle}`
+        }
+      />
+      <Breadcrumbs
+        children={[
+          <ProjectBreadcrumb key="project" data={engagement?.project} />,
+          <EngagementBreadcrumb key="engagement" data={engagement} />,
+          <ProgressReportListBreadcrumb
+            key="report-list"
+            engagement={engagement}
+          />,
+          <ProgressReportBreadcrumb data={report} key="report" />,
+          <Breadcrumb key="status-history" to=".">
+            {!report ? <Skeleton width={200} /> : 'Status History'}
+          </Breadcrumb>,
+        ]}
+      />
+      <Box
+        sx={{
+          display: 'flex',
+          justifyContent: 'space-between',
+          alignItems: 'center',
+        }}
+      >
+        <Typography variant="h3" paragraph>
+          Status History
+        </Typography>
+        <FormControlLabel
+          control={
+            <Switch checked={showNotes} onChange={setShowNotes.toggle} />
+          }
+          label="Show Notes"
+          labelPlacement="start"
+          sx={{ gap: 1 }}
+        />
+      </Box>
+      <WorkFlowEventList
+        events={report?.workflowEvents}
+        showNotes={showNotes}
+      />
+    </Box>
+  );
+};
--- a/src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
+++ b/src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
@@ -93,7 +93,7 @@ export const StatusHistory = () => {
           alignItems: 'center',
         }}
       >
-        <Typography variant="h3" paragraph>
+        <Typography component="h1" variant="h3">
           Status History
         </Typography>
         <FormControlLabel
--- a/src/components/PeriodicReports/ReportLabel.tsx
+++ b/src/components/PeriodicReports/ReportLabel.tsx
@@ -1,4 +1,4 @@
-import { CalendarDate, isSecured, Nullable, SecuredProp } from '~/common';
+import { isSecured, Nullable, SecuredProp } from '~/common';
 import { Redacted } from '../Redacted';
 import { PeriodicReportFragment } from './PeriodicReport.graphql';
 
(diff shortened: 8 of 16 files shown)


src/scenes/ProgressReports/Detail/StatusStepper.tsx
```diff
@@ -39,53 +48,103 @@ export const StatusStepper = ({
   currentIndex =
     currentIndex === Statuses.length - 1 ? currentIndex + 1 : currentIndex;
 
+  const workflowEvents = report?.workflowEvents.filter(
+    (event) => event.status === report.status.value
+  );
```
CarsonF, line 53: What's the reason for this?

src/scenes/ProgressReports/Detail/StatusStepper.tsx
```diff
@@ -39,53 +48,103 @@ export const StatusStepper = ({
...
+  const currentWorkflowEvent = orderBy(workflowEvents, 'at', 'desc')[0];
```
CarsonF, line 54: The API orders the list asc. You could just grab the last one.

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.graphql
```diff
@@ -0,0 +1,25 @@
...
+    value
+  }
+  who {
+    canEdit
+    canRead
+    value {
+      fullName
+      title {
+        value
+      }
```
CarsonF, line 22: Doesn't look like `title` is used anywhere.

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.graphql
```diff
@@ -0,0 +1,25 @@
+fragment workflowEvent on ProgressReportWorkflowEvent {
```
CarsonF, line 1: I'd rename this file to `WorkflowEvent.graphql` to match the fragment name.

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.tsx
```diff
@@ -0,0 +1,76 @@
...
+  checked?: boolean;
+}
+
+export const WorkFlowEventHistory = ({
+  checked,
+  report,
+}: WorkFlowEventHistoryProps) => {
+  if (!report?.workflowEvents) {
+    return <Skeleton variant="rectangular" width="55%" />;
+  } else if (report.workflowEvents.length === 0) {
```
CarsonF, line 21: Not necessarily true. It could just be empty. I don't know if we truly have a way to know this. I'd probably just skip this check.

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.tsx
```diff
@@ -0,0 +1,76 @@
...
+              borderRadius: 1,
+            }}
+          >
+            <Box sx={{ display: 'flex', alignItems: 'center' }}>
+              <Typography variant="body1" sx={{ fontWeight: 'bold' }}>
+                {StatusLabels[event.status]}
+              </Typography>
+              &nbsp;
+              <East />
+              &nbsp;
```
CarsonF, line 56:
```suggestion
              <ArrowForward aria-label="transitioned to" sx={{ mx: 1 }} />
```
Arrow forward more represents our intent here instead of Map East. The icons are only slightly different so this is mostly just on principle.
Using `aria-label` and margin is better for screen readers.

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.tsx
```diff
@@ -0,0 +1,76 @@
...
+    <>
+      {statusEvents.slice(0, statusEvents.length - 1).map((event, i) => (
+        <Fragment key={i}>
+          <Box
+            sx={{
+              py: 2,
+              px: 3,
+              bgcolor: 'primary.contrastText',
+              borderRadius: 1,
+            }}
```
CarsonF, line 48: I'd use a single Card here instead of multiple Boxes. Couple it with a `Stack` to help with the dividers.
```tsx
<Card>
  <Stack divider={<Divider />} spacing={2}>
    {events.map(...)}
  </Stack>
</Card>

```

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.tsx
```diff
@@ -0,0 +1,76 @@
...
+          >
+            <Box sx={{ display: 'flex', alignItems: 'center' }}>
+              <Typography variant="body1" sx={{ fontWeight: 'bold' }}>
```
CarsonF, line 51: I wouldn't try to bold body1 here. Let body1 be body1 how the theme defines it. h4 variant more closing matches body1 bold and probably fits the intent better. 
It would also be better semantically & to be DRY to merge them together.
```tsx
<Typography variant="h4" sx={{ display: 'flex', alignItems: 'center' }}>
  {from}
  <Arrow>
  {to}
</Typography>
```

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.tsx
```diff
@@ -0,0 +1,76 @@
...
+                {StatusLabels[statusEvents[i + 1]!.status]}
+              </Typography>
+              <Typography variant="caption" sx={{ ml: 1 }}>
+                {event.who.value?.fullName} &nbsp;
+                <FormattedDate date={event.at} />
+              </Typography>
+            </Box>
+            <Collapse in={!checked}>
+              <Box fontStyle="italic" sx={{ px: 3, pt: 1 }}>
+                <RichTextView data={event.notes.value} />
```
CarsonF, line 67: I don't think forcing styling here is a good idea. The user could italic pieces of their notes and those distinctions would be lost with this.

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.tsx
```diff
@@ -0,0 +1,76 @@
...
+              </Box>
+            </Collapse>
+          </Box>
+          <Divider sx={{ width: '95%', margin: 'auto' }} />
```
CarsonF, line 71: `95%` here is a magic number. Percentages especially used like this are a bad idea. The amount of gap changes based on the container (& window in this case) size.
Instead you want a fixed margin on either side.

In this case though with the divider component you can just do
```tsx
<Divider variant="middle" />
```

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/TransitionProgressReport.graphql
```diff
@@ -6,5 +6,8 @@ mutation TransitionProgressReport(
     status {
       ...ProgressReportStatus
     }
+    workflowEvents {
```
CarsonF, line 9: I'd add a comment here saying grab the updated list from the change. Just to be extra clear that this isn't directly needed, but instead to provide and updated cache value for other pages.

src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
```diff
@@ -0,0 +1,117 @@
...
+              to={
+                engagement ? `/progress-reports/${idForUrl(report)}` : undefined
+              }
+            >
+              {!report ? (
+                <Skeleton width={200} />
+              ) : (
+                <ReportLabel report={report} />
+              )}
+            </Breadcrumb>,
```
CarsonF, line 86: I'd create a `ProgressReportListBreadcrumb` and `ProgressReportBreadcrumb` following suit of Project & Engagement breadcrumbs. So they can be more easily used in two spots.

src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
```diff
@@ -0,0 +1,117 @@
...
+    | undefined;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- safety check. It's possible with manual user input.
+  if (report && report.__typename !== 'ProgressReport') {
+    return <Navigate to="/" replace />;
+  }
+
+  const engagement = report?.parent;
+
+  return (
+    <div css={{ overflowY: 'auto' }}>
```
CarsonF, line 58: This can be merged into the box below. No need for separate div.

src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
```diff
@@ -0,0 +1,117 @@
...
+      <Box component="main" sx={{ ...flexColumnStyle, p: 4 }}>
```
CarsonF, line 59: Ah here's why that change was made. Instead:
```suggestion
      <Box component="main" css={flexColumn} sx={{ p: 4 }}>
```

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { Fragment } from 'react';
+import { ProgressReportStatusLabels as StatusLabels } from '~/api/schema/enumLists';
+import { FormattedDate } from '~/components/Formatters';
+import { Redacted } from '~/components/Redacted';
+import { RichTextView } from '~/components/RichText';
+import { ProgressReportDetailFragment } from './ProgressReportDetail.graphql';
+
+interface WorkFlowEventHistoryProps {
+  report?: ProgressReportDetailFragment;
+  checked?: boolean;
```
CarsonF, line 12: Checked doesn't describe what this boolean is for at all.
```suggestion
  showNotes?: boolean;
```

src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
```diff
@@ -0,0 +1,117 @@
...
+      id,
+      changesetId,
+    },
+  });
+  if (error) {
+    return (
+      <Error page error={error}>
+        {{
+          NotFound: 'Could not find status history',
+          Default: 'Error loading status history',
```
CarsonF, line 41: It's the progress report that could not be found. I would change the error message to that.
```suggestion
          NotFound: 'Could not find quarterly report',
          Default: 'Error loading quarterly report's status history',
```

src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
```diff
@@ -0,0 +1,117 @@
...
+            pt: 2,
+          }}
+        >
+          <Typography variant="h3" paragraph sx={{ pt: 2 }}>
+            Status History
+          </Typography>
+          <FormControlLabel
+            control={
+              <>
+                <Typography sx={{ pr: 1 }}>Hide Notes</Typography>
```
CarsonF, line 106: I would not have this toggle be a negative. Instead I would say "Show Notes" and default it according to whether we want them shown or hidden by default.

Also why is this here instead of in the `label` prop?
...Oh it's for placement. Instead you can use `labelPlacement="start"`.
If you are looking for more spacing between the switch, then `sx={{ gap: 1 }}` will do the trick.

I'd keep the switch the default size as well. We are not strapped for space here.

src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
```diff
@@ -0,0 +1,117 @@
...
+  if (report && report.__typename !== 'ProgressReport') {
+    return <Navigate to="/" replace />;
+  }
+
+  const engagement = report?.parent;
+
+  return (
+    <div css={{ overflowY: 'auto' }}>
+      <Box component="main" sx={{ ...flexColumnStyle, p: 4 }}>
+        <Helmet title="Status History" />
```
CarsonF, line 60: It would be good to include more info here. Specifically the engagement name & quarter will help identify these pages in browser history.
Currently it's 
Screenshot 2023-02-21 at 11 50 11 AM

Something like this would be better:
```
Status History - Q1 FY2023 - Engagement Name
Q1 FY2023 - Engagement Name
```

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export const WorkFlowEventHistory = ({
```
CarsonF, line 15: Maybe calling this `WorkflowEventList` would help distinguish it from the "history" page.
Be sure the file & prop names match the component name too.

src/scenes/ProgressReports/ProgressReportRouter.tsx
```diff
@@ -2,9 +2,11 @@ import { Route, Routes } from 'react-router-dom';
 import { ChangesetContext } from '../../components/Changeset';
 import { NotFoundRoute } from '../../components/Error';
 import { ProgressReportDetail as ProgressReport } from './Detail';
+import { StatusHistory } from './StatusHistory';
 
 export const ProgressReportRouter = () => (
   <Routes>
+    <Route path=":reportId/history" element={<StatusHistory />} />
```
CarsonF, line 9: "history" is pretty ambiguous. I'd go with "workflow-events" or "workflow-history"

CarsonF (changes requested):
Thanks guys. Here are notes. Happy to talk in Slack/Zoom about them too.

src/scenes/ProgressReports/Detail/StatusStepper.tsx
willdch: We used this to get the last item in the array and used it to display who and when the last change was made at the bottom of the stepper.

CarsonF: I don't see the need to filter the events though

src/scenes/ProgressReports/Detail/WorkflowEventsHistory.tsx
CarsonF: https://youtu.be/I4gUvhG7uFU

src/common/reports.ts
```diff
@@ -0,0 +1,10 @@
+import { CalendarDate } from './CalenderDate';
+
+export const getReportQuarter = (start?: CalendarDate, end?: CalendarDate) => {
+  if (!start || !end) return 'Unknown';
```
CarsonF, line 4: I'd keep this next to the `ReportLabel` component and match it's signature. It's pretty specific to PeriodicReports, so I don't think it belongs in common.
```suggestion
export const getReportLabel = ({ start, end }: Report) => {
  if (!start || !end) return null;
```
Report above coming from https://github.com/SeedCompany/cord-field/blob/ecd9f2ccdb23b384b643371139943672694e0d92/src/components/PeriodicReports/ReportLabel.tsx#L5-L5

src/components/ProgressReportListBreadcrumb/ProgressReportListBreadcrumb.tsx
```diff
@@ -0,0 +1,22 @@
+import { Skeleton } from '@mui/material';
+import { IdFragment } from '~/common';
+import { Breadcrumb, BreadcrumbProps } from '../Breadcrumb';
+import { idForUrl } from '../Changeset';
+
+interface ProgressReportListBreadcrumbProps extends Partial<BreadcrumbProps> {
+  data?: IdFragment;
```
CarsonF, line 7: I'd rename this for clarity.
```suggestion
  engagement?: IdFragment;
```

src/components/progressReportBreadcrumb/ProgressReportBreadcrumb.tsx
```diff
@@ -0,0 +1,20 @@
+import { Skeleton } from '@mui/material';
+import { Nullable } from '~/common';
+import { ProgressReportDetailFragment } from '~/scenes/ProgressReports/Detail/ProgressReportDetail.graphql';
+import { Breadcrumb } from '../Breadcrumb';
+import { idForUrl } from '../Changeset';
+import { ReportLabel } from '../PeriodicReports/ReportLabel';
+
+interface ProgressReportBreadcrumbProps {
+  data?: Nullable<ProgressReportDetailFragment>;
```
CarsonF, line 9: You've got the fragment declared right, you just need to use it here. Using the breadcrumb fragment doesn't cross boundaries like the ProgressReportDetail fragment does. That fragment is specific to a view in the scenes. This component should not care about that. Nothing outside of scenes/ should ever import from scenes/. This is why we make our own fragment here.
```suggestion
  data?: Nullable<ProgressReportBreadcrumbFragment>;
```

src/scenes/ProgressReports/Detail/ProgressReportDetail.tsx
```diff
@@ -10,13 +10,12 @@ import {
...
 } from '@mui/material';
+// import { orderBy } from 'lodash';
 import { Helmet } from 'react-helmet-async';
 import { makeStyles } from 'tss-react/mui';
-import { Breadcrumb } from '../../../components/Breadcrumb';
-import {
-  idForUrl,
-  useChangesetAwareIdFromUrl,
-} from '../../../components/Changeset';
+import { ProgressReportBreadcrumb } from '~/components/progressReportBreadcrumb';
```
CarsonF, line 16: Component folder names should be PascalCase.
```suggestion
import { ProgressReportBreadcrumb } from '~/components/ProgressReportBreadcrumb';
```

src/scenes/ProgressReports/Detail/StatusStepper.tsx
```diff
@@ -19,13 +18,13 @@ import { ProgressReportStatus as Status } from '~/api/schema/schema.graphql';
 import { extendSx } from '~/common';
 
 type StatusStepperProps = {
-  current?: Status | null;
   loading: boolean;
+  current?: Status;
```
CarsonF, line 22: This is what I was talking about on zoom yesterday. You made a change and then rolled it back but in a different way. This still results in a diff, but it's not actually changing anything.

src/scenes/ProgressReports/Detail/WorkflowCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const currentWorkflowEvent = report?.workflowEvents.at(
+    report.workflowEvents.length - 1
+  );
+  return (
+    <Card {...rest}>
+      <Typography variant="h3" sx={{ pl: 2, pt: 2 }}>
+        Status
+      </Typography>
+      <StatusStepper loading={!report} current={currentStatus} />
```
CarsonF, line 23: Inline currentStatus here, also the typecast is not needed.
```suggestion
      <StatusStepper loading={!report} current={report?.status.value} />
```

src/scenes/ProgressReports/Detail/WorkflowCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { StatusStepper } from './StatusStepper';
+
+type WorkflowCardProps = {
+  report?: ProgressReportDetailFragment;
+} & StepperProps;
+
+export const WorkflowCard = ({ report, ...rest }: WorkflowCardProps) => {
+  const currentStatus = report?.status.value as Status | undefined;
+
+  const currentWorkflowEvent = report?.workflowEvents.at(
```
CarsonF, line 15: Events are never current, they happen in an instant. Just like you wouldn't say "current update" but rather "last/latest update"

```suggestion
  const lastWorkflowEvent = report?.workflowEvents.at(
```

src/scenes/ProgressReports/Detail/WorkflowCard.tsx
```diff
@@ -0,0 +1,41 @@
+import { Card, CardActions, StepperProps, Typography } from '@mui/material';
+import { ProgressReportStatus as Status } from '~/api/schema/schema.graphql';
+import { RelativeDateTime } from '~/components/Formatters';
+import { ButtonLink } from '~/components/Routing';
+import { ProgressReportDetailFragment } from './ProgressReportDetail.graphql';
+import { StatusStepper } from './StatusStepper';
+
+type WorkflowCardProps = {
+  report?: ProgressReportDetailFragment;
+} & StepperProps;
```
CarsonF, line 10: This doesn't match where you spread the rest of the props.
```suggestion
} & CardProps;
```
```suggestion
} & StyleProps;
```

src/scenes/ProgressReports/Detail/WorkflowCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+      {currentWorkflowEvent && (
+        <CardActions
+          sx={{
+            display: 'flex',
+            alignItems: 'center',
+            justifyContent: 'space-between',
+          }}
+        >
+          <ButtonLink to="workflow-history">View History</ButtonLink>
+          <Typography variant="caption" sx={{ color: 'gray', pr: 1 }}>
```
CarsonF, line 33: Colors should never be hardcoded. This might not seem like you are referencing a browser preset, but it's no different than saying `#808080`.
```suggestion
          <Typography variant="body2" color="text.secondary" sx={{ pr: 1 }}>
```

src/scenes/ProgressReports/Detail/WorkflowCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <span>Updated by</span> {currentWorkflowEvent.who.value?.fullName}{' '}
```
CarsonF, line 34: What's this span doing?

src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return <Navigate to="/" replace />;
+  }
+
+  const engagement = report?.parent;
+
+  const reportTitle = getReportQuarter(report?.start, report?.end);
+  const engagementTitle = engagement?.language.value?.displayName.value ?? '';
+
+  return (
+    <Box component="main" css={flexColumn} sx={{ p: 4, overflowY: 'auto' }}>
```
CarsonF, line 63:
```suggestion
    <Box component="main" css={flexColumn} sx={{ p: 4, gap: 2, overflowY: 'auto' }}>
```
Instead of the 2 instances of `pt: 2` below

CarsonF (changes requested):

src/components/progressReportBreadcrumb/ProgressReportBreadcrumb.tsx
willdch: Wow, I need to pay closer attention to what intenseness is importing.  I thought I had the ProgressReportBreadcrumbFragment used here.  I trusted auto fill to quickly.

src/scenes/ProgressReports/Detail/WorkflowCard.tsx
willdch: Here, are you suggesting I choose one or use both of them?

CarsonF: One of them.

src/scenes/ProgressReports/Detail/WorkflowCard.tsx
willdch: I am not sure to be honest.  Looking at it now I see that it just ands another element to the DOM and is not necessary.

src/scenes/ProgressReports/Detail/WorkflowEventList.tsx
```diff
@@ -0,0 +1,68 @@
...
+  if (!report?.workflowEvents) {
+    return null;
+  }
+
+  const statusEvents = [
+    {
+      ...report.workflowEvents[0]!,
+      status: 'NotStarted' as keyof typeof StatusLabels,
+    },
+    ...report.workflowEvents,
```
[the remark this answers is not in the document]
willdch: I wasn't sure how to make this implementation work in this component how I have it setup.  I am using the .slice method with the map function to offset the maps event from the StatusEvent array so that on line 49 {StatusLabels[statusEvents[i + 1]!.status]} is not undefined.  If I move the .slice to line 34 the arrays is no longer offset and line 49 throws and error as Undefined because there is not another position in the array with a status property.  Do you know of a better way to handle this?   Andrew agreed that this wouldn't work (at least with my approach) but did help me refactor this and rename things to be more clear.

src/scenes/ProgressReports/StatusHistory/StatusHistory.tsx
```diff
@@ -0,0 +1,114 @@
...
+          Default: "Error loading quarterly report's status history",
+        }}
+      </Error>
+    );
+  }
+
+  const report = data?.periodicReport as
+    | ProgressReportDetailFragment
+    | undefined;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- safety check. It's possible with manual user input.
```
andrewmurraydavid, line 52: Maybe we don't make such a long comment and put the dev comment above the eslint comment. Something like this:
```suggestion
  //  safety check. It's possible with manual user input.
  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition 
 ```

File: src/scenes/ProgressReports/Detail/WorkflowEventList.tsx [hunk 34]
```diff
@@ -0,0 +1,71 @@
...
+  const statusEvents = [
+    {
+      ...events[0]!,
+      status: 'NotStarted' as keyof typeof StatusLabels,
+    },
+    ...events,
+  ];
+
+  return (
+    <Card sx={{ p: 4, overflowY: 'auto' }} elevation={0}>
```
CarsonF, line 30: Why deviate on `elevation` here?

CarsonF (changes requested):
I unresolved a few previous comments too that weren't actually resolved.

src/components/PeriodicReports/reports.ts
```diff
@@ -0,0 +1,14 @@
+import { PeriodicReportFragment } from './PeriodicReport.graphql';
+
+type Report = Pick<PeriodicReportFragment, 'start' | 'end'>;
+
+export const getReportLabel = (report?: Report) => {
```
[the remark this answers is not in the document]
andrewmurraydavid: My initial recommendation was to be in the `~/common/report.ts` file to be able to use it in multiple places where needed. But I strongly disagree that it should be in the `ReportLabel.tsx` file since we are using it for the title as well.

src/scenes/ProgressReports/Detail/WorkflowEventList.tsx
willdch: I was just following what the design shows.  The design does not have elevation.
I can change it I just wasn't sure what freedoms we can take with the design.  I did notice you did not include it in your example but I just thought you were not being specific about the elevation.

src/scenes/ProgressReports/Detail/WorkflowEventList.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <Stack
+        component="ul"
```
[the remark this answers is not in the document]
willdch: Again here, I was just following the design.  So do we want the numbers to show or do we want to still remove the numbers with 'list-Style: none"?

src/scenes/ProgressReports/Detail/WorkflowCard.tsx
willdch: Sorry about this, I have no Idea why this didn't get fixed the first time.  I could have sworn I fixed this and committed it.

src/scenes/ProgressReports/Detail/WorkflowCard.tsx
willdch: Not sure how this didn't get fixed the first time either. I did have to un-commit it a commit I made and I must have made a mistake in that process.

src/scenes/ProgressReports/Detail/WorkflowEventList.tsx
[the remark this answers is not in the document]
CarsonF: Yeah. This is an ordered list so we should use <ol>. The fact that we don't want to style it that way is a css concern. Changing the list style to none makes sense.

It could make sense to use MUI's List components. But don't feel like you need too. May be more work than worth it.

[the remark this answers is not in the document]
willdch: Got it.  I will take a look at the MUI list component and see how I feel about it.

src/scenes/ProgressReports/Detail/WorkflowEventList.tsx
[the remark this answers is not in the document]
willdch: I haven't taken a long look at this yet but I may need you to expand on this a little or possibly zoom or huddle about it at some point.  I can also meet with @andrewmurraydavid if he understands what you mean.

src/components/PeriodicReports/reports.ts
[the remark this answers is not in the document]
willdch: I don't have a strong opinion on this one.  I am glad to put it where you feel it makes most sense.
I misunderstood your previous comment about this one which is why I put it in the periodicReports folder.

CarsonF (approved):
LGTM!

I just squashed your commits down.
I also refactored the styling to use css grid because I think these makes the rows look nicer.

willdch:
This is awesome!  Thanks Carson!

PR #1440, opened 2023-05-05
finalize pipeline triggers

andrewmurraydavid:


bin/start-codepipeline/index.ts
```diff
@@ -0,0 +1,268 @@
...
+            if (err) {
+              console.error(err);
+              reject(err);
+            } else {
+              resolve(okData.pipelineExecution!);
+            }
+          }
+        );
+      }
+    );
```
CarsonF, line 145:
```ts
try {
  const result = await this.codepipeline.getPipelineExecution({
    pipelineName: this.pipelineName,
    pipelineExecutionId: executionId,
  }).promise();
  return result.pipelineExecution!;
} catch (e) {
  console.error(e); // although this could be moved up higher.
  throw e;
}
```

CarsonF (approved):
A good step forward 👍

PR #1442, opened 2023-05-09
Add ability for user to create unknown engagements

### LINKED ISSUE #1121: Anticipated Engagements Count
Often when a Project is in Early Conversations and Concept Development, we have an idea of how many Engagements might be a part of the Project, but don't know the exact Languages, People, or Partners who will be engaged.

Without this number, Leadership is blind to what may be coming in future years, making it difficult to forecast financial and personnel needs. We need a way to capture the number of anticipated engagements in order to provide these projections.

elisabeth-sorrell:
API PR: https://github.com/SeedCompany/cord-api-v3/pull/2801
Closes https://github.com/SeedCompany/cord-field/issues/1121


sethmcknight:

PR #1446, opened 2023-06-09
Initial refactor for sidedrawer

willdch:


sethmcknight:

PR #1461, opened 2023-07-29
feat: extend ms word formatting interceptor to take care of ordered l…

adam-soltech:
…ists and combined text

This should fix the issues related to both of these tickets:

https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/4888292883
https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/4877038624

These changes do the following:

- detect ms word ordered lists and unordered lists being pasted in. Potentially with other content
- If neither are detected nothing the normal paste is executed
- If either type of list is detected then all text is split by newline, paragraphs, ul, and ol are created as blocks and inserted in the right order
- This has been tested using the text dave thomas supplied (screenshot attached) as well as text with bullet points and ordered lists not from word, multiple list of the same and different types, and text with no lists

Screenshot 2023-07-28 at 10 42 47 PM

### DIFF
--- a/src/components/RichText/ms-word-helpers.ts
+++ b/src/components/RichText/ms-word-helpers.ts
@@ -1,31 +1,80 @@
 import { OutputData as RichTextData } from '@editorjs/editorjs';
 
-export const handleMsUnorderedList = (
+const hasWordListMarkers = (text: string) => {
+  return /•\t(.*)/.test(text) || /^\d+\.\s(.*)/m.test(text);
+};
+
+const isUnorderedList = (text: string) => {
+  return /•\t(.*)/.test(text);
+};
+
+const isOrderedList = (text: string) => {
+  return /^\d+\.\s(.*)/.test(text);
+};
+
+const createListBlock = (items: string[], style: 'unordered' | 'ordered') => {
+  return {
+    type: 'list',
+    data: {
+      style: style,
+      items: items,
+    },
+  };
+};
+
+const createParagraphBlock = (text: string) => {
+  return {
+    type: 'paragraph',
+    data: {
+      text: text.trim(),
+    },
+  };
+};
+
+export const handleMsPasteFormatting = (
   event: ClipboardEvent
 ): RichTextData | undefined => {
   const text = event.clipboardData?.getData('text');
 
-  // Match any line that starts with "•\t" followed by anything until the end of the line
-  const matches = text?.match(/•\t(.*)/g);
-
-  if (!matches) {
+  if (!text || !hasWordListMarkers(text)) {
     return undefined;
   }
-  // If there are matches, prevent default actions, then create an array of strings without the "•\t" prefix
-  const listItems = matches.map((item) => item.replace(/•\t/, ''));
 
-  // Now create a new object that conforms to the structure of RichTextData
-  // Assumption is that all the list items should be part of a single unordered list
+  const lines = text.split('\n');
+  const blocks = [];
+  let listItems: string[] = [];
+  let listType: 'unordered' | 'ordered' = 'unordered';
+  let paragraphText = '';
+
+  lines.forEach((line) => {
+    if (!line.trim() && paragraphText) {
+      blocks.push(createParagraphBlock(paragraphText));
+      paragraphText = '';
+    } else if (isUnorderedList(line)) {
+      listType = 'unordered';
+      listItems.push(line.replace(/•\t/, ''));
+    } else if (isOrderedList(line)) {
+      listType = 'ordered';
+      listItems.push(line.replace(/^\d+\.\s/, ''));
+    } else {
+      if (listItems.length) {
+        blocks.push(createListBlock(listItems, listType));
+        listItems = [];
+      }
+      paragraphText += line + ' ';
+    }
+  });
+
+  if (listItems.length) {
+    blocks.push(createListBlock(listItems, listType));
+  }
+
+  if (paragraphText.trim()) {
+    blocks.push(createParagraphBlock(paragraphText));
+  }
+
   return {
     time: Date.now(),
-    blocks: [
-      {
-        type: 'list',
-        data: {
-          style: 'unordered',
-          items: listItems,
-        },
-      },
-    ],
+    blocks: blocks,
   };
 };
--- a/src/components/RichText/RichTextField.tsx
+++ b/src/components/RichText/RichTextField.tsx
@@ -36,7 +36,7 @@ import { getHelperText, showError } from '../form/util';
 import { FormattedNumber } from '../Formatters';
 import { EditorJsTheme } from './EditorJsTheme';
 import type { ToolKey } from './editorJsTools';
-import { handleMsUnorderedList } from './ms-word-helpers';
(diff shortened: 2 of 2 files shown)


src/components/RichText/ms-word-helpers.ts
```diff
@@ -1,31 +1,80 @@
...
+    }
+  });
+
+  if (listItems.length) {
+    blocks.push(createListBlock(listItems, listType));
+  }
+
+  if (paragraphText.trim()) {
+    blocks.push(createParagraphBlock(paragraphText));
+  }
```
CarsonF, line 74: This is really hard to understand because of the mutability. You'll very rarely see me use `let` and `Array.push` because of this, and when I do it is contained to a small block of code.

It's hard to read because I can't just read top-down, but I have to follow the loop to understand the relationship between the subsequent iterations.

It looks like what is being done is:
1. Identify if the line is a paragraph or (un)ordered list item
2. Grouping the sibling list items together
3. Converting the paragraph blocks and list blocks into the JSON structures.

If that's the case, that's how I'd like to see it written in code. That will be much easier to follow.
Something like:
```ts
const parsedLines = text.split('\n').map((line) => {
  if (isUnorderedList(line) {
    return { type: 'ul', text: line.replace(...);
  }
  if (isOrderedList(line) {
    return { type: 'ol', text: line.replace(...);
  }
  return { type: 'p', text: line };
});

const groupedLines = ...; // I don't have this worked out of the top of my head but happy to chat about it.

const blocks = groupedLines.map((lines) => {
  const type = lines[0]!.type;
  const textLines = lines.map(line => line.text);
  if (type === 'p') {
    return createParagraphBlock(textLines);
  }
  if (type === 'ol') {
    return createListBlock(textLines);
  }
  if (type === 'ul') {
    return createListBlock(textLines);
  }
});
```

This is really off the cuff, so don't feel like you need to stick to anything.
The big point here is that all the structures are immutable. And so it's really easy to go line by line in the src code, and logically see what's happening.

Also I know the performance of yours is  ~`O(n)` and mine is ~`O(3n)`, but readability is much more important than a micro optimization like that. So I'm fine with the extra loops here.

Does that makes sense?

CarsonF (changes requested):

src/components/RichText/ms-word-helpers.ts
```diff
@@ -1,31 +1,80 @@
 import { OutputData as RichTextData } from '@editorjs/editorjs';
 
-export const handleMsUnorderedList = (
+const hasWordListMarkers = (text: string) => {
```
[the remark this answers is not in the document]
adam-soltech: yep done!

src/components/RichText/ms-word-helpers.ts
```diff
@@ -1,31 +1,80 @@
...
+  return /•\t(.*)/.test(text) || /^\d+\.\s(.*)/m.test(text);
```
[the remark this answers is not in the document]
adam-soltech: Yes, I redid this part anyway now

src/components/RichText/ms-word-helpers.ts
adam-soltech: I tried this but the issue is that the list items and paragraph pieces need to be cached somehow so that we know when a list is complete and when a paragraph is finished. I'm open to other ideas about how to do that and definitely hear you about clarity.

I've updated to use a clearly defined structure for the caching mechanisms with an interface at the beginning of the file, and where the caches are reset, I've commented so it is more clear. Happy to chat more about it as well

src/components/RichText/ms-word-helpers.ts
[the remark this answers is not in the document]
CarsonF: Looks the same to me?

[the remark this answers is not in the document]
adam-soltech: Yes, you're right. Sorry, I misunderstood. I should have looked at the file more closely. I've updated it now.

src/components/RichText/ms-word-helpers.ts
CarsonF: > I tried this but the issue is that the list items and paragraph pieces need to be cached somehow so that we know when a list is complete and when a paragraph is finished. I'm open to other ideas about how to do that and definitely hear you about clarity.

Right so that was my second step, the `groupedLines`.

While what you have now is technically immutable, it's still relying on the previous iterations and still using a combined state. Notably `BlockState` still has `paragraphText` and `listItems`, both of which could be empty or not.  Also comments like "and reset the paragraph text" point to this iteration-dependent logic.

---

Again, as I called out before these kinds of things could be necessary, but better code abstracts them away. This keeps the business logic separate from the mutations and/or iteration complexity.

I took the time now to write out this function.
```ts
const groupSiblingsBy = <T>(items: readonly T[], by: (item: T) => unknown) =>
  items.reduce((acc: T[][], cur: T) => {
    // If it's the first item or different from the last, start a new group
    if (!acc.length || by(acc.at(-1)![0]) !== by(cur)) {
      acc.push([cur]);
    } else {
      // Otherwise, add it to the current group
      acc[acc.length - 1].push(cur);
    }
    return acc;
  }, []);
```
So this does mutation and previous item comparison, but it's encapsulated in this function and all within a handful of lines.

So now we could parse each line, group with this function, and then take those grouped parsed lines and convert them to blocks.

```ts
const parsedLines = text.split('\n').map((line) => {
  if (isUnorderedList(line) {
    return { type: 'ul', text: line.replace(...);
  }
  if (isOrderedList(line) {
    return { type: 'ol', text: line.replace(...);
  }
  return { type: 'p', text: line };
});

const groupedLines = groupSiblingsBy(parsedLines, line => line.type);

const blocks = groupedLines.map((lines) => {
  const type = lines[0]!.type;
  const textLines = lines.map(line => line.text);
  if (type === 'p') {
    return createParagraphBlock(textLines);
  }
  if (type === 'ol') {
    return createListBlock(textLines);
  }
  if (type === 'ul') {
    return createListBlock(textLines);
  }
});
```

Idk about you but that's much easier for me to read. I can read top down and never have to question if something below affects the line I'm currently reading.

adam-soltech: Works for me. I had to add some additional logic to deal with paragraph grouping as well.

CarsonF: Ok but adding that into the `groupBy` function now couples the business logic with the sibling grouping logic. Which was one of my main points above to avoid / keep decoupled.

If empty lines should be treated differently, handle that in the parsing part.
```ts
const parsedLines = text.split('\n').map((line) => {
  ...
  if (line === '\r') {
    return { type: 'break' };
  }
});
```
Idk if you are trying to group up consecutive breaks or not, but if not then you could add another distinction:
```ts
const parsedLines = text.split('\n').map((line, index) => {
  ...
  if (line === '\r') {
    return { type: 'break', index };
  }
});

const groupedLines = groupSiblingsBy(parsedLines, line => {
  if (line.type === 'break') {
    return `break-${line.index}`;
  }
  return line.type;
});
```

CarsonF (approved):
Thanks @adam-soltech! Good work 🌮

PR #1462, opened 2023-08-03
Only show assignable roles in project member form

brentkulwicki:
Quick PR to fix what I consider to be a UX annoyance. This adds a little utility function to put the available options for roles on top of the list instead of having them all be in the same order all the time
before:
dropdown-before
after:
dropdown-after

Edit: created a Monday ticket for tracking - https://seed-company-squad.monday.com/boards/3451697530/pulses/4926194488


andrewmurraydavid (approved):
Thanks, @brentkulwicki ! LGTM!
I'd wait for @CarsonF's review just to be sure!

PR #1463, opened 2023-08-03
Created confirm dialog to check for empty variants by role

atGit2021:
Andrew, I started a draft PR now that I have some functionality up and running, but it uses some "dummy" buttons which are not tied into the workflow.  I'm also not sure of some of the other design changes that might be needed to incorporate/integrate this into the front end design standards.  Let me know if you have any bandwidth to review/huddle or if you want me to reach out to Brent or Adam. Thank you sir.

[Monday Ticket](https://seed-company-squad.monday.com/boards/3451697530/pulses/3832817727)


atGit2021:
I would like to remove the "any" type in the following function, `const getEmptyResponsesForRole = (list: any, roles: Set<Role>) => {
    if (list.items.length === 0) return [];
    const responsesForRole = list.items[0]?.responses.filter((item: any) => {
      return item.variant.responsibleRole
        ? roles.has(item.variant.responsibleRole)
        : false;
    });`  but the fragment type is not straight forward so I'm not sure how to create an interface or type to replace the "any" parameter type.  Any suggestions are appreciated.

bryanjnelson:
I think you may want this:

```ts
import { PromptResponseListFragment } from '../../../../../common/fragments/promptResponse.graphql';

…

const getEmptyResponsesForRole = (
    list: PromptResponseListFragment,
    roles: Set<Role>
  ) => {```

**Comment by atGit2021** [comment]
Hi Adam, I'm having a lint error issue that I can't seem to figure out.  It's showing up [here](https://github.com/SeedCompany/cord-field/pull/1463/commits/a54f3a1b45921c370ae1399a8dc3b2779df29be4#:~:text=%7BhandleDialogCancel%7D-,onContinue%3D%7BhandleDialogConfirm%7D,-/%3E)

**Comment by atGit2021** [comment]
@adam-soltech It's ready for your review now. Thank you.

File: src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx [hunk 1]
```diff
@@ -0,0 +1,142 @@
...
+            <Typography
+              component="h3"
+              variant="h4"
+              id="investor-connection"
+              fontSize="0.80rem"
+            >
+              INVESTOR CONNECTION
+            </Typography>
+            <ul aria-labelledby="investor-connection">
+              {listItems.map((item) => item)}
```
adam-soltech, line 107: This seems like it would be the same as just {listItems}

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+    communityStories,
+    roles
+  );
+  if (!hasSelectedStoryPrompt || emptyStoryResponsesByRole.length > 0) {
+    listItems.push(<li>Story</li>);
+  }
+
+  const investorConnectionHeader = !!(listItems.length > 0);
+  if (!investorConnectionHeader && !projectMgmtHeader && !open) {
+    updateConfirmStatus();
```
adam-soltech, line 80: Nit: consider renaming this to 'onUpdateConfirmStatus' (callbacks that act on an action are commonly prefixed with 'on' in react)

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+    roles: Set<Role>
+  ) => {
+    if (list.items.length === 0) return [];
+    const responsesForRole = list.items[0]?.responses.filter((item) => {
+      return item.variant.responsibleRole
+        ? roles.has(item.variant.responsibleRole)
+        : false;
+    });
+    const emptyResponsesByRole = responsesForRole?.filter((item) =>
+      !item.response.value && item.response.canEdit ? true : false
```
adam-soltech, line 57: This can be shortened to `!item.response.value && item.response.canEdit` since the comparison itself returns a boolean. In general, there is never a need to make a ternary that ends with `true : false`. If coercion to a boolean is needed the !! operator can be used ie: `!!0 //false !!1 //true`

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const handleCloseDialog = () => {
+    onCancel();
+  };
+
+  const getEmptyResponsesForRole = (
+    list: PromptResponseListFragment,
+    roles: Set<Role>
+  ) => {
+    if (list.items.length === 0) return [];
+    const responsesForRole = list.items[0]?.responses.filter((item) => {
```
adam-soltech, line 51: This list can be filtered once based on both filter criterion (ie: !item.response.value && item.response.canEdit and roles.has(item.variant.responsibleRole)

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+  onContinue,
+  onCancel,
+  updateConfirmStatus,
+}: ConfirmEmptyVariantsProps) => {
+  const { reportFile, teamNews, communityStories } = report;
+  const { session } = useSession();
+  const roles = new Set(session?.roles.value);
+  const hasReportProgressFile = !!reportFile.value;
+  const projectMgmtHeader = !hasReportProgressFile && reportFile.canEdit;
+  const hasSelectedTeamNewsPrompt = !!(teamNews.items.length > 0);
```
adam-soltech, line 34: The !! is not needed. The result of a numerical comparison will be a boolean without it

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const hasSelectedStoryPrompt = !!(communityStories.items.length > 0);
```
adam-soltech, line 35: Same thing here also

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const emptyStoryResponsesByRole = getEmptyResponsesForRole(
+    communityStories,
+    roles
+  );
+  if (!hasSelectedStoryPrompt || emptyStoryResponsesByRole.length > 0) {
+    listItems.push(<li>Story</li>);
+  }
+
+  const investorConnectionHeader = !!(listItems.length > 0);
```
adam-soltech, line 78: Same here. no need for the !!

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+    );
+    return emptyResponsesByRole ? emptyResponsesByRole : [];
+  };
+
+  const emptyTeamNewsResponsesByRole = getEmptyResponsesForRole(
+    teamNews,
+    roles
+  );
+  if (!hasSelectedTeamNewsPrompt || emptyTeamNewsResponsesByRole.length > 0) {
+    listItems.push(<li>Team News</li>);
```
adam-soltech, line 67: In react, each item in a list that is created from an array mapping needs a unique key. Example:

```
              listOrPlaceholders(projects?.items, 3).map((project, index) => (
                <ProjectListItemCard
                  key={project?.id ?? index}
                  project={project}
                  className={classes.listItem}
                />
              ))
            )}
```

adam-soltech (changes requested):
Hey Andre, thanks for asking for my review. Overall looks good. I added some comments when you have time.

atGit2021:
Updated per your feedback, thank you @adam-soltech

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -64,20 +63,20 @@ export const ConfirmEmptyVariants = ({
     roles
   );
   if (!hasSelectedTeamNewsPrompt || emptyTeamNewsResponsesByRole.length > 0) {
```
adam-soltech, line 65: IMO these double ifs and function calls could be simplified a bit IE:

  const listItemsData = [
    {
      key: "teamNews",
      condition: !report.teamNews.items.length || getEmptyResponsesForRole(report.teamNews, roles).length,
      label: "Team News"
    },
    {
      key: "story",
      condition: !report.communityStories.items.length || getEmptyResponsesForRole(report.communityStories, roles).length,
      label: "Story"
    }
  ];

  const listItems = listItemsData.filter(item => item.condition).map(item => (
    <li key={item.key}>{item.label}</li>
  ));

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+        : false;
+    });
+    return emptyResponsesByRole ? emptyResponsesByRole : [];
+  };
+
+  const emptyTeamNewsResponsesByRole = getEmptyResponsesForRole(
+    teamNews,
+    roles
+  );
+  if (!hasSelectedTeamNewsPrompt || emptyTeamNewsResponsesByRole.length > 0) {
```
adam-soltech, line 65: I think these double ifs and function calls could be simplified to something like this:

```
  const listItemsData = [
    {
      key: "teamNews",
      condition: !report.teamNews.items.length || getEmptyResponsesForRole(report.teamNews, roles).length,
      label: "Team News"
    },
    {
      key: "story",
      condition: !report.communityStories.items.length || getEmptyResponsesForRole(report.communityStories, roles).length,
      label: "Story"
    }
  ];

  const listItems = listItemsData.filter(item => item.condition).map(item => (
    <li key={item.key}>{item.label}</li>
  ));
```

Otherwise looks good to me. Thanks for the opportunity to review!

adam-soltech (commented):
One last nit - otherwise lgtm. Good job! Thanks!

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
atGit2021: Thanks for all the help @adam-soltech , updated per your feedback.

adam-soltech (approved):
LGTM - thanks for the opportunity to review

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/SubmitReportStep.tsx
CarsonF, line 1: I applaud your effort here in handling these async actions. The only thing really "wrong" is that async process isn't contained within the FinalForm submit handler. This means the Form thinks it is done submitting when it's not.
What this could look like to users, is they click confirm and then nothing happens (while the API submit is in progress over slow network), the transitions buttons are re-enabled though, so maybe they keep clicking the buttons, causing multiple network requests.
FWIW, my page keeps crashing too after I cancel/confirm on the dialog. I had to go into incognito to work around.

Here's a working solution
```tsx
const handleFormSubmit = async (values: TransitionFormValues) => {
  // This is an extra condition I added, it doesn't really make sense to ask
  // for confirmation if bypassing workflow or transitioning backwards (i.e. asking for changes)
  // This can be taken from the executeTransition.after function which does the same check for confetti
  const isApprovalTransition = true; // TODO
  // This is just an aggregate of the results of the isMissing step calls
  // pulled from the useProgressReportContext, as described in other comment
  const hasMissingFields = false; // TODO
  if (isApprovalTransition && hasMissingFields) {
    // This defer, is just a shortcut for a Promise and its resolve/reject callbacks
    const confirming = defer<boolean>();
    // Store the reference the resolve/"done" function on a ref, will talk about it below
    onConfirmRef.current = confirming.resolve;
    // Await that function to be called
    const confirmed = await confirming;
    // Communicate we are done waiting for this by clearing the function reference
    onConfirmRef.current = undefined;
    // If not confirmed, then return form errors to cancel the submission
    // An empty object is submission errors, just without displayable info
    if (!confirmed) {
      return {};
    }
  }
  // After confirming, if needed, then call and await API change
  await executeTransition(values);
};

// This is below here for reading, but should be higher in actual component than submit function
// This is a ref that can have a function which can be called with true/false.
const onConfirmRef = useRef<((confirm: boolean) => void) | undefined>();

// Unconditionally render this dialog, so it can animate in/out based on open boolean.
<ConfirmEmptyVariants
  report={report}
  // If we currently have a function here, that means we are waiting on the confirmation
  // so show the dialog
  open={!!onConfirmRef.current}
  // Here we create/pass the continue/cancel functions, which have the ref scoped in
  // When these are called, they call the latest/"current" promise resolve function
  // This allows that resolve function to change after these callback functions are created
  onContinue={() => onConfirmRef.current?.(true)}
  onCancel={() => onConfirmRef.current?.(false)}
/>
```
No state besides this ref is needed.

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const hasReportProgressFile = !!reportFile.value;
+  const projectMgmtHeader = !hasReportProgressFile && reportFile.canEdit;
+
+  const handleConfirmDialog = () => {
+    onContinue();
+  };
+
+  const handleCloseDialog = () => {
+    onCancel();
+  };
```
CarsonF, line 41: I don't like handle functions that just forward on to a prop. It's a layer of indirection that just adds complexity without doing anything. Just pass onContinue/Cancel through to the child components.

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+    },
+  ];
+  const listItems = listItemsData
+    .filter((item) => item.condition)
+    .map((item) => <li key={item.key}>{item.label}</li>);
+
+  const investorConnectionHeader = listItems.length > 0;
+  if (!investorConnectionHeader && !projectMgmtHeader && !open) {
+    onUpdateConfirmStatus();
+  }
```
CarsonF, line 81: Obviously all of this goes away based on the first comment. But if it hadn't this would be bad. Renders need to be pure, and calling this function on render is a side effect. For this specific case, there's nothing bad, but it's still a smell. The refactor I suggested solves this.

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+              PROJECT MANAGEMENT
+            </Typography>
+            <ul aria-labelledby="progress">
+              <li>Progress</li>
+            </ul>
+          </>
+        )}
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={handleCloseDialog} variant="text" color="error">
```
CarsonF, line 126: I know this was in the designs but I don't like this. Red buttons indicate a destructive action, like deleting a file. This cancel is not destructive.
```suggestion
        <Button onClick={handleCloseDialog} variant="text" color="secondary">
```

src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmVariantDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { useSession } from '~/components/Session';
+import { ProgressReportEditFragment } from '../../ProgressReportEdit.graphql';
+
+interface ConfirmEmptyVariantsProps {
+  report: ProgressReportEditFragment;
+  open: boolean;
+  onContinue: () => void;
+  onCancel: () => void;
+  onUpdateConfirmStatus: () => void;
+}
```
CarsonF, line 20: It's good practice to forward on the props of the root component you are wrapping.
```suggestion
interface ConfirmEmptyVariantsProps extends DialogProps {
  report: ProgressReportEditFragment;
  //  open: boolean; // already inherited
  onConfirm: () => void; // matches component name better
  // onCancel: () => void; // could probably just be inherited onClose
}
```

Thinking again, maybe just using a `DialogForm` would be better/easier. This handles the title, content, and button actions for you. Here's an example:
https://github.com/SeedCompany/cord-field/blob/master/src/components/posts/DeletePost/DeletePost.tsx#L27-L42

CarsonF (changes requested):
Sorry for the huge delay in reviewing this 😞 

We have a dynamic way to register steps already.
https://github.com/SeedCompany/cord-field/blob/master/src/scenes/ProgressReports/EditForm/Steps/index.ts#L9-L20
It would be better to re-use this to dynamically create your list here.
This could be iterated through just like [ProgressReportStepper](https://github.com/SeedCompany/cord-field/blob/master/src/scenes/ProgressReports/EditForm/ProgressReportStepper.tsx#L65-L106) does.

Each step component could be updated to declare an `isMissing` function.
https://github.com/SeedCompany/cord-field/blob/master/src/scenes/ProgressReports/EditForm/Steps/step.types.ts#L4-L6
```ts
isMissing?: (report: ReportProp & { currentUserRoles: ReadonlySet<Role> }) => boolean;
```
Perhaps this could be called right after confirming the step is enabled, [here](https://github.com/SeedCompany/cord-field/blob/master/src/scenes/ProgressReports/EditForm/ProgressReportContext.tsx#L33-L36)

CarsonF:
@atGit2021 I renamed "missing" to "incomplete" - I think this is a better representation. I also tweaked the dialog: renamed the component to make sense, renamed the filename to match, tweaked the styling to not have duplicate `<DialogContent>`, dropped the dividers to be consistent with our theme & MD, used proper semantic markup for the lists.

CarsonF (approved):

PR #1465, opened 2023-08-14
New icons for Sensitivity

CarsonF:


AmyRickwartz:
Icons look good. High five.

brentkulwicki (approved):
Not sure that it truly matter, but noticed that on the breadcrumbs, the tooltip effect works as expected
image
But on the main projects page, you get the tooltip from the sensitivity box instead
image

I am not calling this a blocker by any means, so ✅

brentkulwicki:
Wonder if we could update + combine the 2 in ProjectOverview
image

CarsonF:
Well the tooltips are trying to provide extra info not immediately visible.
When it's just the icon, we show the text for high/med/low. On the project overview page we show the text plainly there, so I don't see the need to duplicate it with a tooltip.
There is a tooltip there for _translation_ projects showing why the sensitivity is not editable.

So I'm still happy with current functionality.

PR #1478, opened 2023-09-21
Generate the list of sensitive operations dynamically from schema

CarsonF:
This allows the API to declare which queries/mutations are sensitive.
Then those are matched to our operations declared here.
A small win, from a deep rabbit hole lol.

API PR: https://github.com/SeedCompany/cord-api-v3/pull/2895


andrewmurraydavid (approved):
Nicely done, @CarsonF !!

PR #1479, opened 2023-09-21
Ignore current impersonation for sensitive operations

CarsonF:
Normal user flow doesn't allow this.
But there are edge cases where this could come up.


andrewmurraydavid (approved):
Big step for our security! Nice job, @CarsonF !

PR #1482, opened 2023-09-28
Order progress report workflow events desc

willdch:
https://seed-company-squad.monday.com/boards/3451697530/pulses/5243349687

### DIFF
--- a/src/scenes/ProgressReports/StatusHistory/WorkflowEventList.tsx
+++ b/src/scenes/ProgressReports/StatusHistory/WorkflowEventList.tsx
@@ -1,7 +1,10 @@
 import { ArrowForwardRounded as ArrowRightIcon } from '@mui/icons-material';
 import { Box, Card, Divider, Stack, Typography } from '@mui/material';
 import { Fragment } from 'react';
-import { ProgressReportStatusLabels as StatusLabels } from '~/api/schema/enumLists';
+import {
+  ProgressReportStatusLabels as StatusLabels,
+  ProgressReportStatusList as StatusList,
+} from '~/api/schema/enumLists';
 import { RelativeDateTime } from '~/components/Formatters';
 import { RichTextView } from '~/components/RichText';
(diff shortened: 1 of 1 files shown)


src/scenes/ProgressReports/StatusHistory/WorkflowEventList.tsx
```diff
@@ -19,6 +19,8 @@ export const WorkFlowEventList = ({
     return null;
   }
 
+  const reversedEvents = [...events].reverse();
```
CarsonF, line 22: Since we are depending the order to determine the previous status, I'd rather do that logic first. That way the reverse/forwards logic doesn't change the data anymore. We could also pull the first status from the API generated list.
```suggestion
  const list = events
    .map((event, i) => ({
      ...event,
      prevStatus: events[i - 1]?.status ?? StatusList[0]!,
    }))
    .reverse();
```

src/scenes/ProgressReports/StatusHistory/WorkflowEventList.tsx
```diff
@@ -37,10 +39,10 @@ export const WorkFlowEventList = ({
         containerType: 'inline-size',
       }}
     >
-      {events.map((event: WorkflowEventFragment, i: number) => (
+      {reversedEvents.map((event: WorkflowEventFragment, i: number) => (
```
CarsonF, line 42:
```suggestion
      {list.map((event) => (
```
The reverse could happen here too. Up to you.
```suggestion
      {list.reverse().map((event) => (
```

CarsonF (changes requested):
Just a small thing since we have to do the previous status calculation.

src/scenes/ProgressReports/StatusHistory/WorkflowEventList.tsx
willdch: where is this `StatusList[0]` you speak of?  Do you mean `ProgressReportStatusLabels`? it is not an array but dot notation works.

PR #1486, opened 2023-10-19
Handle deprecated variance explanation reasons

wes-cutting:
To be considered with
API PR: https://github.com/SeedCompany/cord-api-v3/pull/2914
[Ticket explaining changes](https://seed-company-squad.monday.com/boards/3451697530/pulses/5123136460)

### DIFF
--- a/src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.graphql
+++ b/src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.graphql
@@ -9,6 +9,7 @@ fragment ExplanationOfProgressStep on ProgressReport {
       ahead
       behind
       onTime
+      deprecated
     }
     reasons {
       ...SecuredStringList
--- a/src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.tsx
+++ b/src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.tsx
@@ -123,7 +123,7 @@ export const ExplanationOfProgress: StepComponent = ({ report }) => {
         autoSubmit
         keepDirtyOnReinitialize
       >
-        {({ handleSubmit, values: { group }, submitting }) => (
+        {({ handleSubmit, values: { group, reasons }, submitting }) => (
           <Card
             component="form"
             onSubmit={handleSubmit}
@@ -171,12 +171,26 @@ export const ExplanationOfProgress: StepComponent = ({ report }) => {
                 name="reasons"
                 label="Select a reason"
                 required
-                options={optionsByGroup[group].filter(
-                  (reason) => !optionsByGroup.deprecated.includes(reason)
-                )}
                 layout="column"
                 disabled={!explanation.reasons.canEdit}
-              />
+              >
+                {optionsByGroup[group].flatMap((reason) => {
+                  const isDeprecated =
+                    optionsByGroup.deprecated.includes(reason);
+                  if (isDeprecated && reason !== reasons) {
+                    // Don't even show deprecated options if they are not currently selected
+                    return [];
+                  }
+                  return (
+                    <EnumOption
+                      key={reason}
+                      label={reason}
+                      value={reason}
+                      disabled={isDeprecated}
+                    />
+                  );
+                })}
+              </EnumField>
             )}
 
             <SecuredField obj={explanation} name="comments">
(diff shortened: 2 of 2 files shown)


src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.tsx
```diff
@@ -169,7 +169,18 @@ export const ExplanationOfProgress: StepComponent = ({ report }) => {
               <EnumField
                 name="reasons"
                 label="Select a reason"
-                options={optionsByGroup[group]}
+                options={() =>
+                  optionsByGroup[group].map((reason: any) => (
```
CarsonF, line 173:
```suggestion
                  optionsByGroup[group].map((reason) => (
```
shouldn't need the `any` here, nor is it a good type ;)

src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.graphql
```diff
@@ -10,6 +10,9 @@ fragment ExplanationOfProgressStep on ProgressReport {
       behind
       onTime
     }
+    deprecated {
+      ...SecuredStringList
+    }
```
CarsonF, line 15: It's a list on the options object:
```suggestion
      onTime
      deprecated
    }
```

src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.tsx
```diff
@@ -169,7 +169,18 @@ export const ExplanationOfProgress: StepComponent = ({ report }) => {
...
+                    <EnumOption<OptionGroup>
+                      name={reason}
+                      variant='radio'
+                      value={reason}
+                      disabled={optionsByGroup[group].deprecated.includes(
```
CarsonF, line 178: I think this would be
```suggestion
                      disabled={optionsByGroup.deprecated.includes(
```

CarsonF (changes requested):

src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.tsx
```diff
@@ -170,8 +170,10 @@ export const ExplanationOfProgress: StepComponent = ({ report }) => {
               <EnumField
                 name="reasons"
                 label="Select a reason"
-                options={optionsByGroup[group]}
                 required
+                options={optionsByGroup[group].filter(
+                  (reason) => !optionsByGroup.deprecated.includes(reason)
+                )}
```
CarsonF, line 176: This is functionally the same as what we had in a previous iteration where the API just removed this value from the list. We want to show this value, just as disabled. Here's what I would do:
```ts
<EnumField
  name="reasons"
  label="Select a reason"
  required
  layout="column"
  disabled={!explanation.reasons.canEdit}
>
  {optionsByGroup[group].flatMap((reason) => {
    const isDeprecated =
      optionsByGroup.deprecated.includes(reason);
    if (isDeprecated && reason !== reasons) {
      // Don't even show deprecated options if they are not currently selected
      return [];
    }
    return (
      <EnumOption
        key={reason}
        label={reason}
        value={reason}
        disabled={isDeprecated}
      />
    );
  })}
</EnumField>
```

CarsonF (changes requested):

src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.tsx
wes-cutting: You've lost me on the if check `isDeprecated && reason !== reasons` where does reasons come from?

wes-cutting: Also, why would we return an empty array in that check when the map return of that flatMap returns an Element?

CarsonF: Oh my b. `reasons` is destructured from `values` on L126.
flatMap can return one or more elements which are flattened into a single list. So returning `[]` basically means skip this input item.

CarsonF (approved):
Thanks @wes-cutting!

PR #1495, opened 2023-11-20
fix: remove aws-sdk npm prefix

adam-soltech:
npm prefix on aws-sdk is breaking package installs. this removes the prefix


andrewmurraydavid (approved):
Thanks, @adam-soltech !

PR #1500, opened 2023-12-11
Changed period name Final to Additional Updates

atGit2021:


andrewmurraydavid (approved):
Looks good! Great job tracking this down!
Sheri and Seth approve this change too (for historical reference).

PR #1502, opened 2023-12-13
Icon Change Request for Partner Profile Status

atGit2021:
[Monday Task](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/5592448432)

### DIFF
--- a/src/components/Icons/InactiveStatusIcon.tsx
+++ b/src/components/Icons/InactiveStatusIcon.tsx
@@ -1,0 +1,9 @@
+import { SvgIcon, SvgIconProps } from '@mui/material';
+
+export const InactiveStatusIcon = (props: SvgIconProps) => (
+  <SvgIcon {...props} viewBox="0 0 21 21">
+    <path d="M19 13H5V11H19V13Z" />
+    <circle cx="18.8" cy="11.8" r="1.8" />
+    <circle cx="4.8" cy="11.8" r="1.8" />
+  </SvgIcon>
+);
--- a/src/scenes/Partners/Detail/PartnerDetail.tsx
+++ b/src/scenes/Partners/Detail/PartnerDetail.tsx
@@ -10,7 +10,7 @@ import {
   Tooltip,
   Typography,
 } from '@mui/material';
-import { Many } from '@seedcompany/common';
+import { Many, Nil } from '@seedcompany/common';
 import { Helmet } from 'react-helmet-async';
 import { useParams } from 'react-router-dom';
 import { PartialDeep } from 'type-fest';
@@ -19,6 +19,7 @@ import { useDialog } from '~/components/Dialog';
 import { Error } from '~/components/Error';
 import { FormattedDateTime } from '~/components/Formatters';
 import { IconButton } from '~/components/IconButton';
+import { InactiveStatusIcon } from '~/components/Icons/InactiveStatusIcon';
 import { TogglePinButton } from '~/components/TogglePinButton';
 import { EnumParam, makeQueryHandler, withDefault } from '~/hooks';
 import { EditablePartnerField, EditPartner } from '../Edit';
@@ -127,15 +128,26 @@ const PartnerHeader = ({ partner, editPartner }: PartnerViewEditProps) => {
   );
 };
 
+const StatusIcon = ({ isActive }: { isActive: boolean | Nil }) =>
+  isActive ? (
+    <TimelineIcon color="info" />
+  ) : (
+    <InactiveStatusIcon color="error" />
+  );
+
 const PartnerDataButtons = ({ partner, editPartner }: PartnerViewEditProps) => (
   <Box mt={3} mb={2} display="flex" gap={2}>
     <DataButton
       label="Status"
       onClick={() => editPartner('active')}
       secured={partner?.active}
-      startIcon={<TimelineIcon color="info" />}
+      startIcon={<StatusIcon isActive={partner?.active.value} />}
       redacted="You do not have permission to view Status"
-      children={partner?.active.value ? 'Active' : 'Inactive'}
+      children={
+        <Box color={partner?.active.value ? 'inherit' : 'text.secondary'}>
+          {partner?.active.value ? 'Active' : 'Inactive'}
+        </Box>
+      }
       loading={!partner}
     />
   </Box>


src/components/Icons/InactiveStatusIcon.tsx
```diff
@@ -0,0 +1,9 @@
+import { SvgIcon, SvgIconProps } from '@mui/material';
+
+export const InactiveStatusIcon = (props: SvgIconProps) => (
+  <SvgIcon {...props} viewBox="0 0 24 24">
+    <path d="M19 13H5V11H19V13Z" />
```
CarsonF, line 5: Is this line intentionally not centered vertically?

atGit2021: > Is this line intentionally not centered vertically?

Not sure about intentions, but this is what I received from Amy: [Figma file](https://www.figma.com/file/IfEmcJeuNY6ucgjYctlKB7/FOR-DEVELOPMENT---Cord-Field?type=design&node-id=11383-66102&mode=design&t=gV2rWeeeOH0FUwwk-0)

atGit2021: Amy responded saying it can be adjusted down so I'm going to look at the code to see how other icons are adjusted to follow same process.  Also she asked that I disable/gray out the text for the word "Inactive" so I need to make that change too.

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -133,9 +134,21 @@ const PartnerDataButtons = ({ partner, editPartner }: PartnerViewEditProps) => (
...
+      }
       redacted="You do not have permission to view Status"
-      children={partner?.active.value ? 'Active' : 'Inactive'}
+      children={
+        partner?.active.value ? (
+          'Active'
+        ) : (
+          <Box color="text.secondary">Inactive</Box>
+        )
+      }
```
andrewmurraydavid, line 151: I would do this to keep things a bit cleaner and easier to read:
```suggestion
      children={
        <Box color={partner?.active.value ? 'inherit' : 'text.secondary'}>
          {partner?.active.value ? 'Active' : 'Inactive'}
        </Box>
      }
```

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -133,9 +134,21 @@ const PartnerDataButtons = ({ partner, editPartner }: PartnerViewEditProps) => (
...
       onClick={() => editPartner('active')}
       secured={partner?.active}
-      startIcon={<TimelineIcon color="info" />}
+      startIcon={
+        partner?.active.value ? (
+          <TimelineIcon color="info" />
+        ) : (
+          <InactiveStatusIcon color="error" />
+        )
+      }
```
andrewmurraydavid, line 143: For this, I would create a separate component, called `StatusIcon` to avoid making this component big since the file already has multiple components and a bigger component in it.
This will also help if we want to reuse this kind of logic in different places.

Something like this:
```tsx
const StatusIcon = ({ isActive }: { isActive: Maybe<boolean> | null }) =>
  isActive ? (
    <TimelineIcon color="info" />
  ) : (
    <InactiveStatusIcon color="error" />
  );
  ```
  
  And then:
```suggestion
      startIcon={<StatusIcon isActive={partner?.active.value} />}
```

andrewmurraydavid (changes requested):
Really solid change! Great work, @atGit2021 !

This is borderline nitpick/future-proofing.

src/scenes/Partners/Detail/PartnerDetail.tsx
atGit2021: @andrewmurraydavid, where in the directory structure should the new StatusIcon component / .tsx file be located? I assume when you say it should be a separate component you are saying create a new file called StatusIcon.tsx where the StatusIcon function will be exported.

atGit2021: I added the function to the InactiveStatusIcon.tsx file so that future users can see how those can be used together.  Let me know if you want me to move it somewhere else.

andrewmurraydavid: For now we can keep it in the `src/scenes/Partners/Detail/PartnerDetail.tsx` file since it's the only place we're using it. When we end up chaning other places then a better place can be found for it.

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -127,15 +129,26 @@ const PartnerHeader = ({ partner, editPartner }: PartnerViewEditProps) => {
   );
 };
 
+const StatusIcon = ({ isActive }: { isActive: Maybe<boolean> | Nil }) =>
```
andrewmurraydavid, line 132: Why `Nil`? Asking because I haven't seen it before.

atGit2021: Nil covers null or undefined and I noticed that undefined was a possibility per the lint check, so I thought to use Nil to cover both.

andrewmurraydavid: The `Maybe` helper already adds `null` and `undefined`, just FYI

atGit2021: when i change it back to null, i get the lint error on the isActive property, so not sure how you are able to do it with just null

CarsonF: I'd prefer `boolean | Nil` to be consistent across our repos. `Maybe` is not ours and we don't really reference it anywhere.

andrewmurraydavid: For historical reference, I was importing the wrong `maybe` helper.

atGit2021: just seeing the above conversation; do you want me to remove the Maybe completely?

atGit2021: hopefully this last push works

andrewmurraydavid: @CarsonF up to you if you're ok with that or if we change it. `Maybe` comes form the graphql schema. It was used in another place too

CarsonF: LGTM

andrewmurraydavid (approved):
Great work, @atGit2021 ! 🚀

PR #1504, opened 2024-01-01
Partner Projects - server side pagination & sorting

adam-soltech:
[ticket for this](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/5675458057?filter=XQAAAALKAAAAAAAAAABBqQqHk62Wwa8yt9cbbXI48HpL2ei1XixEiNVOhiWmf4IrTn4lrcVxz0d8EzlqDpxL1dc7i9JjI1nNW0HEFgqaXhsDsHWTHjMEE6-qlUKWgWTaunFWUdBXUlo9GrL7-Dxs3yXtVsCpk7bU6tq4ASH03XfQsurYW6oz6SfWXIHjSrPNZIVE3__87GiYdcf5oUuHV6IpVDC18IDUOyTd3hSlyn4xwV6TGJGJ____cnOAAA)

- adds server side pagination to the partner projects table
- utilizes existing server pagination implementation as well as appolo and MUI


src/scenes/Partners/Edit/EditPartner.graphql
CarsonF, line 1: I don't think this change is needed.

src/scenes/Partners/Detail/PartnerDetail.graphql
```diff
@@ -62,12 +65,18 @@ fragment PartnerDetails on Partner {
...
   ...TogglePin
 }
+
+fragment PartnerProjects on SecuredProjectList {
+  canRead
+  canCreate
+  hasMore
+  total
+  items {
+    ...ProjectListItem
```
CarsonF, line 80: I would use your specific `PartnerDetailProjectsTableListItem` fragment here instead.

src/scenes/Partners/Detail/PartnerDetail.tsx
CarsonF, line 1: There's a lot going on here for projects, which isn't really the main concern of `PartnerDetail`. PartnerViewEditProps now have an optional `onPageChange`, which maps to the `projects`. It only makes sense for one tab.

The project tab should use its own GQL query. This saves allows us to skip loading project data when the tab isn't needed. It also allows us to _not_ re-fetch data that doesn't change, like primaryLocation, for each page request.

src/scenes/Partners/Detail/Tabs/Projects/PartnerDetailProjectsTable.tsx
```diff
@@ -1,93 +1,114 @@
...
-  projects: readonly Project[];
-  GridProps?: Omit<DataGridProps<Project>, 'columns' | 'isCellEditable'>;
+  projects: readonly PartnerDetailProjectsTableListItemFragment[];
+  onPageChange?: (page: number) => void;
+  totalRows?: number;
+  GridProps?: Omit<
+    DataGridProps<Project>,
+    'columns' | 'rows' | 'isCellEditable'
+  >;
 }
```
CarsonF, line 27: This component is really just a configured `DataGrid`, so I would expose the `DataGrid` props directly.

Having the prop names consistent reduces cognitive complexity too.
projects -> rows
totalRows -> rowCount

```ts
export interface PartnerDetailProjectsTableProps
  extends Omit<DataGridProps<Project>, 'columns'> {}
```

src/scenes/Partners/Detail/Tabs/Projects/PartnerDetailProjectsTable.tsx
```diff
@@ -1,93 +1,114 @@
...
+
   return (
-    <DataGrid<Project>
+    <DataGrid
+      rows={projects}
+      columns={columns}
+      pageSize={25}
+      rowCount={totalRows ?? projects.length}
+      pagination
+      paginationMode="server"
```
CarsonF, line 92: sorting also needs to be server side as well. Sorting a single page is kinda useless as it becomes a random subset of the unsorted entire list.

I don't believe the API supports sorting on `status` or _engagement total_ yet either. So that would need to be [added there](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/project/project.repository.ts#L284-L289) as well.

CarsonF (changes requested):

src/scenes/Partners/Detail/PartnerDetail.graphql
```diff
@@ -57,12 +60,18 @@ fragment PartnerDetails on Partner {
     canEdit
     value
   }
-  projects {
-    canRead
-    canCreate
-    items {
-      ...ProjectListItem
-    }
+  projects(input: $ProjectListInput) {
```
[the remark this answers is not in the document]
adam-soltech: I believe it is needed in the fragment for gql generation. In any case without matching the query with the var like this we get:

`Error 0: Fields "projects" conflict because they have differing arguments. Use different aliases on the fields to fetch both if this was intentional.`

src/scenes/Partners/Edit/EditPartner.graphql
adam-soltech: Without this update we get `Error 0: Variable "$projectListInput" is not defined by operation "UpdatePartner".` when generating our gql

src/scenes/Partners/Detail/Tabs/Projects/PartnerDetailProjectsTable.tsx
adam-soltech: I don't quite understand this one, unfortunately. Could you explain what you mean by "expose directly"

src/scenes/Partners/Detail/PartnerDetail.graphql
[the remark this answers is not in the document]
CarsonF: Ah I see, so the shared fragment is actually using forwarding the args to the operation, idk if that's new or not but that makes sense.
This would go away with the new specific `PartnerProjects` query

src/scenes/Partners/Detail/Tabs/Projects/PartnerDetailProjectsTable.tsx
adam-soltech: At the moment the internal team is not able to view all of the partner related projects because of the response limit. Should sorting be a separate ticket so that it does not block this fix?

CarsonF: IMO no. If we are looking for a short term fix, I'd just set the `count: 100` and try to work around the pagination entirely.

adam-soltech: Alright. If someone from product mentions that this is emergent (they have not yet) we can just go that route as a temp band aid

src/scenes/Partners/Detail/Tabs/Projects/PartnerDetailProjectsTable.tsx
CarsonF: I just mean expose the GridProps directly instead of under a GridProps key. Since this component is a decorated Grid, there's nothing really else it could be.

src/scenes/Partners/Detail/Tabs/Projects/PartnerDetailProjects.tsx
```diff
@@ -1,13 +1,66 @@
...
-  const projects = partner?.projects;
+  const { data, refetch } = useQuery(PartnerProjectsDocument, {
+    variables: {
+      id: partnerId,
+      projectListInput: {
+        count: 25,
+      },
+    },
+    notifyOnNetworkStatusChange: true,
+    fetchPolicy: 'no-cache',
```
adam-soltech, line 25: @CarsonF 

When adding sorting I got the following warning in the console:

Cache data may be lost when replacing the partner field of a Query object. This could cause additional (usually avoidable) network requests to fetch data that were otherwise cached. To address this problem (which is not a bug in Apollo Client), either ensure all objects of type Partner have an ID or a custom merge function, or define a custom merge function for the Query.partner field, so InMemoryCache can safely merge these objects

I am pretty sure this is because Apollo is trying to merge the new list of items, but because it is a sort some of the items are the same just in different positions so they are being seen as potential duplicates... I went in to typePolicies.base.ts and tried to add some logic that should handle sort like so:

```
  Partner: {
    fields: {
      projects: {
        keyArgs: [['projectListInput', ['page', 'count', 'sort', 'order']]],
        merge(
          existing: SecuredProjectList | undefined,
          incoming: SecuredProjectList,
          { readField }
        ) {
          // Pagination logic: If the id of the last item of existing is not equal to the id of the first item of incoming,
          // that means incoming data is a new page of data, so concatenate it.
          const existingLastItem = existing && last(existing.items);
          const incomingFirstItem = first(incoming.items);
          const existingLastItemId = existingLastItem && readField('id', existingLastItem);
          const incomingFirstItemId = incomingFirstItem && readField('id', incomingFirstItem);
          if (existingLastItemId !== incomingFirstItemId) {
            // Here, we spread the existing and incoming fields to maintain other properties of SecuredProjectList
            return {
              ...existing,
              ...incoming,
              items: [...(existing?.items || []), ...incoming.items],
            };
          }
          // Otherwise, the incoming data is an update operation, so replace the existing one.
          return incoming;
        }
      },
    },
  },
```

But that would not fix the issue. I do not think busting the cache here is ideal but I don't see any noticable performance impact (probably beacause of the pagination). Since you are more experienced with Apollo (and I noticed that you have added before to typePolicies) I'd love a better solution if you have one!

src/api/schema/typePolicies/typePolicies.base.ts
```diff
@@ -79,4 +79,27 @@ export const typePolicies: TypePolicies = {
       },
     },
   },
+  Partner: {
+    fields: {
+      projects: {
+        merge(existing, incoming: SecuredProjectList) {
```
adam-soltech, line 85: Adding this custom merge function to the type policies along with "    fetchPolicy: 'cache-first'," to the query is producing the desired behavior for server sort / pagination. That is:

1. New network calls are made for a sort / pagination combo that has not yet been fetched
2. Calls for previously fetched sort / pagination combos are taken from apollo cache with no network call
3.  The apollo caching warning I mentioned is gone

That said I would have preferred not to have to do this custom logic to achieve this result. The [apollo docs](https://www.apollographql.com/docs/react/v2/data/pagination/) on this matter say that the natural behavior of thier caching mechanism is as follows:

"Once the new data is returned from the server, the updateQuery function is used to merge it with the existing data, which will cause a re-render of your UI component with an expanded list."

That is why I was seeing items add to the re-render instead of overwriting the previous page. I see no documentation around how to do what we are trying to do (some tutorials are there but only for client pagination). I even tossed a generic version of the issue to [stackoverfow 
](https://stackoverflow.com/questions/77813536/apollo-server-pagination-list-only-items-recently-fetched-from-cache-or-network/77822005#77822005) and got crickets lol (I am going to post a bounty later)

@CarsonF If you have a better way I'd love to implement it... Otherwise this wors without error and doesn't seem to create a serious readability issue or any performance issues at all

CarsonF (approved):
Thanks @adam-soltech! :shipit:

PR #1506, opened 2024-01-04
Add Organization.acronym

atGit2021:
[Monday Task](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/5428604749)

### DIFF
--- a/src/scenes/Partners/Detail/PartnerDetail.tsx
+++ b/src/scenes/Partners/Detail/PartnerDetail.tsx
@@ -90,18 +90,15 @@ const PartnerHeader = ({
   partner,
   editPartner: edit,
 }: PartnerViewEditProps) => {
-  const partnerName = partner?.organization.value?.name.value;
-  const abrev =
-    partnerName && partnerName.length >= 30
-      ? partner.organization.value.avatarLetters
-      : null;
+  const name = partner?.organization.value?.name.value;
+  const acronym = partner?.organization.value?.acronym.value;
 
   return (
     <>
-      <Helmet title={partnerName ?? undefined} />
+      <Helmet title={acronym ?? name ?? undefined} />
       <Stack direction="row" gap={1}>
         <Typography variant="h2" lineHeight="inherit" mr={1}>
-          {abrev || partnerName}
+          {acronym ?? name}
         </Typography>
         <Tooltip title="Edit Partner">
           <IconButton
@@ -126,9 +123,9 @@ const PartnerHeader = ({
           }
         />
       </Stack>
-      {abrev && (
+      {acronym && (
         <Typography variant="h4" gutterBottom>
-          {partnerName}
+          {name}
         </Typography>
       )}
       {partner && (
--- a/src/scenes/Partners/Detail/PartnerDetail.tsx
+++ b/src/scenes/Partners/Detail/PartnerDetail.tsx
@@ -107,7 +107,11 @@ const PartnerHeader = ({
           <IconButton
             loading={!partner}
             onClick={() =>
-              edit(['organization.name', 'partner.globalInnovationsClient'])
+              edit([
+                'organization.name',
+                'organization.acronym',
+                'partner.globalInnovationsClient',
+              ])
             }
           >
             <Edit />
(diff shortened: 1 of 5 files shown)


src/scenes/Partners/Edit/EditOrganizationName.graphql
```diff
@@ -1,12 +1,19 @@
-mutation UpdateOrganizationName($id: ID!, $name: String!) {
-  updateOrganization(input: { organization: { id: $id, name: $name } }) {
+mutation UpdateOrganization($id: ID!, $name: String!, $acronym: String) {
+  updateOrganization(
+    input: { organization: { id: $id, name: $name, acronym: $acronym } }
+  ) {
```
CarsonF, line 4: Let's use the input object instead of enumerating the fields.
```suggestion
mutation UpdateOrganization($input: UpdateOrganization!) {
  updateOrganization(input: { organization: $input }) {
```
Will you rename the file to match as well?

src/scenes/Partners/Edit/EditOrganizationName.graphql
```diff
@@ -1,12 +1,19 @@
...
         canRead
         canEdit
         value
       }
+      acronym {
+        canRead
+        canEdit
+        value
+      }
       avatarLetters
```
CarsonF, line 17: I'd like to use a fragment for organization to dedupe the fields between this mutation and the query.

It could go below `PartnerDetails`

src/scenes/Partners/Edit/EditPartner.tsx
```diff
@@ -145,6 +148,7 @@ export const EditPartner = ({
         financialReportingTypes: partner.financialReportingTypes.value,
         address: partner.address.value,
         organizationName: partner.organization.value!.name.value!,
+        acronym: partner.organization.value?.acronym.value || ' ',
```
CarsonF, line 151: How come you are falling back to `" "` here?

src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -87,23 +87,24 @@ export const PartnerDetail = () => {
 
 const PartnerHeader = ({ partner, editPartner }: PartnerViewEditProps) => {
   const partnerName = partner?.organization.value?.name.value;
-  const abrev =
-    partnerName && partnerName.length >= 30
-      ? partner.organization.value.avatarLetters
-      : null;
```
CarsonF, line 93: Just want to double check that we are intentionally going away from the "long name is shown smaller with a large abbreviation"?

CarsonF (changes requested):

src/scenes/Partners/Detail/PartnerDetail.tsx
atGit2021: I'm not sure.  I thought this code that uses the avatar letters in place of the acronym was just a place holder until the acronym was implemented.

src/scenes/Partners/Edit/EditPartner.tsx
atGit2021: When I passed down a null or empty string (use case: when the acronym is cleared out on the UI) the back end API would treat it as if there were no changes to update and would not clear out/update the acronym value correctly.  I didn't want to mess with the back end API, so I used this string with an empty space as a work around.

src/scenes/Partners/Edit/EditOrganizationName.graphql
atGit2021: Will do, along with renaming the file

CarsonF:
Let me get #1518 in first, since it provides a better foundation to continue adding organization form fields. I'll rebase this on that.

CarsonF:
Ok I additionally have #1521 which fixes the weirdness with clearing text fields.

Separately I'm less convinced now of what the display UI should look like, so took a stab at something different. But I want to run it by product/design first.

atGit2021:
> Ok I additionally have #1521 which fixes the weirdness with clearing text fields.
> 
> Separately I'm less convinced now of what the display UI should look like, so took a stab at something different. But I want to run it by product/design first.

Just out of curiosity what is the business goal/reason for having avatarLetters and the acronym?  Is it basically giving the user the option to create a custom abbreviation when the code generated avatar letters are not exactly what is desired? Is there a business reason/use for having both fields that seem to do the same thing from a newbie's perspective?

CarsonF:
Avatar letters just grabs the first letters of the words in the name. So it's naive. I'm not sure of its usefulness now that we have an actual field. That's what I want to ask about.

PR #1513, opened 2024-01-23
Refactor/Simplify file preview code

CarsonF:
## Data Management
Previously we'd:
1. Fetch the temporary download url (removed with #1511)
2. Load the url into memory as a File/Blob.
3. Previewers would either (as a side effect):
	b) need a url instead, so they'd convert the blob back to an additional (object) url.
	a) do additional parsing and save result to additional state.
4. The outer wrapper (`FilePreview`) held the `loading` & `error` state, and passed getters/setters down to preview implementations.
     Every Previewer had 3 set calls for loading/error/data.

Now:
1. `FilePreview` just gives our API File fragment to the Previewer
2. Previewers either:
	a) render native elements with the url
	b) use the `useFilePreview` hook to load the file into memory and do additional parsing.
3. The hook uses suspense and error boundaries to handle error/loading states.
	Previewers only need to handle happy path.

## Unified file type list

There were two mime type lists. One held the icon and whether the type was previewable.
The other held what the previewer was for each type.

Now we just put the previewer with the type in one list.

The entries that were "other" with no previewer were just dropped, as this is the default fallback functionality.

I also converted the list to a map keyed by the mime type, since that's the only way we access it.

## Styling

Converted to new styling practices. New MUI components. No useStyles hooks.

## No page in context
The context was storing preview page. Moved this to local component state where applicable.
Not every previewer needed it. And each one that did had a side effect to reset it on mount.
Dumb. Local state doesn't have that problem.


brentkulwicki (approved):
Very solid! Much more easily digestible. 

Reviewed + briefly tested on this branch. ✅

andrewmurraydavid (approved):
Looks awesome! Great work, @CarsonF !
I love how clean `src/components/files/FilePreview/FilePreview.tsx` is now.
My only concern (not related to this refactor) is the HtmlPreview XSS potential. Though as we discussed privately, our users are not uploading html, so it's not a concern.
I'd say 🚀  it!

CarsonF:
Thank you guys. I just refactored a little bit more. Care to peek & review?

brentkulwicki (approved):

PR #1518, opened 2024-02-15
Refactor partner form

CarsonF:
- Organize fields, so they use real, namespaced names
- Secure fields, so they are disabled or hidden based on permission
- Combine mutation to simplify code and have only one HTTP request.


atGit2021 (approved):
LGTM!

PR #1521, opened 2024-02-20
Fix form field clearing with nulls

CarsonF:


atGit2021 (approved):
LGTM! @CarsonF , the part that got a little fuzzy for me was how the final form generics work in this section, but I see how you're overriding to null. 
`const { input, meta } = useFinalField<
    Value<T, Multiple>,
    El,
    Value<T, Multiple>
  >(name, {
    // Override default FF behavior to use null instead of undefined when given an empty string.
    // We want to send that explicit null to API instead of undefined which treats it as not given & unchanged.
    parse: (value) => (value === '' ? (null as Value<T, Multiple>) : value),`

CarsonF:
@atGit2021 yeah the typecast is just ignore that error. Not really a good way to express this.
This path isn't hit when the type isn't nullable.

PR #1525, opened 2024-04-03
Update budget record edit check only check amount field edit

CarsonF:
With https://github.com/SeedCompany/cord-api-v3/pull/3141 the amount edit flag is correct & sufficient for use here.


andrewmurraydavid (approved):
Nice!
Now `record.amount.canEdit` returns the correct edit check!

PR #1526, opened 2024-04-09
Adapt to TranslationProject Split & add type chips to UI

willdch:
API PR: https://github.com/SeedCompany/cord-api-v3/pull/3160

## Description

Added a chip displaying the project type on the Project Card, Project Overview Page and Partner table project tab. 

Project Card:
Screenshot 2024-04-08 at 3 44 50 PM

Project Overview:
Screenshot 2024-04-08 at 3 45 11 PM

Partner Table Project tab:
Screenshot 2024-04-09 at 11 19 29 AM

- [x] Change the task url above to the actual Monday task
- [ NA] Add/update tests if needed
- [x] Add reviewers to this PR

### DIFF
--- a/src/components/ProjectListItemCard/ProjectListItem.graphql
+++ b/src/components/ProjectListItemCard/ProjectListItem.graphql
@@ -23,14 +23,7 @@ fragment ProjectListItem on Project {
   }
   modifiedAt
   ...TogglePin
-  ... on InternshipProject {
-    engagements {
-      total
-    }
-  }
-  ... on TranslationProject {
-    engagements {
-      total
-    }
+  engagements {
+    total
   }
 }
--- a/src/components/form/Lookup/Project/ProjectLookup.graphql
+++ b/src/components/form/Lookup/Project/ProjectLookup.graphql
@@ -1,7 +1,5 @@
 query ProjectLookup($query: String!) {
-  search(
-    input: { query: $query, type: [TranslationProject, InternshipProject] }
-  ) {
+  search(input: { query: $query, type: [Project] }) {
     items {
       ... on Project {
         ...ProjectLookupItem
@@ -14,7 +12,7 @@ query TranslationProjectLookup($query: String!) {
   search(input: { query: $query, type: [TranslationProject] }) {
     items {
       ... on TranslationProject {
-        ...TranslationProjectLookupItem
+        ...ProjectLookupItem
       }
     }
   }
@@ -24,27 +22,13 @@ query InternshipProjectLookup($query: String!) {
   search(input: { query: $query, type: [InternshipProject] }) {
     items {
       ... on InternshipProject {
-        ...InternshipProjectLookupItem
+        ...ProjectLookupItem
       }
     }
   }
 }
 
 fragment ProjectLookupItem on Project {
-  id
-  name {
-    value
-  }
-}
-
-fragment TranslationProjectLookupItem on TranslationProject {
-  id
-  name {
-    value
-  }
-}
-
-fragment InternshipProjectLookupItem on InternshipProject {
   id
   name {
     value
--- a/src/scenes/Engagement/Delete/DeleteEngagement.tsx
+++ b/src/scenes/Engagement/Delete/DeleteEngagement.tsx
@@ -3,10 +3,7 @@ import { DeleteOutline } from '@mui/icons-material';
 import { Tooltip, Typography } from '@mui/material';
 import { removeItemFromList } from '~/api';
 import { callAll } from '~/common';
-import {
-  Id_InternshipProject_Fragment as InternshipProjectIdFragment,
-  Id_TranslationProject_Fragment as TranslationProjectIdFragment,
-} from '~/common/fragments';
+import { ProjectIdFragment } from '~/common/fragments';
 import { useDialog } from '../../../components/Dialog';
 import { DialogForm } from '../../../components/Dialog/DialogForm';
 import { SubmitError } from '../../../components/form';
@@ -17,10 +14,6 @@ import {
   DeleteEngagementDocument,
   EngagementToDeleteFragment,
 } from './DeleteEngagement.graphql';
-
-type ProjectIdFragment =
-  | TranslationProjectIdFragment
-  | InternshipProjectIdFragment;
 
 interface DeleteEngagementProps extends IconButtonProps {
   project: ProjectIdFragment;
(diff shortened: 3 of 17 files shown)


src/components/ProjectListItemCard/ProjectListItem.graphql
```diff
@@ -28,7 +28,12 @@ fragment ProjectListItem on Project {
       total
     }
   }
-  ... on TranslationProject {
+  ... on MomentumProject {
+    engagements {
+      total
+    }
+  }
+  ... on MultiplicationProject {
```
CarsonF, line 36: There's no need to intersect this on concretes. The Project interface has this field.
```gql
...TogglePin
engagements { total }
```

src/components/form/Lookup/Project/ProjectLookup.graphql
```diff
@@ -1,6 +1,9 @@
 query ProjectLookup($query: String!) {
   search(
-    input: { query: $query, type: [TranslationProject, InternshipProject] }
+    input: {
+      query: $query
+      type: [MomentumProject, MultiplicationProject, InternshipProject]
```
CarsonF, line 5: I think we can just use Project here.
```suggestion
      type: [Project]
```

src/components/form/Lookup/Project/ProjectLookup.graphql
```diff
@@ -11,10 +14,15 @@ query ProjectLookup($query: String!) {
 }
 
 query TranslationProjectLookup($query: String!) {
-  search(input: { query: $query, type: [TranslationProject] }) {
+  search(
+    input: { query: $query, type: [MomentumProject, MultiplicationProject] }
```
CarsonF, line 18: The rest of this file shouldn't need to change.

src/scenes/Engagement/Delete/DeleteEngagement.tsx
```diff
@@ -19,7 +20,8 @@ import {
 } from './DeleteEngagement.graphql';
 
 type ProjectIdFragment =
-  | TranslationProjectIdFragment
+  | MomentumProjectIdFragment
+  | MultiplicationProjectIdFragment
   | InternshipProjectIdFragment;
```
CarsonF, line 25: I'd move this type alias to a common spot. Probably in `common/fragments/` somewhere.

src/scenes/Engagement/LanguageEngagement/Create/CreateLanguageEngagement.graphql
```diff
@@ -6,7 +6,7 @@ mutation createLanguageEngagement($input: CreateLanguageEngagementInput!) {
   }
 }
 
-fragment TranslationProjectSensitivity on TranslationProject {
+fragment TranslationProjectSensitivity on MomentumProject {
```
CarsonF, line 9: Revert

src/scenes/Projects/ChangeRequest/Update/UpdateProjectChangeRequest.tsx
```diff
@@ -29,7 +29,10 @@ import {
 
 export interface UpdateProjectChangeRequestFormParams {
   project: {
-    __typename?: 'TranslationProject' | 'InternshipProject';
+    __typename?:
+      | 'MomentumProject'
+      | 'MultiplicationProject'
+      | 'InternshipProject';
     id: string;
   };
```
CarsonF, line 37:
```ts
project: ProjectIdFragement
```

src/scenes/Projects/List/ProjectList.tsx
```diff
@@ -56,7 +56,8 @@ export const ProjectList = () => {
       input: {
         ...sort.value,
         filter: {
-          ...omit(filters, 'tab'),
+          type: filters.type ? [filters.type] : undefined,
+          ...omit(filters, 'tab', 'type'),
```
CarsonF, line 60: What's this about?

CarsonF (changes requested):

src/scenes/Projects/List/ProjectList.tsx
andrewmurraydavid: You made the `type` an array on the API and that broke the filter as it's expecting an array. To keep the radio/single select functionality, we're sending the type value wrapped in a list

src/scenes/Engagement/LanguageEngagement/Create/CreateLanguageEngagement.graphql
andrewmurraydavid: needs concrete

src/scenes/Projects/List/ProjectList.tsx
CarsonF: Let's just change these two spots to match the new API instead
https://github.com/SeedCompany/cord-field/blob/22709b16472c1fc73043dcaacdf4915aa879584d/src/scenes/Projects/List/ProjectFilterOptions.tsx#L22-L22
https://github.com/SeedCompany/cord-field/blob/22709b16472c1fc73043dcaacdf4915aa879584d/src/scenes/Projects/List/ProjectFilterOptions.tsx#L47-L48

CarsonF:
My only remaining comment, is that the ProjectOverview uses the default Chip font size which is 13px, but everywhere else (in lists) its explicitly `body2` which is size 14px.

We should pick one size and use that as the default for all chips.

andrewmurraydavid:
> My only remaining comment, is that the ProjectOverview uses the default Chip font size which is 13px, but everywhere else (in lists) its explicitly `body2` which is size 14px.
> 
> We should pick one size and use that as the default for all chips.

@willdch is working on that

CarsonF:
@willdch @andrewmurraydavid I rejiggered the commits here to be single responsibility so we don't need to squash.
I also removed the `body2` size.

Want to give it a final look over before I merge?

CarsonF (approved):

andrewmurraydavid (changes requested):
There is a caching issue between Momentum and Multiplication. That's what was blocking us from addressing other things as that was a lot bigger priority than small UI tweaks.
Because Momentum and Multiplication are both Translation, Apollo doesn't make a call to fetch when switching between Momentum and Multiplication.

I will need to check if your changes fixed that.

CarsonF:
> Apollo doesn't make a call to fetch when switching between Momentum and Multiplication.

Wdym? how to reproduce?

andrewmurraydavid:
ok, not apollo, but the call to the API doesn't go through because it thinks it's cached.

I said apollo since it's responsible with caching as i recall.

To reproduce, go to the Project list, select Internship, apply and you will see the Internship projects. Then select multiplication and you will see Momentum, not multiplication.

andrewmurraydavid:
I just confirmed, @CarsonF , it's still happening

andrewmurraydavid (approved):
Thanks for fixing the filter issue, @CarsonF !

PR #1531, opened 2024-05-24
Extract data handing for tables to a generic hook

CarsonF:
https://seed-company-squad.monday.com/boards/5989610236/pulses/6654772311

Extracting this logic to be generic & type safe is harder, but then result is
```ts
const [props] = useTable({
  query: PartnerProjectsDocument,
  variables: { id: partnerId },
  listAt: 'partner.projects',
  initialInput: {
    sort: 'name',
    count: 20,
  },
});
```
which is really nice IMO.

I also fixed the bug where parsed temporal objects would be written back to cache and then double transformed. Both by avoiding writing them & by checking to not double transform (with a dev warning).


src/api/schema/typePolicies/scalars/scalars.parser.ts
```diff
@@ -5,11 +5,32 @@ import { Scalars } from '../../schema.graphql';
 export const Parsers: {
   [K in keyof Scalars]?: (val: any) => Scalars[K]['output'];
 } = {
-  Date: (val) => CalendarDate.fromISO(val),
-  DateTime: (val) => DateTime.fromISO(val),
+  Date: (val) => {
```
adam-soltech, line 8: +1

src/hooks/useTable.tsx
```diff
@@ -0,0 +1,196 @@
...
+  initialInput?: Partial<ListInput>;
+  keyArgs?: string[];
+}) => {
+  const [input, onChange] = useState(() => ({
+    ...defaultInitialInput,
+    ...initialInput,
+  }));
+
+  const queryForItemRef = useMemo(() => {
+    const queryForItemRef = structuredClone(query);
```
adam-soltech, line 72: Do we want to shadow like this?

adam-soltech (approved):
Good abstraction. Nit about a shadowed variable when shadowing function is later used as a function value

src/hooks/useTable.tsx
CarsonF: I don't have a problem with shadowing per se. Both of these variables represent the same thing, it's just the outer one is the result of the memoization. This is `useMemo` not `useCallback`, so the type of both is the same thing, never a function.

I still think this code is clear and straight forward. That doesn't mean other code that involves shadowing could be confusing, like if it shadows something completely different.

andrewmurraydavid (approved):
Nice work, @CarsonF !
This works locally, tested.

There will be breaking changes from v7 which will cause some types issues and props, but should be easy to fix.

CarsonF:
> There will be breaking changes from v7 which will cause some types issues and props, but should be easy to fix.

One spot now instead of N 😉

PR #1533, opened 2024-06-06
Partner Engagements tab max width

rdonigian:
https://seed-company-squad.monday.com/boards/5989610236/views/132288519/pulses/6786242569

### DIFF
--- a/src/scenes/Partners/Detail/PartnerDetail.tsx
+++ b/src/scenes/Partners/Detail/PartnerDetail.tsx
@@ -189,6 +189,14 @@ const usePartnerDetailsFilters = makeQueryHandler({
     'profile'
   ),
 });
+const tabStyles = {
+  profile: { maxWidth: 'lg' },
+  finance: { maxWidth: 'lg' },
+  people: { maxWidth: 'lg' },
+  projects: { maxWidth: 'lg' },
+  engagements: { maxWidth: 'auto' },
+  notes: { maxWidth: 'lg' },
+};
 const PartnerTabs = (props: PartnerViewEditProps) => {
   const [filters, setFilters] = usePartnerDetailsFilters();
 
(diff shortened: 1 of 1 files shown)


src/scenes/Partners/Detail/PartnerDetail.tsx
```diff
@@ -205,7 +205,7 @@ const PartnerTabs = (props: PartnerViewEditProps) => {
         <Tab label="Engagements" value="engagements" />
         <Tab label="Notes" value="notes" />
       </TabList>
-      <Paper sx={{ maxWidth: 'lg' }}>
+      <Paper sx={{ maxWidth: filters.tab === 'engagements' ? 'auto' : 'lg' }}>
```
CarsonF, line 208: I'd rather not do this. It breaks the open/close principle.
Can we do
```ts
<Paper sx={(theme) => ({
  '--maxWidth': theme.breakpoints.values.lg,
  maxWidth: 'var(--maxWidth)',
}}
```
And in the engagements tab
```ts
sx={{
  ':root': {
    '--maxWidth': 'initial',
  }
}}
```

CarsonF (changes requested):

CarsonF:
Does this work?

rdonigian:
It seemed like it but after a second glance apparently all the tabs are getting set to full width, so just working through on how to properly address it.

CarsonF:
It appears that this wouldn't work like I was thinking it would. I think we need to shift the `<Paper>` down to each tab. I originally moved it up here to keep it DRY, but not we can't customize.
I took the opportunity to fix this and a handful of other things on this page: #1535

rdonigian:
This is been solved by https://github.com/SeedCompany/cord-field/pull/1535

PR #1536, opened 2024-06-10
MUI DataGrid v5 -> v7

CarsonF:


rdonigian:
There are a few things that I noticed:  

1.  The partner-engagements table width is extending past the screen and causing a horizontal scroll of the page.
2. The Chip component seems a bit too large for the cell at it's current size.
3. There is an error in the console "Usage of the `rowCount` prop with client side pagination (`paginationMode="client"`) has no effect.`rowCount` is only meant to be used with `paginationMode="server"`."

CarsonF:
Thanks Rob. Let me check those out. I did notice the chips, wanted to keep and push that to product so they'd see how bad they look in general.

rdonigian (approved):
This looks good!

PR #1546, opened 2024-06-19
Fix pin button tooltip changing position when toggling

CarsonF:
### Before &nbsp;&nbsp;&nbsp;&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; After
Before
After

The button has padding on it, and the bottom corner was rotating down. This isn't visually noticed normally because the button is a circle. But it affected the position of the tooltip,
the bottom point pushed it down.

Changing instead to animating the svg icon within the button.

Previously, the element was mounted/unmounted with the icon switch. The animation was on the parent button, so it was fine. Now that the animation is on the svg icon, it needs to stay mounted for the animation to appear.
Fixed this by swapping out the paths within the svg element.

Also took the opportunity to swap off the old styles.

### DIFF
--- a/src/components/TogglePinButton/TogglePinButton.tsx
+++ b/src/components/TogglePinButton/TogglePinButton.tsx
@@ -1,25 +1,12 @@
 import { useMutation } from '@apollo/client';
-import { Tooltip, TooltipProps } from '@mui/material';
-import { makeStyles } from 'tss-react/mui';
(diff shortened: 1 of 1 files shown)


brentkulwicki (approved):
Pulled and tested locally. Code looks good. ✅

PR #1547, opened 2024-06-19
Fix CardGroup & ResponsiveDivider regression

CarsonF:
8de22b710107212c2d1569ee78a5a1f908e7a60c did not work like I thought it did.
Already fixed `PaperTooltip`. 
This addresses `CardGroup` & `ResponsiveDivider`.
### Fixed

----

Additionally I unified the two `ResponsiveDivider` components & changed implementation to only use one `<hr>` element.

---

Also fixed how the project details "Team Members" & "Partnerships" cards would smoosh weirdly.
### After

"See All" doesn't jump around now and looks good on all screen sizes.
The off center divider is accepted for this small width range.


brentkulwicki (approved):
Pulled and tested locally. Looks good

PR #1553, opened 2024-06-25
Theme Refresh

CarsonF:
Mostly trying to get back to MUI defaults in a lot of places.
Driven by filters in tables which I bring up half way down.

## Stop coupling color to typography
  MUI components will inherit typography variants, but use a different action color.
  For example, `<FormLabel>` defaults to `secondary` text color and `body1` typography.
  So we probably want to respect that color.

## Primary text color `#3c444e` -> 75% black
Using opacity for gray renders better than solid colors, esp on colored backgrounds.
For context, MUI defaults is 87%.
75% matches close to what we had before.

Before
Screenshot 2024-06-25 at 5 02 08 PM
After
Screenshot 2024-06-25 at 5 01 50 PM

## Secondary text color `#8f928b` -> 60% black (MUI default) 

Following suit of primary.
Opacity has to drop all the way to **45%** to be close to what we had before.
For context, the `disabled` text color is only slightly lower than that at 38%.

The contrast just wasn't enough at 45% esp on labels of `filled` inputs, which have a gray background.

Before
Screenshot 2024-06-25 at 5 05 13 PM
After
Screenshot 2024-06-25 at 5 04 44 PM

## ~~Re-allow input labels to act as placeholders that shrink~~ & remove all caps
~~Placeholders seem to still be desired by design, and MUI's default of the label acting as the placeholder and then shrinking to the label looks nice IMO. Better than having both a label & a placeholder.
I kept the slightly bolder weight when label is shrunk, but I would be down to remove that too.~~
I reverted this - can reconsider separately.

Before
Screenshot 2024-06-25 at 5 08 04 PM
Screenshot 2024-06-25 at 5 08 27 PM
After
Screenshot 2024-06-25 at 5 08 50 PM
Screenshot 2024-06-25 at 5 09 00 PM

### Header filters in Tables

Those screenshots seem like subtle changes.
All of the changes so far (other than primary color) are to make the table header filters look better.

Before
Screenshot 2024-06-25 at 5 15 43 PM
After
Screenshot 2024-06-25 at 5 16 10 PM

The less opaque & non caps help the text to be so in your face. Esp comparing to the column title, which is smaller.

## `IconButton` default size `large` -> `medium` (MUI/MD default)

48px -> 40px. Seems fine to me, and it sounds like in general we want to shrink our UI elements more.

Before
Screenshot 2024-06-25 at 5 22 45 PM
After
Screenshot 2024-06-25 at 5 23 08 PM

## ~~Background color `#fafafa` -> `#ffffff`~~ (MUI/MD default)

I reverted this per design request.

## `Card` elevation 8 -> 2 / `Paper` 1 -> 2 
1 is default for MUI.
To me, there is just not enough shadow on 1, to make the surface distinct from the background.
Paper is basically the same thing, so bumping to be consistent with Cards.

Before
Screenshot 2024-06-25 at 5 26 28 PM
After
Screenshot 2024-06-25 at 5 26 04 PM

- `AppBar` elevation 1 -> 2 
It is weird to have that floating surface lower than surfaces on page.

Before
Screenshot 2024-06-25 at 5 29 58 PM
After
Screenshot 2024-06-25 at 5 30 26 PM

https://seed-company-squad.monday.com/boards/3451697530/pulses/7550535438

### DIFF
--- a/src/theme/palette.ts
+++ b/src/theme/palette.ts
@@ -32,7 +32,8 @@ export const createPalette = ({ dark }: { dark?: boolean }) => {
       main: '#f2994a',
     },
     text: {
-      primary: dark ? '#f3f4f6' : '#3c444e',
+      // Close to #3c444e while still using alpha
+      ...(!dark ? { primary: 'rgba(0, 0, 0, 0.75)' } : {}),
       secondary: '#8f928b',
     },
 
--- a/src/theme/palette.ts
+++ b/src/theme/palette.ts
@@ -34,7 +34,9 @@ export const createPalette = ({ dark }: { dark?: boolean }) => {
     text: {
       // Close to #3c444e while still using alpha
       ...(!dark ? { primary: 'rgba(0, 0, 0, 0.75)' } : {}),
-      secondary: '#8f928b',
+      // Close to #8f928b while still using alpha
+      // Still it looks so contrast-less for form labels
+      // secondary: 'rgba(0, 0, 0, 0.45)',
     },
 
     // TODO theme.palette.augmentColor()
--- a/src/theme/palette.ts
+++ b/src/theme/palette.ts
@@ -1,10 +1,10 @@
 import { PaletteColor, PaletteColorOptions } from '@mui/material';
 import { grey } from '@mui/material/colors';
-import { lighten, PaletteOptions } from '@mui/material/styles';
+import { PaletteOptions } from '@mui/material/styles';
 import { Role } from '~/api/schema/schema.graphql';
 
 export const createPalette = ({ dark }: { dark?: boolean }) => {
-  const mainGreen = '#467f3b';
+  const mainGreen = '#1EA973';
   const roleLuminance = dark ? 32 : 84;
   const palette: PaletteOptions = {
     mode: dark ? 'dark' : 'light',
@@ -14,8 +14,6 @@ export const createPalette = ({ dark }: { dark?: boolean }) => {
     primary: {
       main: mainGreen,
       contrastText: '#ffffff',
-      // default is lighten main color by 0.2
-      light: lighten(mainGreen, 0.4),
     },
     secondary: {
       main: dark ? grey[50] : '#3c444e',
--- a/src/components/Changeset/PropertyDiff.tsx
+++ b/src/components/Changeset/PropertyDiff.tsx
@@ -19,7 +19,7 @@ const useStyles = makeStyles()(({ palette, shape, spacing }) => ({
   },
   current: {
     color: palette.primary.dark,
-    background: fade(palette.primary.light, 0.5),
+    background: fade(palette.primary.main, 0.5),
   },
 }));
 
--- a/src/components/Widgets/Widget.tsx
+++ b/src/components/Widgets/Widget.tsx
@@ -10,8 +10,6 @@ export interface WidgetProps extends ChildrenProp {
 export const Widget = ({ colSpan, rowSpan, sx, ...props }: WidgetProps) => (
   <Card
     component={Stack}
-    elevation={0}
-    variant="outlined"
     sx={[
       {
         gridColumn: `span ${colSpan}`,
--- a/src/scenes/ProgressReports/EditForm/ProgressReportStepper.tsx
+++ b/src/scenes/ProgressReports/EditForm/ProgressReportStepper.tsx
@@ -67,12 +67,7 @@ export const ProgressReportStepper = () => {
     useProgressReportContext();
 
   return (
-    <Paper
-      elevation={4}
-      sx={{ p: 2 }}
-      component="nav"
-      aria-label="Quarterly Report Steps"
-    >
+    <Paper sx={{ p: 2 }} component="nav" aria-label="Quarterly Report Steps">
       <Typography component="h3" paragraph aria-hidden>
         Steps:
       </Typography>
--- a/src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmIncompleteSubmissionDialog.tsx
+++ b/src/scenes/ProgressReports/EditForm/Steps/SubmitReportStep/ConfirmIncompleteSubmissionDialog.tsx
@@ -66,7 +66,6 @@ const IncompleteSteps = () => {
             {steps.map(({ label: stepName, severity }) => (
               <Typography component="li" key={stepName}>
                 <Link
-                  underline="always"
                   // Carson doesn't love this logic being duplicated here,
                   // but values semantic markup more
                   to={{ search: `?step=${kebabCase(stepName)}` }}
--- a/src/scenes/ProgressReports/EditForm/Steps/VariantAccordion.tsx
+++ b/src/scenes/ProgressReports/EditForm/Steps/VariantAccordion.tsx
@@ -18,7 +18,7 @@ export const VariantAccordion = ({
   const [expanded, { toggle }] = useToggle(expandedInput ?? false);
 
   return (
-    <Accordion expanded={expanded} elevation={2} square>
+    <Accordion expanded={expanded} square>
       <AccordionSummary
         aria-controls={`${variant.key}-content`}
         expandIcon={<ExpandMore />}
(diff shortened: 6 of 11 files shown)


AmyRickwartz:
Great! A lot of these things I have been wanting to change as well.

> ## Stop coupling color to typography
> MUI components will inherit typography variants, but use a different action color. For example, `<FormLabel>` defaults to `secondary` text color and `body1` typography. So we probably want to respect that color.
>
> 
This one seems ok to change to in general, but I'm not sure how some of the things we did in the past will look once we change back to the theme, Is there a way to preview. I'd also like the ability to override if I needed to.
> 
> 
> ## Primary text color `#3c444e` -> 75% black
> Using opacity for gray renders better than solid colors, esp on colored backgrounds. For context, MUI defaults is 87%. 75% matches close to what we had before.
> 
> 
I studied this quite a bit when I first started on CF on the Material site. Can we change the primary and secondary color to be just black and the opacities are what make it primary, secondary, ect? Before I think we were using different values of black on the spectrum to render this affect and that does make it harder to read and be flexible to change the overall colors in the future....
> 
> 
> ## Secondary text color `#8f928b` -> 60% black (MUI default)
> Following suit of primary. Opacity has to drop all the way to **45%** to be close to what we had before. For context, the `disabled` text color is only slightly lower than that at 38%.
> 
> 
Same thing here.
> 
> ## Re-allow input labels to act as placeholders that shrink & remove all caps
> Placeholders seem to still be desired by design, and MUI's default of the label acting as the placeholder and then shrinking to the label looks nice IMO. Better than having both a label & a placeholder. I kept the slightly bolder weight when label is shrunk, but I would be down to remove that too.
> 
> 
This is another one that I have done some reading on as well. So I agree the labels that are big then move and get small are cleaner, however they work better on simple forms. We have many places in CF that have more complex forms and keeping the label stationary is better for scanning and some other things. I agree that we don't need place holder text for everything because it can be redundant and looks cluttered. I have often wondered if the text field needs helper text, then that's exactly where we place it, in the helper text line at a secondary (lighter color). But then if there is an error, will it be replaced by the error text, which shows up in the helper text area? We would need to work that out.

The all caps has not been my favorite for I would agree with making it Title Case (maybe sentence case, I'd have to see what that looked like).
> 
> 
> ### Header filters in Tables
> Those screenshots seem like subtle changes. All of the changes so far (other than primary color) are to make the table header filters look better.
> 
> 


> 
> The less opaque & non caps help the text to be so in your face. Esp comparing to the column title, which is smaller.
> 
> ## `IconButton` default size `large` -> `medium` (MUI/MD default)
> 48px -> 40px. Seems fine to me, and it sounds like in general we want to shrink our UI elements more.
>
> 
Agreed. The medium buttons are good.
> 
> 
> 
> ## Background color `#fafafa` -> `#ffffff` (MUI/MD default)
> Pics below. Idk I could revert this. Just trying to follow defaults for this iteration.
>
> 
This one is a little trickier because a lot of the designs have started with a slightly tinted screen. Also, the bright white is harder on the eyes. So lets stay with the `#fafafa` for now.
> 
> 
> ## `Card` elevation 8 -> 2 / `Paper` 1 -> 2
> 1 is default for MUI. To me, there is just not enough shadow on 1, to make the surface distinct from the background. Paper is basically the same thing, so bumping to be consistent with Cards.
> 
> 
Yes agreed. We have talked about this one. Moving to 2 for default is great with me.
> 
> 
> * `AppBar` elevation 1 -> 2
>   It is weird to have that floating surface lower than surfaces on page.
> 
Yes, agree on this one as well.

CarsonF:
> This one seems ok to change to in general, but I'm not sure how some of the things we did in the past will look once we change back to the theme, Is there a way to preview. I'd also like the ability to override if I needed to.

We can preview in dev for a bit. We can always make a different decision in the future.

> Can we change the primary and secondary color to be just black and the opacities are what make it primary, secondary, ect? Before I think we were using different values of black on the spectrum to render this affect and that does make it harder to read and be flexible to change the overall colors in the future....

That's exactly what I changed here. 

> This is another one that I have done some reading on as well. So I agree the labels that are big then move and get small are cleaner, however they work better on simple forms. We have many places in CF that have more complex forms and keeping the label stationary is better for scanning and some other things. I agree that we don't need place holder text for everything because it can be redundant and looks cluttered. I have often wondered if the text field needs helper text, then that's exactly where we place it, in the helper text line at a secondary (lighter color). But then if there is an error, will it be replaced by the error text, which shows up in the helper text area? We would need to work that out.

Not sure I agree on being better for "simple" forms. I think in general the thought is we don't like complex forms, separate from the label position.
As far as the different components:
Labels are short identifiers.
Placeholders should help give an example value when values are not immediately obvious.
Helper text gives extra info not yet provided. I rarely see the need for these. Here is only of the only places I can think of that we use them:
Screenshot 2024-07-02 at 1 50 39 PM

Maybe we can try out this change and the title case change for a couple days?


> This one is a little trickier because a lot of the designs have started with a slightly tinted screen. Also, the bright white is harder on the eyes. So lets stay with the `#fafafa` for now.

Roger, I'll revert.

AmyRickwartz:
> > This is another one that I have done some reading on as well. So I agree the labels that are big then move and get small are cleaner, however they work better on simple forms. We have many places in CF that have more complex forms and keeping the label stationary is better for scanning and some other things. I agree that we don't need place holder text for everything because it can be redundant and looks cluttered. I have often wondered if the text field needs helper text, then that's exactly where we place it, in the helper text line at a secondary (lighter color). But then if there is an error, will it be replaced by the error text, which shows up in the helper text area? We would need to work that out.
> 
> Not sure I agree on being better for "simple" forms. I think in general the thought is we don't like complex forms, separate from the label position. As far as the different components: Labels are short identifiers. Placeholders should help give an example value when values are not immediately obvious. Helper text gives extra info not yet provided. I rarely see the need for these. Here is only of the only places I can think of that we use them: Screenshot 2024-07-02 at 1 50 39 PM
> 
> Maybe we can try out this change and the title case change for a couple days?
> 
I'm remembering now that the issue has been for fields that later need to be edited. There were so many disscusions about editing in place vs going to another page to edit. But I think that is what got weird. The labels were jumping around when we then went to edit the information.

CarsonF:
> I'm remembering now that the issue has been for fields that later need to be edited. There were so many disscusions about editing in place vs going to another page to edit. But I think that is what got weird. The labels were jumping around when we then went to edit the information. It's even made me think about moving the label outside the box. Vonshel has made some new designs like this. So it might be something we move to in the future.
> 
> For sure undo the caps and make the helper text less bold. We can start there.

I don't think this has anything to do with editing in place vs dialog. This also only affects the empty state. I'm not really sure what you mean by "jumping around when we then went to the edit in the information."

Sure, we can undo the caps, and circle back to re-enabling the floating label/placeholder behavior.

src/theme/palette.ts
```diff
@@ -14,6 +14,7 @@ export const createPalette = ({ dark }: { dark?: boolean }) => {
     primary: {
       main: mainGreen,
       contrastText: '#ffffff',
+      // TODO could not be needed anymore with change to primary green
       // default is lighten main color by 0.2
       light: lighten(mainGreen, 0.4),
```
CarsonF, line 19: Look for usages of `primary.light` and consider a) replacing this with default lighten by 20% and/or b) replacing usage with `primary.main`

CarsonF (commented):
@rdonigian @atGit2021 Would you guys play around with this and see if you can spot anything that looks weird?

src/theme/palette.ts
atGit2021: @CarsonF  Given the new color is already a "brighter" green, it didn't really "stand out" much more when utilizing the `primary.light: lighten(mainGreen, 0.2)`, so I opted to refactor for the primary.main.  If that's ok, I'll push a commit with the changes, let me know.

CarsonF: @atGit2021 Does it stand out more with the 40% as it is currently?
I'm just looking for what looks the best. Commit away.

atGit2021: It didn't to me, but then again i'm not known for design :)

PR #1551, opened 2024-06-25
Projects data grid

rdonigian:
https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/6654729716

PR #1559, opened 2024-06-28
Abstract recent partner data grid work

CarsonF:
Abstracts partner tabs styles, and project & engagement columns to components/ for reuse.
Also synced the partner project's grid to have all the same settings as engagements.

Parts of this were pulled from #1551 and polished.


CarsonF:
Good catch, just renamed!

PR #1567, opened 2024-07-10
Comments

andrewmurraydavid:
Added functionality to create, read, edit and reply to comments.

https://seed-company-squad.monday.com/boards/3451697530/pulses/6959737996

https://seed-company-squad.monday.com/boards/5989610236/pulses/6959748328


src/components/Comments/CommentsBar.tsx
rdonigian, line 1: This works, I do think that later we can abstract this out as the sidebar becomes an option for other features.

src/components/Comments/CommentReply/CommentReply.tsx
```diff
@@ -0,0 +1,84 @@
...
+  sx,
+}: CommentReplyProps) => {
+  const [createOrReplyComment] = useMutation(CreateOrReplyCommentDocument, {
+    update: addItemToList({
+      listId: 'commentThreads',
+      outputToItem: (data) => data.createComment.commentThread,
+    }),
+  });
+
+  const [initialValues] = useState(() => ({
```
rdonigian, line 36: Do we need the useState here?

src/components/Comments/CommentThread/CommentThreadAccordion.tsx
```diff
@@ -0,0 +1,56 @@
...
+          margin: 0,
+        },
+        padding: 0,
+      }}
+    >
+      {children}
+      <AccordionDetails
+        sx={{
+          padding: 0,
+          margin: 0,
```
CarsonF, line 38: FWIW there is a shortcut here, which I think you would like since you like tailwind.
```tsx
sx={{ p: 0, m: 0 }}
```
No need to change just FYI

src/components/Comments/CommentsBar.tsx
```diff
@@ -0,0 +1,45 @@
...
+      sx={[
+        !open && { display: 'none' },
+        { overflowY: 'auto', display: 'flex' },
+        open && { width: width, flexShrink: 0 },
+      ]}
+    >
+      <Stack alignItems="start" spacing={1}>
+        <IconButton size="small" onClick={() => toggleCommentsBar(false)}>
+          <CloseIcon />
+        </IconButton>
```
CarsonF, line 34: There's no padding here between the drawer and the icon.

It's also kinda a shame it takes up its own row, though I understand the dilemma.
Maybe we could consider a "Hide" button to the left of the submit "Comment" button on the next pass?

src/components/Comments/CommentsBar.tsx
```diff
@@ -0,0 +1,45 @@
...
+  resourceId?: string;
+}
+
+export const CommentsBar = ({ resourceId }: CommentsBarProps) => {
+  const { isCommentsBarOpen, toggleCommentsBar } = useCommentsContext();
+  const open = isCommentsBarOpen && !!resourceId;
+  const width = 350;
+
+  return (
+    <Drawer
```
CarsonF, line 17: Currently this drawer can cover main layout content. It would be better to treat this is something in the layout so the main content is pushed to the left. Happy to brainstorm on ways we could do this when you get here.

src/components/Comments/CommentsBar.tsx
```diff
@@ -0,0 +1,45 @@
...
+    >
+      <Stack alignItems="start" spacing={1}>
+        <IconButton size="small" onClick={() => toggleCommentsBar(false)}>
+          <CloseIcon />
+        </IconButton>
+        {resourceId ? (
+          <CommentsThreadList resourceId={resourceId} />
+        ) : (
+          <Box height={1} width={1} pt={1} px={2}>
+            <Typography variant="h6">Comments not available here</Typography>
```
CarsonF, line 39: Couldn't this be:
```suggestion
          <Typography variant="h6" pt={1} px={2}>Comments not available here</Typography>
```

src/components/Comments/CommentsBarContext.tsx
```diff
@@ -0,0 +1,93 @@
...
+CommentsBarContext.displayName = 'CommentsBarContext';
+
+export const CommentsBarProvider = ({ children }: ChildrenProp) => {
+  const [cookieCommentsBarOpen, setCookieCommentsBarOpen] = useCookieState(
+    'show-comments',
+    { defaultValue: 'false' }
+  );
+  const [isCommentsBarOpen, setCommentsBarOpen] = useState(
+    cookieCommentsBarOpen === 'true'
+  );
```
CarsonF, line 32: I think this just duplicates state from the cookie hook.
```suggestion
  const isCommentsBarOpen = cookieCommentsBarOpen === 'true';
```

src/components/Comments/CommentsBarContext.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [expandedThreads, setExpandedThreads] = useState<string[]>([]);
```
CarsonF, line 33: I think a set would be more useful here. ahooks' as one that could be adapted.
```ts
const [currentExpandedThreads, setExpandedThreads] = useSet<string>();
const expandedThreads = useMemo(() => ({
  ...setExpandedThreads,
  has: currentExpandedThreads.has.bind(currentExpandedThreads),
  toggle: (threadId: string, next?: boolean) => {
    next = next ?? !currentExpandedThreads.has(threadId);
    setExpandedThreads[next ? 'add' : 'remove'](threadId);
  },
}), [expandedThreadsHook];
```
Then
```ts
const { expandedThreads } = useContext();

expandedThreads.has(threadId);
expandedThreads.toggle(threadId);
expandedThreads.add(threadId);
expandedThreads.remove(threadId);
```

src/components/Comments/CommentsThreadList.graphql
```diff
@@ -0,0 +1,40 @@
...
+  commentThreads(resource: $resourceId, input: $input) {
+    total
+    hasMore
+    items {
+      ...CommentThreadProps
+    }
+  }
+}
+
+fragment CommentProps on Comment {
```
CarsonF, line 11: "Props" suffix here isn't really idiomatic. This seems to be mainly used in the `CommentItem` component - can we rename to `commentItem` and move to `CommentItem/commentItem.graphql`?

src/components/Comments/CommentsThreadList.graphql
```diff
@@ -0,0 +1,40 @@
...
+    id
+    fullName
+    avatarLetters
+  }
+  createdAt
+  modifiedAt
+  canDelete
+}
+
+fragment CommentThreadProps on CommentThread {
```
CarsonF, line 28: Same here. `CommentThread/commentThread.graphql` -> `fragment commentThread`

CarsonF (approved):
I think I'm good with this now! 🎉

PR #1568, opened 2024-07-18
Replace ROD header label with ROLV

rdonigian:
https://seed-company-squad.monday.com/boards/5989610236/pulses/7052298793


CarsonF:
Can we begin work to rename on API too?

CarsonF (approved):

PR #1575, opened 2024-07-31
Add Partner.projects.partnershipTypes column

atGit2021:
[Monday](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/7046940396)


atGit2021:
@CarsonF There is a scenario that fails to return the correct partnerships so I think we may need your help with the cypher intersect logic.

CarsonF:
> @CarsonF There is a scenario that fails to return the correct partnerships so I think we may need your help with the cypher intersect logic. 

Probably need to customize the client side filtering for when we use that instead of API filters.

src/scenes/Partners/Detail/Tabs/Projects/PartnerDetailProjects.tsx
```diff
@@ -45,14 +48,35 @@ export const PartnerDetailProjects = () => {
...
+      width: 160,
+      sortable: false,
+      serverFilter: ({ value }) => ({
+        partnerships: { types: typeof value === 'string' ? [value] : value },
+      }),
+      valueGetter: (_, { partnership }) => partnership.types.value,
+    },
+    ...ProjectColumns.slice(getStatusColumnIndex('status', 0)),
+  ];
+
```
willdch, line 71: @CarsonF  Hope you can help us here. We tried for a long time to use the .splice method here but the type errors we encountered with it were beyond us.   Andrew gave it a go as well but wasn't able to figure it out. 

Here is what we attempted to do: 

```tsx
import {
  PartnerProjectDataGridRowFragment as PartnerProject
} from './PartnerProjects.graphql';
```

```tsx
const partnershipTypeColumn = {
    field: 'partnerships.types',
    ...multiSelectColumn(PartnerTypeList, PartnerTypeLabels),
    headerName: 'Partnership Type',
    width: 160,
    sortable: false,
    serverFilter: ({ value }) => ({ name: value }),
    valueGetter: (_, { partnership }) => partnership.types.value,
  } satisfies GridColDef<PartnerProject>;

  const statusColumnIndex =
    ProjectColumns.findIndex((column) => column.field === 'status') + 1;

  const ProjectPartnerColumns = [
    ...splice(ProjectColumns, statusColumnIndex, 0, partnershipTypeColumn),
  ];
  ```

Comment (CarsonF) [reply src/scenes/Partners/Detail/Tabs/Projects/PartnerDetailProjects.tsx:71]: Actually there's a builtin `.toSpliced()` method now, but our `splice()` should do the same thing.
```ts
['a', 'b', 'c', 'd', 'e'].toSpliced(2, 0, 'z')
=> ['a', 'b', 'z', 'c', 'd', 'e']
```

atGit2021:
@willdch I revised to use toSpliced() per Carson's feedback.  I think all that is left is to remove the getStatusColumnIndex method and just use the findIndex for status directly if I remember Carson's other feedback.

atGit2021:
@CarsonF This is ready for your review now.

src/scenes/Partners/Detail/Tabs/Projects/PartnerDetailProjects.tsx
```diff
@@ -61,3 +68,22 @@ export const PartnerDetailProjects = () => {
...
+    field: 'partnerships.types',
+    ...multiSelectColumn(PartnerTypeList, PartnerTypeLabels),
+    headerName: 'Partnership Type',
+    width: 160,
+    serverFilter: ({ value }) => ({
+      partnerships: { types: value },
+    }),
+    valueGetter: (_, { partnership }) => partnership.types.value,
+  },
+];
```
willdch, line 83: @CarsonF 
We tried to specifically type this to the correct fragment like you said to do but we were unable to get it working.  It says that the `ProjectDataGridRowFragment` and the `PartnerProjectDataGridRowFragment` are incompatible.  We tried several things to get past this without success.  I am sure you will be able to figure this out.  Maybe you, @atGit2021 and I can zoom about this tomorrow.

CarsonF (approved):
Thanks guys! What do you guys think of my changes?

atGit2021:
> Thanks guys! What do you guys think of my changes?



> Thanks guys! What do you guys think of my changes?

Looks great @CarsonF!  I'm guessing that the value of 'is' and 'isAnyOf' could not be changed/overridden? Not sure if you were able to test that.

CarsonF:
> I'm guessing that the value of 'is' and 'isAnyOf' could not be changed/overridden? Not sure if you were able to test that.

I worked around it by just changing the labels. The filter functions implement the "has" words already. As does the API. So the operator value kinda doesn't matter.

willdch:
Great job Carson, I like it.  Thanks for taking time to help us with the type issues.

willdch:
Is the benefit of passing the `valueFormatter` to your `valuesFormatter` mainly a type reinforcement thing or are there other benefits to doing it that way?

CarsonF:
> @CarsonF Is the benefit of passing the `valueFormatter` to your `valuesFormatter` mainly a type reinforcement thing or are there other benefits to doing it that way?

Well the filter operator's `getValueAsString` was missed and an important thing. It's how the filter value is rendered in the tooltip on the filters button.
I extracted to `valuesFormatter` so it could be used in the `isAnyOf` filter and column's `valueFormatter`.
Both of which needed to map to the enum value's label.
The column's `valueFormatter` previously missed this. It doesn't matter for this use, where the `PartnerType` doesn't have different labels but it does for other enums.
I just thought it was fun to re-use all of the configuration from `enumColumn` where possible.
Since it all still applies here, just slightly different as a list.
Using the `base.valueFormatter` keeps the label formatter solely within `enumColumn`.
This also means that this "multi" "override" could just be a wrapper for `enumColumn`, even though I didn't do that yet.
```ts
...multiEnumColumn(enumColumn(PartnerTypeList, PartnerTypeLabels)),
```

PR #1577, opened 2024-08-02
Prevent Calendar Popper from Overflowing

rdonigian:
https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/7106310825

### DIFF
--- a/src/components/form/DateField.tsx
+++ b/src/components/form/DateField.tsx
@@ -123,6 +123,16 @@ export const DateField = ({
           ],
         },
       }}
+      PopperProps={{
+        modifiers: [
+          {
+            name: 'preventOverflow',
+            options: {
+              altAxis: true,
+            },
+          },
+        ],
+      }}
       showDaysOutsideCurrentMonth
       {...rest}
       ref={ref}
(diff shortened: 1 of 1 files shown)


CarsonF (commented):
😧 can you explain this change? Why isn't it the default?

rdonigian:
> 😧 can you explain this change? Why isn't it the default?

Renato posted the image of what's going on when a user zooms, which is cutting off the bottom of the calendar from an overflow issue.  This ensures the calendar stays on the page as a whole.

rdonigian:
Here is a before and after 
Screenshot 2024-08-02 at 1 42 30 PM
Screenshot 2024-08-02 at 1 42 50 PM

rdonigian:
I don't believe I see anything in our config that's preventing that from being the default.

willdch (approved):
I like it, I like it a lot.  
Great work @rdonigian

CarsonF (approved):
LGTM. Can we just add a comment saying this modifier prioritizes keeping the popup on screen sacrificing keeping the text input visible?

CarsonF (approved):
Thanks @rdonigian

PR #1581, opened 2024-08-12
HOTFIX: Render Visual Media in Quarterly Reports

andrewmurraydavid:
There was an issue where the user uploaded a video and because the UI rendered only Image files, the user or support team were not able to remove the file.

This change allows visual media files to be rendered and adds a filter for image uploads on Quarterly Reports.


CarsonF (commented):
If we are only allowing images to be uploaded then is there a need to check and show the other types?

andrewmurraydavid:
@CarsonF the reason we're checking for other types in this change is for existing files that we are unaware of that are not displayed and can not be removed. That way the user has a way to remove the files and upload images instead.

andrewmurraydavid:
@CarsonF you can see the before and after in the screenshots in the description.
Hopefully that provides more context as to why this change is needed for files we haven't been told about.

src/scenes/ProgressReports/EditForm/Steps/Media/ImageField.tsx
```diff
@@ -88,15 +91,31 @@ export const ImageField = ({
...
+          sx={{
+            display: 'flex',
+            justifyContent: 'center',
+            alignItems: 'center',
+            maxWidth: 1,
+            borderRadius: 1,
+            aspectRatio: current.dimensions.aspectRatio,
+          }}
+        >
+          <Typography variant="button">Not an image</Typography>
```
CarsonF, line 116: This isn't my favorite thing. Can we use another variant? Body1/2?

src/scenes/ProgressReports/EditForm/Steps/Media/ImageField.tsx
```diff
@@ -88,15 +91,31 @@ export const ImageField = ({
...
+          src={current.url}
+          crossOrigin="use-credentials"
+          sx={{
+            maxWidth: 1,
+            borderRadius: 1,
+          }}
+        />
+      ) : (
+        <Card
+          elevation={5}
```
CarsonF, line 106: The Card within the accordion "card" looks weird. Maybe this can just be an outlined variant?

CarsonF (approved):
LGTM just a couple nit picks. Thanks @andrewmurraydavid!

src/scenes/ProgressReports/EditForm/Steps/Media/ImageField.tsx
andrewmurraydavid: Sure. I used button as it was upper case. I can use body1 instead if you'd like.

src/scenes/ProgressReports/EditForm/Steps/Media/ImageField.tsx
andrewmurraydavid: Yup. I was going back and forth with this one. I liked the outline varian, but the border color didn't match the fields borders color.

PR #1586, opened 2024-09-12
Dashboard Grid Layout

andrewdavid48:
In this change we're adding the ability for configurable widgets and a My Dashboard page.

The Dashboard grid uses a CSS Grid layout (defaults to 12 columns and rows). My hope is that this can also be a foundation for customizable (drag and drop and/or resizable widgets)

[Monday Task](https://seed-company-squad.monday.com/boards/5989610236/views/132288519/pulses/7306763130)


src/components/Widgets/TableWidget.tsx
```diff
@@ -0,0 +1,40 @@
...
+  ...props
+}: TableWidgetProps) => {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const maxHeight = containerRef.current?.clientHeight ?? 300;
+
+  return (
+    <GenericWidget {...props} ref={containerRef}>
+      <Box
+        sx={({ spacing }) => ({
+          height: '100%',
```
willdch, line 19: May not matter but this would be more consistent with what you did in other places.
```css
  height: 1
```

src/components/Widgets/BaseWidget.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { BaseWidgetProps } from './widgetConfig';
+
+export const BaseWidget = ({ ...props }: BaseWidgetProps) => {
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  return (
+    <GenericWidget {...props} ref={containerRef}>
+      <Box
+        sx={({ spacing }) => ({
+          height: '100%',
```
willdch, line 13: same here
 ```css 
height : 1
``` 
would be more consistent.

**Review by willdch — APPROVED** [review]
I think this looks good. Just noticed the two small nit picks I commented on.

File: src/components/Widgets/GenericWidget.tsx
Comment (CarsonF) [inline src/components/Widgets/GenericWidget.tsx:1]: Instead of inheritance style, where there's one component aware of all of the needed props for data grids, widget size, widget header content, widget internal layout, I would prefer composition. Break all of these components up so each has a small piece with small number of props.

```tsx
<Widget colSpan={8} rowSpan={6}>

  <WidgetHeader expandTo="/dashboard/progress-reports">
    Quarterly Reports
  </WidgetHeader>

  <WidgetBody></WidgetBody>

  <TableWidgetBody>
    <DataGridPro {...} />
  </TableWidgetBody>

</Widget>
```

src/components/Widgets/GenericWidget.tsx
```diff
@@ -0,0 +1,135 @@
...
+          {
+            display: 'grid',
+            flexGrow: 1,
+            gridColumn: colSpan ? `span ${colSpan}` : cols,
+            gridRow: rowSpan ? `span ${rowSpan}` : rows,
+          },
+          ...extendSx(sx),
+        ]}
+      >
+        <Card
```
CarsonF, line 63: I think these two divs could be combined.

src/components/Widgets/TableWidget.tsx
```diff
@@ -0,0 +1,40 @@
...
+        })}
+      >
+        <DataGridPro
+          density="compact"
+          {...dataGridProps}
+          columns={columns}
+          sx={{
+            '--unstable_DataGrid-radius': '0px',
+            border: 'none',
+            maxHeight,
```
CarsonF, line 33: Instead of using JS to find element height, you can use a container query.
```tsx
<Box sx={{ containerType: 'size' }}>
  <DataGridPro sx={{ height: '100cqh' }} />
</Box>
```

CarsonF (changes requested):
Great start guys

src/components/Widgets/TableWidget.tsx
rdonigian: The DataGrid implementation is being moved to the specific Widget (ie: TeamNewsWidget) and will be addressed there.

src/components/Widgets/TableWidget.tsx
rdonigian: The 1 here would be a reference to MUI's theme.spacing which is not what we are going for.

src/components/Widgets/BaseWidget.tsx
rdonigian: Same as above.

CarsonF:
Thanks guys! Let's just merge this into #1587 to track in one PR

PR #1588, opened 2024-09-16
Explanation of Progress - Schedule Status Conditions

atGit2021:


src/scenes/ProgressReports/EditForm/Steps/TeamNews/TeamNewsStep.tsx
```diff
@@ -40,15 +40,21 @@ export const TeamNewsStep: StepComponent = ({ report }) => {
...
+        (responsibleRole ? currentUserRoles.has(responsibleRole) : false) &&
+        // empty and editable
+        !response.value &&
+        response.canEdit
+    ) ?? true;
+  return {
+    isIncomplete: isIncomplete,
+    severity: NotificationSeverity.Suggested,
+  };
+};
```
CarsonF, line 57: I think this looks cleaner
```suggestion
TeamNewsStep.isIncomplete = ({ report, currentUserRoles }) => ({
  isIncomplete:
    report.teamNews.items[0]?.responses.some(
      ({ variant: { responsibleRole }, response }) =>
        // responsible
        (responsibleRole ? currentUserRoles.has(responsibleRole) : false) &&
        // empty and editable
        !response.value &&
        response.canEdit
    ) ?? true,
  severity: 'suggested',
});
```

src/scenes/ProgressReports/EditForm/ProgressReportContext.tsx
```diff
@@ -53,11 +58,29 @@ export const ProgressReportContextProvider = ({
...
+              label,
+              step,
+              incompleteStatus: { severity },
+            } = item;
+            return [label, step, severity] as [
+              string,
+              StepComponent,
+              NotificationSeverityType
+            ];
+          });
```
CarsonF, line 83: I would combine the map & filter into flatMap to simplify and to not have weird fallback values.
Also there shouldn't be a need to type cast the final tuple - that can hide errors. If actually needed for some reason, which I'm not convinced it is, you should you use `satisfies` instead of `as`.

```ts
const incompleteSteps = steps
  .flatMap(([label, step]) => {
    if (!report || !step.isIncomplete) {
      return [];
    }
    const { isIncomplete, severity } = step
      .isIncomplete({ report, currentUserRoles });
    return [[label, step, severity] as const];
  });
```

src/scenes/ProgressReports/EditForm/Steps/step.types.ts
```diff
@@ -7,10 +7,28 @@ export type StepComponent = ComponentType<ReportProp> & {
...
+) => {
+  isIncomplete: boolean;
+  severity: NotificationSeverityType;
+};
 
 export type GroupedStepMapShape = {
   [Section in string]: ReadonlyArray<[label: string, component: StepComponent]>;
 };
+
+export type GroupedStepMapShapeWithSeverity = {
```
CarsonF, line 27: I'd call this
```suggestion
export type GroupedIncompleteSteps = {
```

src/scenes/ProgressReports/EditForm/Steps/step.types.ts
```diff
@@ -7,10 +7,28 @@ export type StepComponent = ComponentType<ReportProp> & {
...
+  [Section in keyof GroupedStepMapShape]: Array<
+    [
+      ...GroupedStepMapShape[Section][number],
+      severity: NotificationSeverityType
```
CarsonF, line 31: I wouldn't bother trying to spread from the other type here. While clever, it's confusing. And this type also has nothing to do with the other one.
```suggestion
      label: string,
      component: StepComponent,
      severity: NotificationSeverityType
```

Alternatively, you could use an object instead of a tuple here. That might be clearer everywhere.
```ts
ReadonlyArray<{
  label: string;
  step: StepComponent;
  severity: IncompleteSeverity;
}>

File: src/scenes/ProgressReports/EditForm/ProgressReportContext.tsx (hunk 2 above)
Comment (atGit2021) [reply src/scenes/ProgressReports/EditForm/ProgressReportContext.tsx]: @CarsonF I made an attempt at the revisions inside the ProgressReportContext file, and although it looks like the incompleteSteps function starting on line 57 is outputting the correct shape, I'm getting an error with the shape down on line 93.  The error I get is:

`Type '{ [k: string]: (readonly [string, StepComponent, IncompleteSeverity])[]; }' is missing the following properties from type 'readonly [label: string, step: StepComponent, severity: IncompleteSeverity][]': length, concat, join, slice, and 26 more.ts(2740)
ProgressReportContext.tsx(15, 3): The expected type comes from property 'incompleteSteps' which is declared here on type 'ProgressReportContext'
(property) ProgressReportContext.incompleteSteps: GroupedIncompleteSteps`

I've tried a bunch of things and tried some chatGPT suggestions, but none seem to resolve my issue. I will need your help with this one.

Comment (atGit2021) [reply src/scenes/ProgressReports/EditForm/ProgressReportContext.tsx]: @CarsonF I pushed up another commit trying to conform the code to what you were requesting but had to make other modifications to the type to make it work.  Please take a look, thank you.

Comment (atGit2021) [reply src/scenes/ProgressReports/EditForm/ProgressReportContext.tsx]: Ok, I was under the impression that you wanted the type to be literally just the array based on your other feedback, see below.  Sorry, my misunderstanding, thank you.

`Alternatively, you could use an object instead of a tuple here. That might be clearer everywhere.
ReadonlyArray<{
  label: string;
  step: StepComponent;
  severity: IncompleteSeverity;
}>`

Comment (CarsonF) [reply src/scenes/ProgressReports/EditForm/ProgressReportContext.tsx]: Yeah sorry I was just referencing the tuple (within record) to change to an object (within the record).

**Comment by atGit2021** [comment]
@CarsonF , There is one more requirement in the task that I'm looking at which pertains to the "required" error message for the Select a Reason radio button component, see req below.  While investigating I encountered an existing bug in the logic and secondly I'm not sure how to reposition the error message at the top of the component, aligned with the title/label. Do you have time tomorrow to huddle on this one? 
```
Move "Required" text to same row as "Select a Reason" when Behind/Delayed or Ahead are selected
Error message
Please select a reason
A reason must be selected when reporting 10% behind or 10% ahead
```

File: src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.tsx [hunk 5]
```diff
@@ -51,6 +51,20 @@ interface FormShape {
...
+  <Typography variant="h3" gutterBottom>
+    Explanation of Progress
+  </Typography>
+);
+
+const ExplanationHeaderNotice = ({ group }: { group: string }) =>
+  group !== 'onTime' ? (
+    <Typography paragraph>
+      Select a reason below: Required for Language Engagements showing progress
+      is ahead or behind planned progress.
```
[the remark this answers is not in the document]
atGit2021: Sheri asked me to change it

src/scenes/ProgressReports/EditForm/Steps/ExplanationOfProgress/ExplanationOfProgress.tsx
```diff
@@ -169,8 +172,15 @@ export const ExplanationOfProgress: StepComponent = ({ report }) => {
...
                 name="reasons"
-                label="Select a reason"
-                required
+                label={
+                  <>
+                    Select a reason.{' '}
+                    <span style={{ color: 'red' }}>
+                      A reason must be selected when progress variance is
+                      greater than 10%.
+                    </span>
```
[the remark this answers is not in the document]
atGit2021: I'll remove the color red styling, and add "(Required)" after the field label

CarsonF (approved):

PR #1591, opened 2024-10-01
Adjust ProgressReport rich text cell styling

CarsonF:
- More consistent font size
- lists render better & float around the variant icon
- Fixed line clamp in Safari
- Variance Explanation uses same wrapper now so it line clamps & expands like the other two.
- I increased its default width too

Screenshot 2024-10-01 at 6 50 36 PM
Screenshot 2024-10-01 at 6 50 42 PM

Safari proof:
Screenshot 2024-10-01 at 6 50 57 PM

### DIFF
--- a/src/scenes/Dashboard/ProgressReportsWidget/ProgressReportsExpandedGrid.tsx
+++ b/src/scenes/Dashboard/ProgressReportsWidget/ProgressReportsExpandedGrid.tsx
@@ -1,5 +1,7 @@
+import { Box } from '@mui/material';
 import {
   DataGridProProps as DataGridProps,
+  GridRenderCellParams,
   GridRowId,
   GridToolbarColumnsButton,
   GridToolbarFilterButton,
@@ -24,9 +26,34 @@ import {
 
 const COLLAPSED_ROW_HEIGHT = 54;
 
+// Position cell text consistently regardless of expansion
+const NonExpansionCell = ({ formattedValue }: GridRenderCellParams) => (
+  <Box
+    sx={{
+      '--height': `${COLLAPSED_ROW_HEIGHT}px`,
+      // Causes row to grow by 1px on row height auto
+      // Shift the 1px to padding to prevent.
+      height: 'calc(var(--height) - 1px)',
+      pt: '1px',
+      // This is exactly what MUI does when not auto height
+      lineHeight: 'calc(var(--height) - 1px)',
+
+      display: 'flex',
+      alignItems: 'center',
+    }}
+  >
+    {formattedValue}
+  </Box>
+);
+
 const columns = entries(ProgressReportsColumnMap).map(([name, col]) => ({
   field: name,
   ...col,
+  ...(!(
+    'cellClassName' in col && col.cellClassName.includes(ExpansionMarker)
+  ) && {
+    renderCell: NonExpansionCell,
+  }),
 }));
 
 const initialState = {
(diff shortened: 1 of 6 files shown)


src/scenes/Dashboard/ProgressReportsWidget/ProgressReportsExpandedGrid.tsx
rdonigian, line 1: 16px is creating a small jump in lines that have no text.  16.5px seems to work better.

src/scenes/Dashboard/ProgressReportsWidget/ProgressReportsExpandedGrid.tsx
```diff
@@ -96,7 +96,7 @@ export const ProgressReportsExpandedGrid = (props: Partial<DataGridProps>) => {
           [`.MuiDataGrid-row--dynamicHeight .MuiDataGrid-cell:not(.${ExpansionMarker})`]:
             {
               // Magic number to keep text vertically unchanged when the row is expanded
-              pt: '17.5px',
+              pt: '16px',
```
rdonigian, line 99: 16px is creating a small jump in lines that have no text. 16.5px seems to work better.

src/scenes/Dashboard/ProgressReportsWidget/RichTextCell.tsx
```diff
@@ -0,0 +1,96 @@
...
+  value,
+  api,
+  sx,
+  className,
+}: Pick<CellParams, 'value' | 'id' | 'api'> & StyleProps) => {
+  const selectedRows = useGridSelector(
+    { current: api },
+    (state: GridState) => state.rowSelection
+  );
+  const isExpanded = selectedRows[0] === id;
```
rdonigian, line 29: We are calling this `selectedRows` then treating it as if there will be only one element.  I'd recommend either calling this `selectedRow` or running .find on this array.

src/scenes/Dashboard/ProgressReportsWidget/ProgressReportsExpandedGrid.tsx
CarsonF: 16.5 causes a jump in Safari. 16 doesn't cause a jump for me in either...

src/scenes/Dashboard/ProgressReportsWidget/RichTextCell.tsx
CarsonF: We can't just rename it, it'll still be an array.
This logic goes hand-in-hand with https://github.com/SeedCompany/cord-field/blob/7d22d905654fe30af90cd8289b9ad9c79a538447/src/scenes/Dashboard/ProgressReportsWidget/ProgressReportsExpandedGrid.tsx#L85-L89
which only ever has one element.

If you feel strongly about this though we can switch it to an `.includes()`

src/scenes/Dashboard/ProgressReportsWidget/RichTextCell.tsx
```diff
@@ -0,0 +1,96 @@
...
+  <Typography variant="body2" paragraph>
+    <Text data={data} />
+  </Typography>
+);
+
+const List: RenderFn<{ style: 'unordered' | 'ordered'; items: string[] }> = ({
+  data,
+}) => {
+  // eslint-disable-next-line react/jsx-no-useless-fragment
+  if (!data) return <></>;
```
rdonigian, line 65: Why is List handling no data?  Can we handle this elsewhere?

CarsonF: We could. But would need more work, in a different PR. Currently we don't own the component that calls these components.

CarsonF: Actually no we couldn't. Because certain blocks could have no data but that doesn't mean the block shouldn't render. Like a `delimiter` just renders a `Divider`, no data needed for that.

rdonigian (approved):

src/scenes/Dashboard/ProgressReportsWidget/ProgressReportsExpandedGrid.tsx
CarsonF: Ok I think I've got it 000e304
You were on the right track with the `line-height` - it just needed to be the full height (overriding `inherit` set by auto row height).

PR #1596, opened 2024-10-03
Fix DataGrid sorting for initialInput 'DESC' 

rdonigian:


CarsonF:
I was bothered by the null chaining here. We know we have a sort item, so I updated types to reflect that.

I also thought that your logic was a bit hard to follow. And it is by chance that the sort selection order is asc -> desc -> null. I think this could be flipped around separately. So I think what I have is more clear on what we are actually doing and ensure we flip the order correctly regardless of grid configuration.

What do you think?

rdonigian:
> I was bothered by the null chaining here. We know we have a sort item, so I updated types to reflect that.
> 
> I also thought that your logic was a bit hard to follow. And it is by chance that the sort selection order is asc -> desc -> null. I think this could be flipped around separately. So I think what I have is more clear on what we are actually doing and ensure we flip the order correctly regardless of grid configuration.
> 
> What do you think?

rdonigian:
I like that approach, I was fighting the optional chaining.  LGTM

CarsonF (approved):

PR #1603, opened 2024-10-10
Comments Tweaks

CarsonF:
- Badge color is now primary
- Badge count is thread count, not all comments
- Threads are sorted DESC
- Threads are only loaded if the sidebar is open
- Thread count for badge is loaded separately if needed
- Some other minor cache tweaks

I moved to loading from `Commentable` this opens open the room for something like
```gql
query {
  language(id: "") {
    commentThreads { total }
  }
}
```
To load the badge count (or more info) without a separate network request.
This is only a half baked idea, so nothing like that changed here.
In general, I think it makes more sense from DX because we view the comments under a resource and now this & cache reflect that.
Performance wise before and after are the same. We always loaded the `Commentable` first anyways in the API.


andrewmurraydavid (approved):
Seems awkward having the count load separately from the rest of the data. But overall looks good.

CarsonF:
> Seems awkward having the count load separately from the rest of the data.

That's an optimization. If the sidebar isn't open we only load the total count. If it is open, then the sidebar loads the total count and the button just uses that query (via cache).

PR #1607, opened 2024-10-15
Enable sort and filter on progress column

atGit2021:
[Monday](https://seed-company-squad.monday.com/boards/3451697530/pulses/7551127713?doc_id=7551149783)

### DIFF
--- a/src/scenes/Dashboard/ProgressReportsWidget/ProgressReportsGrid.tsx
+++ b/src/scenes/Dashboard/ProgressReportsWidget/ProgressReportsGrid.tsx
@@ -97,7 +97,7 @@ export const ProgressReportsColumnMap = {
     ...enumColumn(ScheduleStatusList, ScheduleStatusLabels, {
       orderByIndex: true,
     }),
-    valueGetter: (_, row) => row.varianceExplanation.scheduleStatus,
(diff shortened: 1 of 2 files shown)


atGit2021:
@CarsonF will need your help with an ASC/DESC sort issue.  When the sort arrow is clicked to ASC the API sends back the list in DESC order.  When the DESC arrow is clicked it sends back the list in ASC order.

CarsonF:
API appears to sort fine to me.
I fixed the client side sorting here by ordering by index based on the API defined order (which I also just enabled https://github.com/SeedCompany/cord-api-v3/commit/4beaa9d656e1df7545c0cb5ff98537f735362369)

Also the variance reason given can override the schedule status according to the variance explanation.
```ts
varianceExplanation: {
  reasons: ["I am ahead"]
  scheduleStatus: Behind // doesn't make sense
}
```
So that's why we follow the reason there.

The schedule status on cumulative summary just looks at its own variance.
So that more closely follows what we want to display here I think, since the summary numbers are right next to it.

I had, and maybe you did too, picked a random reason, and was seeing "Ahead" but it wasn't sorting that way. Because the sorting just looks at the variance number, not the reasons.

PR #1610, opened 2024-10-17
Fix panel "Add filter" bug

andrewmurraydavid:
It seems as if when the slot props where merging, the slot props from the expanded where not merging properly, causing the filter to not work.

Instead of duplicating the code for generating the dataGridProps, we are instead exposing the dataGripProps "build" and reusing it in the Expanded view.


atGit2021 (approved):
Add Filter works great now, thank you @andrewmurraydavid

CarsonF:
Fixed with 37caedcbc385765cb69e3193e3f0b83071081ec5

PR #1632, opened 2025-01-17
Pnp Problems Widget

rdonigian:
Creation of the PnP Problems Widget and adding it to the the dashboard.


CarsonF (approved):
Made a few changes here. There's still a lot duplicated, but I'm hesitant to DRY more because a) it's a lot of work and b) it could artificially constrain functionality that happens to be currently aligned but could change in future.

I think this is good for now 👍

PR #1634, opened 2025-01-27
Create Field Region and Field Zone 

rdonigian:
User will have the option to create either a new field region or new field zone on the UI if necessary.  This was currently being handled exclusively by dev's through the api.


src/components/FieldRegion/CreateFieldRegion.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { CreateFieldRegionDocument } from './CreateFieldRegion.graphql';
+
+/*
+  CreateFieldRegionInput doesn't properly reflect the input from the fields, therefore the need for FieldRegionInput
+*/
+export interface FieldRegionInput {
+  fieldRegion: { name: string };
+  fieldZone: Partial<FieldZone>;
+  director: Partial<User>;
+}
```
CarsonF, line 18: Take a look here:
https://github.com/SeedCompany/cord-field/blob/884a41d348a8594e2c6cb509f348f507ef19f539/src/scenes/Locations/LocationForm/LocationForm.tsx#L32-L44

We should try to follow suit of the input shape. That would put `director` & `fieldZone` under the `fieldRegion` sub object.
Also we shouldn't pull those objects (`FieldZone`, `User`) from the schema file directly. Instead we should use the fragments that those fields require: `DisplayFieldZoneFragment`, `UserLookupItemFragment`

CarsonF (approved):
LGTM!

I just changed the form values to be consistent with the API, so that it can submit field errors that we can connect automagically.

PR #1637, opened 2025-01-30
Current and Next Quarterly Report Link

rdonigian:
On the engagement page, turn current and next text into corresponding links to improve user experience.


src/components/ProgressReportsOverviewCard/ProgressReportsOverviewCard.tsx
```diff
@@ -78,9 +78,11 @@ const ReportInfo = ({
       <Typography variant="body2" color="text.secondary">
         {title}
       </Typography>
-      <Typography variant="h4" sx={{ my: 1 }}>
-        <ReportLabel report={report} />
-      </Typography>
+      <Link to={`/progress-reports/${report.id}`} underline="none">
+        <Typography variant="h4" sx={{ my: 1 }}>
```
CarsonF, line 82: The `Typography` props can be collapsed into `Link`. `Link` already wraps `Typography`.
I wonder if we should keep the underline as standard?

rdonigian: Yea I was 50/50 on that.  I can put it back it just to help call it out a bit better.

CarsonF (approved):

PR #1643, opened 2025-03-05
Add EthnologueLanguage to Global Search

rdonigian:


CarsonF (commented):
No we don't want this. We don't want Eth Langs coming back as top level results. We only want Langs coming back.

Your work on the API is sufficient to have Langs come back based on matches on the eth lang data points.

PR #1644, opened 2025-03-05
Convert Language List to Datagrid

rdonigian:


src/components/LanguageDataGrid/LanguageColumns.tsx
```diff
@@ -0,0 +1,88 @@
...
+    width: 100,
+    valueGetter: (_, { population }) => population.value,
+    filterable: false,
+  },
+  {
+    field: 'id',
+    headerName: 'ID',
+    valueGetter: (_, language) => language.id,
+    width: 150,
+  },
```
CarsonF, line 56: I think it's fine to have this ID column here for identification, but I'm wondering if it should just be hidden by default. I think a lot of data points we have aren't important enough to take up space here, but could be needed in certain situations. So I think it's fine to have users turn on the column. Though obviously this could use some feedback. And whether or not they know how to do that at all is something different entirely, since it applies to all grids.

CarsonF (approved):
LGTM! There are definitely more columns that could be added here but this is fine to get started with.

I did declare a fragment for this specific view instead of using the card one.

I'm fixing the population sorting here https://github.com/SeedCompany/cord-api-v3/pull/3404

PR #1656, opened 2025-04-24
Add Created At column to Partners Table 

rdonigian:


src/components/PartnersDataGrid/PartnerColumns.tsx
```diff
@@ -44,6 +44,13 @@ export const PartnerColumns: Array<GridColDef<Partner>> = [
     valueGetter: dateColumn.valueGetter((_, { startDate }) => startDate.value),
     width: 130,
   },
+  {
+    headerName: 'Created Date',
+    field: 'createdAt',
+    ...dateColumn(),
+    valueGetter: dateColumn.valueGetter((_, { createdAt }) => createdAt),
```
CarsonF, line 51: This is the default no?

src/components/PartnersDataGrid/PartnerColumns.tsx
```diff
@@ -44,6 +44,12 @@ export const PartnerColumns: Array<GridColDef<Partner>> = [
     valueGetter: dateColumn.valueGetter((_, { startDate }) => startDate.value),
     width: 130,
   },
+  {
+    headerName: 'Created Date',
+    field: 'createdAt',
+    ...dateColumn(),
```
CarsonF, line 50: This is a DateTime, not a Date. I think it will be ok for everything but this instanceof check.
https://github.com/SeedCompany/cord-field/blob/add-startDate-to-partners-table/src/components/Grid/ColumnTypes/dateColumn.tsx#L45-L45
We probably could just widen it to allow for both types:
```ts
    if (DateTime.isDateTime(value)) {
```

src/components/PartnersDataGrid/PartnerColumns.tsx
rdonigian: I hit send and then picked that up and resubmitted.

CarsonF (approved):

PR #1657, opened 2025-04-27
Make Partner Start Date column editable

rdonigian:


CarsonF (approved):
Created #1660 & #1661 out of the pain I saw here.

Rebased this onto those.

PR #1660, opened 2025-04-29
Stop implicitly parsing ISO date strings

CarsonF:
Now do it explicitly when/where needed.

It is too hard to get right currently.
The outstanding apollo client issue doesn't have any appearance of resolving this read issue.

This change better reflects reality, as we're no longer assuming luxon instances when the values are actually ISO strings.

Though I did cheat a little bit and say that they could be either an ISO string or a luxon instance, which isn't really true. But it seemed easiest to go with this for now, and then narrow to luxon instance explicitly with a function call.

### DIFF
--- a/src/common/CalenderDate.ts
+++ b/src/common/CalenderDate.ts
@@ -19,15 +19,11 @@ export type ISOString = Tagged<string, 'ISOString'>;
 export type CalendarDateOrISO = CalendarDate | ISOString;
 export type DateTimeOrISO = DateTime | ISOString;
 
-export const asDateTime = <T extends DateTimeOrISO | Nil>(
-  date: T
-): DateTime | (T extends Nil ? null : never) =>
-  asLuxonInstance(date, DateTime) as any;
+export const asDateTime = <T extends DateTimeOrISO | Nil>(date: T) =>
+  asLuxonInstance(date, DateTime) as T extends Nil ? null : DateTime;
 
-export const asDate = <T extends CalendarDateOrISO | Nil>(
-  date: T
-): CalendarDate | (T extends Nil ? null : never) =>
-  asLuxonInstance(date, CalendarDate) as any;
+export const asDate = <T extends CalendarDateOrISO | Nil>(date: T) =>
+  asLuxonInstance(date, CalendarDate) as T extends Nil ? null : CalendarDate;
 
 function asLuxonInstance(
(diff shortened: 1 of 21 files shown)


src/scenes/Projects/Overview/ProjectOverview.tsx
```diff
@@ -362,9 +362,7 @@ export const ProjectOverview = () => {
               <ChangesetPropertyBadge
                 current={project}
                 prop="mouRange"
-                identifyBy={(range) =>
-                  `${range.start?.toMillis()}/${range.end?.toMillis()}`
-                }
+                identifyBy={(range) => `${range.start}/${range.end}`}
```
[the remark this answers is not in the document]
CarsonF: They will be consistent. Strings are ISO, and the object's string conversion produces an ISO string.

src/common/CalenderDate.ts
```diff
@@ -11,6 +12,29 @@ import {
...
+
+export type ISOString = Tagged<string, 'ISOString'>;
+
+export type CalendarDateOrISO = CalendarDate | ISOString;
+export type DateTimeOrISO = DateTime | ISOString;
+
+export const asDateTime = <T extends DateTimeOrISO | Nil>(
+  date: T
+): DateTime | (T extends Nil ? null : never) =>
+  asLuxonInstance(date, DateTime) as any;
```
rdonigian, line 25: We don't want to cast as `any` if we don't have to.  What about this is an alternative
`export const asDateTime = <T extends DateTimeOrISO | Nil>(
  date: T
): DateTime | (T extends Nil ? null : never) =>
  asLuxonInstance(date, DateTime) as any;

export const asDateTime = <T extends DateTimeOrISO | Nil>(
  date: T
): T extends Nil ? null : DateTime =>
  asLuxonInstance(date, DateTime) as T extends Nil ? null : DateTime;`

CarsonF: Typecasts are always unsafe. Your `as` is not much different. I don't like repeating the same thing return type twice just to make TS happy. Here's the return type, and I know I'm right. The "unsafe-ness" is contained within this small function.
I could just let the return type be inferred from this type cast. It's basically the same thing.

It is surprising that that conditional generic returns a union. But after studying it is more right and will produce the union, given the input generic is a union.
Only better for the case of `asDateTime(null) -> null`.

PR #1664, opened 2025-05-14
Create AI Guidelines Markdowns

rdonigian:


CarsonF (commented):
There's a lot here that's incorrect.

There's a lot here that I don't feel like we need to call out.
"Use PascalCase for React Components"

PR #1668, opened 2025-05-23
Fix Create Field Region & Field Zone

rdonigian:


CarsonF (approved):
Looks fine to me. What does this actually change though?

PR #1706, opened 2025-06-27
Fix first tab in list from being cut off

rdonigian:


CarsonF:
Have you seen [this](https://github.com/SeedCompany/investor-portal/commit/6910c77d331e221de5b72903082134d29ed24239)? Is this the same problem you are trying to fix here? I've been meaning to port that IRP tab functionality here, which scales the tab in away that doesn't cause the ripple to overflow.

CarsonF (approved):
I improved the implementation by scaling down unselected items instead of scaling up the selected item.
This ensures that every tab's width is consistent with the space needed for the label/text, and thus the tab indicator line is consistent as well. It's 2 margin is consistent visually across the dynamic tab widths, and there's no need to hack/arbitrarily increase it.

Overall I think this still makes the tabs further apart from each other, but now they are correctly laid out for their larger selected size. Which obviously fixes the cut off bug.
I don't think there's any way around the increased spacing, this was the design chosen. I don't leave like it for the record lol

PR #1710, opened 2025-07-07
Add Field Region / Zone to Global Search

rdonigian:
This creates region and zone cards for display and then expands global search to those objects.


src/components/FieldRegionCard/FieldRegionCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+              <Skeleton width="75%" />
+            ) : name?.canRead === true ? (
+              name.value
+            ) : (
+              <Redacted
+                info="You don't have permission to view this field region's name"
+                width="75%"
+              />
+            )}
+          </Typography>
```
CarsonF, line 38: I wonder if we should include the director here as well? It seems like a key piece of info.

src/components/FieldRegionCard/FieldRegionCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+          color="textSecondary"
+          sx={(theme) => ({
+            pr: theme.spacing(1),
+          })}
+        >
+          {loading ? (
+            <Skeleton width="25%" />
+          ) : (
+            <>
+              Created <FormattedDateTime date={createdAt} />
```
CarsonF, line 60: Created date is kinda worthless here. These just have always existed. I'd remove.
Same with Field Zone, obv.

CarsonF (changes requested):

src/components/FieldRegionCard/FieldRegionCard.tsx
rdonigian: Yea I did wonder about that, we have the created at else where on cards so it was just carry over.

PR #1720, opened 2025-07-23
Tech Debt - Replace ErrorButton/CreateButton -> Button, remove makeStyles Part 1 

rdonigian:
This is the beginning of a work-through the code base to clean up tech debt related to the deprecated `makesStyles`.  This will also remove unnecessary components `ErrorButton`/`CreateButton`, as MUI's native `Button` can accomplish the same goal.

### DIFF
--- a/src/components/PartnershipSummary/PartnershipSummary.tsx
+++ b/src/components/PartnershipSummary/PartnershipSummary.tsx
@@ -24,7 +24,7 @@ export const PartnershipSummary = ({
       title="Partnerships"
       to="partnerships"
       members={members}
-      icon={<PeopleJoinedIcon style={{ fontSize: 42, marginTop: -8 }} />}
+      icon={<PeopleJoinedIcon sx={{ fontSize: 42, mt: -1 }} />}
     />
   );
 };
--- a/src/components/ProgressButton/ProgressButton.tsx
+++ b/src/components/ProgressButton/ProgressButton.tsx
@@ -34,10 +34,7 @@ export const ProgressButton = forwardRef<
         <div
           style={{
             position: 'absolute',
-            top: 0,
-            bottom: 0,
-            left: 0,
-            right: 0,
+            inset: 0,
             display: 'flex',
             alignItems: 'center',
             justifyContent: 'center',


src/components/PartnershipSummary/PartnershipSummary.tsx
```diff
@@ -25,15 +17,14 @@ export const PartnershipSummary = ({
...
-  const { classes } = useStyles();
 
   return (
     <MemberListSummary
       total={partnerships?.total}
       title="Partnerships"
       to="partnerships"
       members={members}
-      icon={<PeopleJoinedIcon className={classes.icon} />}
+      icon={<PeopleJoinedIcon style={{ fontSize: 42, marginTop: -8 }} />}
```
CarsonF, line 27: Not my favorite. I'd prefer that icon to accept `sx/className` props

src/components/ProgressButton/ProgressButton.tsx
```diff
@@ -46,22 +25,34 @@ export const ProgressButton = forwardRef<
...
     <>
       {progress ? (
-        <div className={classes.progressWrapper}>
+        <div
+          style={{
+            position: 'absolute',
+            top: 0,
+            bottom: 0,
+            left: 0,
+            right: 0,
```
CarsonF, line 40: Fwiw this could be `inset: 0`

src/scenes/Root/Creates/CreateMenu.tsx
```diff
@@ -46,15 +45,17 @@ export const CreateButtonMenu = (props: CreateButtonMenuProps) => {
 
   return (
     <>
-      <CreateButton
+      <Button
+        variant="contained"
+        color="error"
```
CarsonF, line 50: The big reason why I had a CreateButton component was to decouple the "create action" from the error palette. Those are kinda two different things entirely. Now what we could do is add a custom palette that would allow us to do `color="create"`. MUI has docs for this. 

That said it's possible we don't even like this style or maybe this is the only place it happens. In which case I'd say don't worry about it.

src/scenes/Root/Creates/CreateMenu.tsx
```diff
@@ -46,15 +45,17 @@ export const CreateButtonMenu = (props: CreateButtonMenuProps) => {
 
   return (
     <>
-      <CreateButton
+      <Button
+        variant="contained"
```
CarsonF, line 49: Is this the default?

CarsonF (approved):

PR #1730, opened 2025-09-30
Add User.photo

rdonigian:
API PR: https://github.com/SeedCompany/cord-api-v3/pull/3565

### DIFF
--- a/src/common/fragments/user.graphql
+++ b/src/common/fragments/user.graphql
@@ -76,4 +76,12 @@ fragment DisplayUser on User {
     canEdit
     value
   }
+  photo {
+    canRead
+    canEdit
+    value {
+      id
+      url
+    }
+  }
 }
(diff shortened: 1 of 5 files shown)


CarsonF:
Pulled gender into its own PR #1733

CarsonF:
Screenshot 2025-10-01 at 6 15 41 PM

I don't really like the current style.
Seems like the photo could be smaller, and pulled to the left of the text or the name...not sure. Just the linear list looks bleh.

As far as the form stuff and complication with file uploads, my best thought is that we do photo uploads completely separate from other user fields.
We could skip it on user create too.
Some kind of edit functionality and/or drop functionality onto the Avatar seems like an approach that would be more intuitive and would work better with our code.

Will continue thinking about this tomorrow.

rdonigian:
closed by #1735

PR #1741, opened 2025-10-16
Example of live query working with user detail page

CarsonF:
Run this with https://github.com/SeedCompany/cord-api-v3/pull/3587 to see it in action.

Open up multiple browsers/windows/tabs pointing to the same user detail page.
Any update will propagate immediately to all the windows.


CarsonF:
API & UI are all merged now. Idk if this should be merged yet. But still up for example/trial.

PR #1743, opened 2025-10-23
Query for initial Tool List 

rdonigian:
In order to display all `tool` options in drop down we're now sending an `tool` list query.


CarsonF:
@rdonigian how's this look/function to you?

PR #1776, opened 2026-03-02
Create Rev79 Id input fields for Projects and Engagements

rdonigian:
API PR: https://github.com/SeedCompany/cord-api-v3/pull/3666

This adds a "uses Rev79" checkbox to the Project edit for Momentum Projects.  If true, a Rev79 project id and engagement community id input will be rendered.


brentkulwicki (approved):
I did not pull and test locally, but everything makes sense and I have no issues when looking at the code. ✅

PR #1779, opened 2026-03-11
Fix dark mode login

### LINKED ISSUE #1758: Add a Dark Mode using the updated Seed Company colors
Create a dark mode view that respects the browser/machine settings using the new brand colors:

Natural -  #F7F1E7
Stone - #CDC3B0
White - #FFFFFF
Light Gray - #EBEBEC
Dark Gray - #636466
Black - #323232

DanCoria:
## Summary
Fixes dark mode based on the user's system color scheme preference (`prefers-color-scheme`).

### Login Page
- `AuthLayout`: removed hardcoded forced-dark `ThemeProvider` override; login page now inherits app-level theme
- Added `Scrim` overlay — dark tint in dark mode, transparent in light mode — for background image contrast
- Removed `lazy` from `<Picture>` to prevent solid black flash on logout navigation

### Known Limitation
- One-frame white flash on first paint in dark mode due to SSR hydration mismatch. `noSsr: true` minimizes this but cannot eliminate it without `InitColorSchemeScript` (MUI v6+). Tracked as future tech debt.

## Testing
- ✅ Light mode (macOS System Settings → Appearance → Light)
- ✅ Dark mode (macOS System Settings → Appearance → Dark)
- ✅ Login page renders correctly in both modes
- ✅ Logout redirects to login without solid black flash
- ✅ Rest of app (projects, partners, languages) renders correctly in both modes

### DIFF
--- a/src/scenes/Authentication/AuthLayout.tsx
+++ b/src/scenes/Authentication/AuthLayout.tsx
@@ -1,6 +1,8 @@
-import { styled } from '@mui/material/styles';
+import { CssBaseline } from '@mui/material';
+import { styled, ThemeProvider } from '@mui/material/styles';
 import { Outlet } from 'react-router-dom';
 import { ChildrenProp } from '~/common';
+import { createTheme } from '~/theme';
 import { Picture } from '../../components/Picture';
 import backgroundImg from './background.png';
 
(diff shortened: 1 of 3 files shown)


sethmcknight (dismissed):

rdonigian (commented):
The login page is natively dark, I don't see a problem with keeping it that way.  In which case we only need to set
```
const Root = styled('div')(({ theme }) => ({
  ...,
  backgroundColor: theme.palette.background.default,
}));
```

Otherwise if you think the login should represent light/dark as well, then we can make the same tweak to Root styles and remove the `ThemeProvider` & `CssBaseline` from AuthLayout.  

We set the `color=text.primary` in the overrides file last PR so no need for the redundancy.

rdonigian (approved):

rdonigian (approved):

sethmcknight (approved):

PR #1797, opened 2026-04-06
Add Apollo preflight header

rdonigian:
Adds apollo-require-preflight: 1 to the Apollo HttpLink headers.

This repo uses APQ, which sends an initial GET request — Apollo Server's csrfPrevention rejects these with a 400 before auth runs.

The apollo-require-preflight header is the recommended bypass with no semantic side effects


bryanjnelson (approved):
🚀

PR #1798, opened 2026-04-07
Inject Seed Env Vars

rdonigian:


bryanjnelson (dismissed):
👍

brentkulwicki (dismissed):
Do you know if these are being used client side? If they aren't necessary for client side, then I believe you can drop the `RAZZLE_` and it becomes only serverside available.

rdonigian:
@brentkulwicki good point, this isn't being used client side.

bryanjnelson (approved):

PR #1802, opened 2026-04-09
Give preference to Language.publicName

rdonigian:


src/components/EngagementDataGrid/EngagementColumns.tsx
```diff
@@ -59,7 +59,8 @@ export const EngagementColumns: Array<GridColDef<Engagement>> = [
     width: 200,
     valueGetter: (_, row) => {
       return row.__typename === 'LanguageEngagement'
-        ? row.language.value?.name.value
+        ? row.language.value?.displayName.value ??
+            row.language.value?.name.value
         : row.__typename === 'InternshipEngagement'
```
brentkulwicki, line 63: Not going to say this is blocking, but the ternary chaining here might be getting out of hand and need to be re-thought at some point.

brentkulwicki (dismissed):

src/components/EngagementDataGrid/EngagementColumns.tsx
rdonigian: Lol fair, I think the more permanent solution here would probably create a resolve field on the api that can handle this logic and then serve that up similar to how we handle `User` `fullName`, which is just a resolve field that hands the FE first the users `displayFirst` + `displayLast`, if it exists, otherwise `realFirst` + `realLast`

brentkulwicki (approved):

PR #1803, opened 2026-04-10
Enhance Project Workflow Visibility

### LINKED ISSUE #1781: Enhance the visibility of the workflow UI feature/page
Currently, users are primarily informed about the Project Workflow feature via a direct link to the workflow page: (https://cordfield.com/projects/workflow). 

Image

There is also an "info" button on the project page's status button that links to the same page:

Image

This feature is very helpful for both Field and Finance users, but it is not easily discoverable. I frequently have to direct users to this page, including the same users on multiple occasions, which suggests that the current entry points are not intuitive enough.

DanCoria:
Replace InfoIcon with a styled Button for viewing workflow in ProjectWorkflowDialog

### DIFF
--- a/src/scenes/Projects/Update/ProjectWorkflowDialog.tsx
+++ b/src/scenes/Projects/Update/ProjectWorkflowDialog.tsx
@@ -1,6 +1,5 @@
 import { useMutation } from '@apollo/client';
-import { InfoOutlined as InfoIcon } from '@mui/icons-material';
-import { Grid, IconButton, Stack, Tooltip, Typography } from '@mui/material';
+import { Button, Grid, Stack, Tooltip, Typography } from '@mui/material';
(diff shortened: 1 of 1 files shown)


rdonigian (commented):
For a lot of these common components we already have custom ones built out. So we can re-use to keep consistency.
```
<ButtonLink
            to={`/projects/workflow${step ? `?state=${step}` : ''}`}
            variant="text"
            size="small"
            color="inherit"
            sx={{ color: 'text.secondary' }}
          >
            View Workflow
 </ButtonLink>
```

DanCoria:
Thanks for pointing that out @rdonigian. I have updated the file to use the ButtonLink component and changed the variant="text" to variant="outline" to keep the same pill-look we are going for.

rdonigian (approved):

PR #1815, opened 2026-04-24
Language Lookup filter for isAvailableForReporting

### LINKED ISSUE #1808: Filter Language of Reporting dropdown on Partner page to only show languages available for reporting
## Problem

The Language of Reporting dropdown on the Partner record is currently showing all languages, even though only languages marked as available for reporting should be selectable.

As a result, a user can select an invalid option from the dropdown and then receive an error when saving, rather than only seeing valid options up front.

Image

## Expected Behavior

The Language of Reporting dropdown should display only languages marked as available for reporting.

## Actual Behavior

The dropdown currently displays all languages. If a user selects an option that is not available for reporting, the system throws an error rather than preventing the selection in the UI.

rdonigian:


brentkulwicki (approved):
Full disclosure, not tested locally, but the code looks good.

PR #1852, opened 2026-08-10
Fix: show final notes on progress report status card

### LINKED ISSUE #1851: Final Notes on Quarterly Reports Not Visible in UI (Data Saves but Isn't Displayed)
### Summary
The "Final Notes" field on quarterly reports appears to save correctly to the database/logs, but is **not visible anywhere in the UI**. Currently, the only way to verify or access this data is by inspecting network logs via browser developer tools — which is not a practical or sustainable workflow for regular use.

Image

### Steps to Reproduce
1. Enter text into the "Final Notes" field on a quarterly report.
2. Save the report.
3. Attempt to locate the "Final Notes" content anywhere in the UI (report view, PDF export, etc.).
4. Note that the field is not displayed.
5. Open browser developer tools and inspect network requests/logs to confirm the data was in fact saved.

### Expected Behavior
"Final Notes" entered on a quarterly report should be visible and accessible within the UI — either on the report view, in an export, or in another appropriate location — without requiring developer tools.

### Actual Behavior
The data saves successfully on the backend, but there is no UI surface where users can view it. 

### Impact / Why This Matters
- This is flagged as **essential information for decision-making**, particularly for projects with investor involvement.
- If "Final Notes" are being filled out but are effectively inaccessible, it raises the question of who (if anyone) is currently able to use this data — the effort to enter it may be going to waste.
- Risk that this pattern (data saved but not surfaced) could apply to other fields as well; worth checking for similar gaps.

### Suggested Next Steps
- [ ] Investigate why "Final Notes" isn't rendered anywhere in the current UI (check if this was an oversight, a removed feature, or a regression).
- [ ] Determine the intended/expected location(s) for this data to be displayed (report detail view? PDF/print export? summary card?).
- [ ] Scope a fix — this may require a **new UI feature** to properly surface "Final Notes" rather than a simple bug fix.
- [ ] Confirm with stakeholders (e.g., Field Ops, whoever owns quarterly report requirements) on the correct display format and placement.
- [ ] Check for other fields with a similar "saves but doesn't display" pattern.

DanCoria:
## Summary
- The "Final Notes" field on a progress report's status transition (entered on the Submit Report step) persisted to the backend but had no UI surface to display it.
- `WorkflowCard` now finds the most recent workflow event with notes and renders them in a "Final Notes" section below the status stepper.

Closes #1851

## Test plan
- [x] Submitted a report transition with text in "Final Notes"
- [x] Confirmed the note renders under the status stepper on the report detail page

### DIFF
--- a/src/components/RichText/RichTextField.tsx
+++ b/src/components/RichText/RichTextField.tsx
@@ -358,7 +358,7 @@ const isEditorEmpty = (ref: RefObject<HTMLElement>) =>
   // If not the default block, then assume there are more actions the remaining block could take.
   !!ref.current.querySelector('.codex-editor .ce-block .ce-paragraph');
 
-const isDataEmpty = (value: Nullable<RichTextData>) =>
+export const isDataEmpty = (value: Nullable<RichTextData>) =>
   !value || value.blocks.length === 0;
 
 const isRichTextEqual = (
--- a/src/scenes/ProgressReports/Detail/WorkflowCard.tsx
+++ b/src/scenes/ProgressReports/Detail/WorkflowCard.tsx
@@ -6,7 +6,7 @@ import {
   Typography,
 } from '@mui/material';
 import { RelativeDateTime } from '~/components/Formatters';
-import { RichTextView } from '~/components/RichText';
+import { isDataEmpty, RichTextView } from '~/components/RichText';
 import { ButtonLink } from '~/components/Routing';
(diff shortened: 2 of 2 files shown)


sethmcknight:
@DanCoria should this be closed in favor of https://github.com/SeedCompany/cord-field/pull/1862 or are both still needed?

sethmcknight:
Closing in favor of https://github.com/SeedCompany/cord-field/pull/1862

